PR #1, opened 2025-03-18
fix(rpc): cast send --legacy support

Reecepbcups:
## Summary

title. adds non EIP1559 txs support so `cast` works as expected.


vladjdk:
@Reecepbcups could you repoint this to the main branch? We grafted the repo to include historical changes from the original repo, so all the hashes changed.

Reecepbcups:
moved to https://github.com/cosmos/evm/pull/27 bc histories

PR #2, opened 2025-03-18
feat: evm `PostTxProcessing` hooks

### LINKED ISSUE #18: [Feature]: add hooks after an EVM Tx is processed
### Summary

`PostTxProcessing` to allow other modules to hook into properly processed EVM transactions. 

### Problem Definition

This is needed to extend EVM functionality without forking the EVM module.

Any modules that register in app.go can add additional logic once a EVM tx has been processed, but before it is committed. Modules can then validate the eth receipt, and return nil or an error depending on their logic. if nil, the transaction is committed as normal. Else, show an error to the user and do not commit.

This is a require feature for modules to write things like blocklist based on the tx origin or To() recipient. Used by myself with a new `x/restriction` module that uses a small fork the [sdk's x/bank blockedAddresses to also block EVM txs](https://github.com/strangelove-ventures/cosmos-sdk/pull/2/files).

Adding sanctions requirements to the EVM is out of scope for this PR. This is just to allow modules to do similar things for application logic.

### Proposed Feature

https://github.com/cosmos/evm/pull/2

### Work Breakdown

.

Reecepbcups:
closes #18

## Summary

Adds hooks to the EVM for PostTxProcessing


vladjdk:
@Reecepbcups could you repoint this to the main branch? We grafted the repo to include historical changes from the original repo, so all the hashes changed.

PR #3, opened 2025-03-19
fix: deps bump

vladjdk:
* remove local example_chain
* bump ibc-go
* bump cosmos-sdk


example_chain/go.mod
```diff
@@ -17,11 +17,11 @@ require (
 	cosmossdk.io/x/upgrade v0.1.4
 	github.com/cometbft/cometbft v0.38.15
 	github.com/cosmos/cosmos-db v1.1.1
-	github.com/cosmos/cosmos-sdk v0.50.11
-	github.com/cosmos/evm v0.0.0-20240806115830-ab675cc0c776
+	github.com/cosmos/cosmos-sdk v0.50.13
```
reecepbcups, line 20: github.com/evmos/cosmos-sdk replace is still at the bottom of the file (v0.50.9)

All it has is a deep copy, ideally that could just be moved into a new patch release of the SDK as non breaking (like a new DeepCopy function). Dependend on by CosmWasm in their testing

example_chain/go.mod
```diff
@@ -245,7 +245,7 @@ require (
 )
 
 // LOCAL REPLACEMENTS FOR Cosmos EVM
-replace github.com/cosmos/evm => ..
+//replace github.com/cosmos/evm => ..
```
Eric-Warehime: I think we want this replace still. Why should we remove it?

example_chain/go.mod
vladjdk: Yeah, we have a semi-deforked branch on https://github.com/cosmos/cosmos-sdk/commits/feature/evm/, and I'm going to create a new defork PR once we fork their geth into the org as well.

Currently sitting on https://github.com/cosmos/evm/tree/chore/migrate-forks

example_chain/go.mod
vladjdk: It seems to break those that import this repo as a module

vladjdk: Nvm you're right

vladjdk: fixed

Eric-Warehime (approved):

PR #8, opened 2025-03-20
tests: Remove some example chain deps

Eric-Warehime:
# Description

* First set of changes which work towards removing the example chain from the top go mod dependencies.
* Updated `x/evm` in a few places I saw it.
* Enables blank issues in the issue templates--will make it easier for us to manage work.

---


vladjdk:
some unit tests are failing

PR #36, opened 2025-03-31
feat(erc20): add convert coin endpoint

Reecepbcups:
# Description

ref: https://github.com/cosmos/evm/issues/17

Issue #17, opened 2025-03-22
[Feature]: permissionless erc20 <-> cosmos coin


Reecepbcups:
### Summary

Currently only IBC coins and governance proposals set ERC20->Cosmos bank native pairs. 

1. Functionality should be added to allow permissionless conversion of an ERC20 to a new (namespaced) bank pair.
    * 0x5FbDB... becomes erc20-0x5FbDB...  for example
2. A user should be able to convert that cosmos bank pair back to the ERC20 (if required, evm prank as the contract/Ownable admin to mint?)

### Problem Definition

IBC and gov only ERC20s is far too limiting. Cosmos bank tokens are superior and being able to convert to and from an ERC20<>bank is ideal. This way a user can:

`eth mainnet -> LZ/IBC -> Chain WTOKEN -> ConvertERC20ToCosmosToken -> perform actions with it (staking, other) -> RedeemERC20FromCosmosToken -> LZ/IBC back out`

This `Chain WTOKEN` is __not__ the native token (the native_precompile), it is some new token.

### Proposed Feature

Rough POC I put together. This assumes the codebase stays close to its current form and just appends on features. 

* Ideally LinkERC20 isn't required and you can just ConvertERC20 to the cosmos coin for mint/redemption. Then remove IBC/Gov requirements for this
* The bank denom should match the erc20 denom name since the focus is EVM. 
* If the linking nature persists, I do not see why we could not allow anyone to link, linking is sandboxed to its own namespace. Maybe a param for an optional fee cost here, [similarly like token factory](https://github.com/strangelove-ventures/tokenfactory/blob/main/x/tokenfactory/keeper/createdenom.go#L76).

### Script

```bash
# deploy some standard `contract MyERC20 is ERC20, ERC20Burnable`
forge script ./contracts/script/DeployERC20.s.sol --rpc-url http://localhost:8545 -vvvv --broadcast --base-fee 0 --gas-price 0 --priority-gas-price 0 --legacy

CONTRACT=0x5FbDB2315678afecb367f032d93F642f64180aa3

# this is what is being created
ondod tx controlpanel link-erc20 ${CONTRACT} --from=acc0 --yes
ondod q erc20 token-pairs # validate it is a new pair


# redeem ERC20 -> native token
printf "12345678\n" | ondod keys unsafe-import-eth-key my-account 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --keyring-backend=test
[...]

PR #37, opened 2025-03-31
feat: permissionless erc20 registration

Reecepbcups:
Depends on https://github.com/cosmos/evm/pull/36

ref: #17


vladjdk (changes requested):
lints

PR #54, opened 2025-04-04
feat: post tx hooks

### LINKED ISSUE #18: [Feature]: add hooks after an EVM Tx is processed

vladjdk:
Reopening the PR #2 from @Reecepbcups

Closes #18


x/vm/keeper/state_transition.go
```diff
@@ -180,10 +185,56 @@ func (k *Keeper) ApplyTransaction(ctx sdk.Context, tx *ethtypes.Transaction) (*t
...
+			contractAddr = crypto.CreateAddress(msg.From(), msg.Nonce())
+		}
+
+		cumulativeGasUsed := res.GasUsed
+		if ctx.BlockGasMeter() != nil {
+			limit := ctx.BlockGasMeter().Limit()
+			cumulativeGasUsed += ctx.BlockGasMeter().GasConsumed()
+			if cumulativeGasUsed > limit {
+				cumulativeGasUsed = limit
+			}
```
Eric-Warehime, line 201: Not too familiar with this code path at this point, but is this the expected behavior? Possibly we should be stopping execution if we've gone over the gas limit and returning a out of gas error?

Eric-Warehime (commented):
Looks good, but I think we should add tests around some of the success and failure conditions here, especially since this is modifying the state transition function.

x/vm/keeper/state_transition.go
vladjdk: I viewed some historical Ethermint updates, and it seems like this value from the receipt is only used with external tooling. I still haven't found a concrete example of when something like this could happen, but if you look at https://github.com/cosmos/evm/blob/6124709a66c2f58ddd2ed0a6ddd4a039c2557802/x/vm/keeper/state_transition.go#L410-L431, the minimumGasUsed could be higher than the actual gas used, which may cause the cumulativeGasUsed + BlockGasMeter to exceed the limit.

vladjdk: I added a test where this case would be hit. This could probably happen in the real world where someone's transaction is at the end of the FIFO list and the gas limit exceeds the rest of the Cosmos block gas limit, but the consumed gas still remains under it. In this case, the EVM would refund the gas.

x/vm/keeper/state_transition.go
```diff
@@ -179,10 +184,56 @@ func (k *Keeper) ApplyTransaction(ctx sdk.Context, tx *ethtypes.Transaction) (*t
...
+		}
+
+		signerAddr, err := signer.Sender(tx)
+		if err != nil {
+			return nil, errorsmod.Wrap(err, "failed to extract sender address from ethereum transaction")
+		}
+
+		if err = k.PostTxProcessing(tmpCtx, signerAddr, msg, receipt); err != nil {
+			// If hooks returns an error, revert the whole tx.
+			res.VmError = fmt.Sprintf("failed to execute post transaction processing: %s", err)
```
Eric-Warehime: nit: can we use `errorsmod.Wrap` here?

x/vm/keeper/state_transition.go
Eric-Warehime: Maybe a separate question, but since we're refunding any gas not consumed via `res.GasUsed`, does that mean the post tx hooks are essentially free to execute?

Eric-Warehime (approved):
LGTM. Let's add some comment around gas usage not being accounted in the post hooks and use errorsmod if we can.

vladjdk:
Thanks. Added a comment that the hooks won't charge for any gas and can be thought of like EndBlockers for individual EVM transactions in that sense. Fixed the error as well.

x/vm/keeper/state_transition.go
vladjdk: Yeah, added a comment above them to clarify this

Issue #47, opened 2025-04-03
[EPIC] Removal of x/authz dependency on precompiles


vladjdk:
# Removal of x/authz dependency on precompiles

### For erc20 precompile
- Add `Allowance` state and CRUD methods in `x/erc20` module.
- Replace the parts of the ERC20 precompile that read/write allowances through authzKeeper with the use of erc20Keeper’s `Allowance` CRUD methods.

### For other precompiles
- Remove `authzKeeper` field and related methods from common Precompile struct. [ref. [Precompile.AuthzKeeper](https://github.com/cosmos/evm/blob/d781e119de10b210801921014c2d2c2e13b51045/precompiles/common/precompile.go#L21)]
- Remove precompile methods that use AuthzKeeper. [ref. e.g. [staking.Approve](https://github.com/cosmos/evm/blob/d781e119de10b210801921014c2d2c2e13b51045/precompiles/staking/approve.go#L30-L32)]

### 1. Modularity of precompiles

- Cosmos SDK and IBC-go modules such as bank, staking, and ibctransfer do not have a dependency on the authz module by default. However, the precompiles that are implemented to use these modules currently depend on the authz module.
- By applying a modular structure that separates dependencies, chains that do not use the authz module can avoid having to include it just to use cosmos/evm.

### 2. Compatibility between ERC20 and authz spec

- The Authz [Grant](https://github.com/cosmos/cosmos-sdk/blob/53dea340efc74cc6877530627cace6e7dc049db4/x/authz/authz.pb.go#L73-L81) data structure used to store permissions in the authz module includes an Expiration field, whereas allowances in the ERC20 spec do not. When approve is called through ERC20 precompile, Expiration is set to `1 year` by default. However, When it is called thruogh x/auth module, it could be different value.
- The [SendAuthorization](https://github.com/cosmos/cosmos-sdk/blob/53dea340efc74cc6877530627cace6e7dc049db4/x/bank/types/send_authorization.go#L21) type from the x/bank module, which is used for storing ERC20 allowances, includes an AllowList field—a recipient address whitelist—which also does not exist in the ERC20 specification. When approve is called through the ERC20 precompile, AllowList is set to empty list that means all addresses can be recipient address, by default. However, if someone directly creates a `SendAuthorization` through the authz module, those fields can be used. This may lead to cases where an `allowance` exists, but `transferFrom` fails for certain addresses, or existing allowances disappear without any Transfer or Approval event being emitted.

[...]

PR #62, opened 2025-04-09
feat: remove x/authz dependency from precompiles

cloudgray:
# Description

## Removal of x/authz dependency on precompiles

- Remove authzKeeper dependency from common Precompile struct
- For staking and ics20 precompile
   - remove authorization methods (`Approve`, `Revoke`, `IncreaseAllowance`, `DecreaseAllowance`, `Allowance`) and its usage.
- For erc20 precompile 
   - Add `Allowance` state CRUD methods in x/erc20 module
   - Use added `Allowance` state of x/erc20 module instead of `Grant` state of x/authz module for authorization methods (`Approve`, `Revoke`, `IncreaseAllowance`, `DecreaseAllowance`, `Allowance`)

Closes: #47 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


evmd/precompiles.go
```diff
@@ -60,15 +60,14 @@ func NewAvailableStaticPrecompiles(
 		panic(fmt.Errorf("failed to instantiate bech32 precompile: %w", err))
 	}
 
-	stakingPrecompile, err := stakingprecompile.NewPrecompile(stakingKeeper, authzKeeper)
+	stakingPrecompile, err := stakingprecompile.NewPrecompile(stakingKeeper)
```
dongsam, line 61: In the arguments of `NewAvailableStaticPrecompiles`, it also seems possible to remove `authzKeeper`.

Additionally, I noticed that dependencies on Erc20Keeper,  werc20 and erc20 precompiles (which are planned to be removed later as they will have their own state) still remain, so I’m leaving a note just in case it was overlooked.

FYI: temporary removed for IBC v2 test on this [commit](https://github.com/cosmos/evm/commit/901ac041b056b7f01de4cfa8f4fe01daaf833ced)

cloudgray: Thank you for feedback.

1. In case of `NewAvailableStaticPrecompiles`, I missed it, I'll fix it.
2. In case of erc20 and werc20, I intended that the task order that 1) add allowance state in x/erc20 module, 2) modify erc20 precompile, 3) modify werc20 precompile.
3. And I will consider [commit](https://github.com/cosmos/evm/commit/901ac041b056b7f01de4cfa8f4fe01daaf833ced).

precompiles/staking/testdata/StakingCaller.sol
```diff
@@ -90,78 +86,85 @@ contract StakingCaller {
...
     /// @param _creationHeight The creation height of the unbonding delegation.
     function testCancelUnbonding(
-        address _addr,
         string memory _validatorAddr,
         uint256 _amount,
         uint256 _creationHeight
     ) public {
-        staking.STAKING_CONTRACT.cancelUnbondingDelegation(
-            _addr,
+        _dequeueUnbondingEntry();
```
zsystm, line 157: Hey @cloudgray, I noticed that `_dequeueUnbondingEntry()` is called at the beginning of several test cases like `testUndelegate`, `testDelegate`, and `testRedelegate`.
Just curious — is there a specific reason we need to dequeue at the start of each test? Would love to understand the motivation behind it.

zsystm (commented):
Looks good to me so far — the authz removal seems to be handled well.

I had one question while reviewing the test code, which I left as a comment.

Also, this is not directly related to the scope of your PR, but I noticed that some of the existing test cases don’t validate balance changes. For example, after a `delete`, there’s no assertion checking whether the delegator’s balance was reduced accordingly. Since the precompiled contract involves complex state transitions, I think it would be valuable to add more thorough state validation to ensure the transitions happen as expected.

This can probably be addressed in a separate PR, so just leaving this here for reference. I’ll opened [an issue](https://github.com/cosmos/evm/issues/91) to track it.

precompiles/staking/testdata/StakingCaller.sol
cloudgray: @zsystm (cc. @dongsam )
Currently, the `StakingCaller` contract acts as the delegator and manages staking directly. It also manages the mapping of `EoA -> delegation and EoA -> UnbondingQueue` internally. Because of this, when an item in the unbonding queue is dequeued by the `x/staking` module, the change must be reflected back into the contract.

However, in order to reflect the `unbondingQueue` dequeued from the `x/staking` module into the `StakingCaller` contract as soon as it happens, it would likely require either a separate backend service or additional implementation in the SDK module’s EndBlocker, which I felt would go beyond the scope of this PR.

Therefore, I added a simpler version of the logic: when an EoA performs a `delegate` or `undelegate` action, the contract checks if there are any unbondingEntries that have passed their unbonding period and processes them accordingly.

This logic isn’t directly used in integration_test.go, but without it, the contract would be incomplete—so it was included.

evmd/app.go
```diff
@@ -568,7 +568,6 @@ func NewExampleApp(
 			app.DistrKeeper,
 			app.BankKeeper,
 			app.Erc20Keeper,
-			app.AuthzKeeper,
 			app.TransferKeeper,
```
dongsam, line 584: Since the existing precompiles no longer depend on authz, it seems that the authz keeper can also be removed from `erc20keeper.NewKeeper`.

dongsam (commented):
I haven’t gone through everything yet, but from what I’ve seen so far in app.go and the ics20 precompile section, it looks good

I’ll continue the review when I have more time.

evmd/app.go
cloudgray: Thank you for feedback.
As you said, authzKeeper is  unnecessary not only in erc20 keeper so I'll remove it.

cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/aacfa50c3378e0569625768d79ac4617a06a2885

x/erc20/keeper/allowance.go
```diff
@@ -0,0 +1,164 @@
+package keeper
```
dongsam, line 1: This comment is related to a change that wasn't mentioned in the review of the previous PR (#90), so I'm leaving it here.

With the addition of the `allowance` state to the ERC20 module, a [state-breaking change](https://docs.cosmos.network/v0.50/build/building-modules/upgrade#consensus-version) has occurred.  
However, it looks like the consensus version is still set to [4](https://github.com/cosmos/evm/blob/aacfa50c3378e0569625768d79ac4617a06a2885/x/erc20/module.go#L30)

There seem to be two possible options to address this:

1. **Increment the consensus version to `5`** — to maintain continuity with the version history used in `evmos/os`.
2. **Reset the consensus version to `1`** — since we're starting fresh with `cosmos/evm`, setting it to the initial version (`1`) could also make sense.

If this isn't the appropriate PR to make a decision on this, I’d appreciate it if we could continue the discussion in a separate issue.

c.c @cloudgray @zsystm @vladjdk

cloudgray: I missed modifying consensus version of x/erc20 module. 🙏

I think consensus version should be `5` in this PR. 

And resetting the consensus version to `1` is something that applies to other modules as well, so I think it should be handled in a separate issue. Because x/vm module and `x/feemarket` module also have consensus version from `evmos/os` [[x/evm: 8](https://github.com/cosmos/evm/blob/feat%2Fremove-authz-from-precompiles/x/vm/module.go#L29)] [[x/feemarket: 5](https://github.com/cosmos/evm/blob/feat%2Fremove-authz-from-precompiles/x/feemarket/module.go#L29)]

c.c @dongsam @zsystm @vladjdk

precompiles/staking/testdata/StakingCaller.sol
zsystm: @cloudgray 
Thank you for the explanation — it makes sense.
Could you briefly summarize this and add it as a comment above the `_dequeueUnbondingEntry()` function?
It would be helpful to have the context documented for future reference.

cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/f9d88877b7cb21bfbe505333f6765d590c4d1f79

precompiles/testutil/contracts/FlashLoan.sol
```diff
@@ -13,11 +13,12 @@ contract FlashLoan {
 
     function flashLoan(
         address _token,
-        string memory _validator,
-        uint256 _amount
-    ) public returns (bool) {
+        string memory _validator
+    ) public payable returns (bool) {
```
vladjdk, line 17: Why are we making this payable? The token we're using for the flashloan is an ERC20, but we're overriding the amount with the `msg.value`, which is the native token.

precompiles/erc20/approve.go
```diff
@@ -213,89 +210,38 @@ func (p Precompile) DecreaseAllowance(
 	}
 
 	// TODO: check owner?
-	if err := p.EmitApprovalEvent(ctx, stateDB, p.Address(), spender, amount); err != nil {
+	if err := p.EmitApprovalEvent(ctx, stateDB, owner, spender, amount); err != nil {
```
vladjdk: We should keep p.Address() in the event instead - owner can be deduced as the caller here

precompiles/erc20/approve_test.go
```diff
@@ -521,51 +410,22 @@ func (s *PrecompileTestSuite) TestDecreaseAllowance() {
...
-			malleate: func() []interface{} {
-				return []interface{}{
-					s.keyring.GetAddr(1), big.NewInt(decreaseAmount),
-				}
-			},
-			errContains: "does not exist or is expired",
-		},
-		{
-			name: "fail - decrease allowance with existing authorization only for other denominations",
+			name: "fail - decrease allowance without existing allowane",
```
vladjdk: nit: typo on allowane

precompiles/erc20/approve_test.go
```diff
@@ -82,59 +78,30 @@ func (s *PrecompileTestSuite) TestApprove() {
 			errContains: "causes integer overflow",
 		},
 		{
-			name: "fail - approve to zero with existing authorization only for other denominations",
```
vladjdk, line 85: Let's add these tests in anyways, but instead, we set up multiple allowances for different tokens

precompiles/erc20/approve_test.go
```diff
@@ -143,45 +110,22 @@ func (s *PrecompileTestSuite) TestApprove() {
...
+				s.requireAllowance(
+					s.precompile.Address(),
+					s.keyring.GetAddr(0),
+					s.keyring.GetAddr(1),
+					big.NewInt(amount),
 				)
 			},
 		},
 		{
-			name: "pass - approve with existing authorization in different denomination",
```
vladjdk, line 155: Same as above, let's have multiple token allowances here

precompiles/erc20/approve_test.go
```diff
@@ -521,51 +410,22 @@ func (s *PrecompileTestSuite) TestDecreaseAllowance() {
...
-						sdk.NewInt64Coin(s.bondDenom, 1),
-					),
-					[]string{},
-				)
-			},
+			errContains: "does not exist",
 		},
 		{
-			name: "pass - decrease allowance with existing authorization",
+			name: "pass - decrease allowance with existing allowane",
```
vladjdk: nit: typo on allowane

precompiles/erc20/approve_test.go
```diff
@@ -597,71 +458,23 @@ func (s *PrecompileTestSuite) TestDecreaseAllowance() {
...
+					s.precompile.Address(),
+					s.keyring.GetAddr(0),
+					s.keyring.GetAddr(1),
+					common.Big0,
 				)
 			},
 		},
 		{
-			name: "pass - decrease allowance to zero for denom with existing authorization in other denominations",
+			name: "fail - decrease allowance with existing allowane but decreased amount too high",
```
vladjdk: nit: typo on allowane

precompiles/erc20/approve_test.go
```diff
@@ -597,71 +458,23 @@ func (s *PrecompileTestSuite) TestDecreaseAllowance() {
...
+				s.requireAllowance(
+					s.precompile.Address(),
+					s.keyring.GetAddr(0),
+					s.keyring.GetAddr(1),
+					common.Big0,
 				)
 			},
 		},
 		{
-			name: "pass - decrease allowance to zero for denom with existing authorization in other denominations",
```
vladjdk, line 632: Let's keep this check in - use two separate allowances and check that decreasing one doesn't decrease another

precompiles/erc20/tx.go
```diff
@@ -123,21 +141,10 @@ func (p *Precompile) transfer(
...
-		// the maxUint256 value.
-		newAllowance = abi.MaxUint256
-	} else {
-		newAllowance = new(big.Int).Sub(prevAllowance, amount)
-	}
-
-	if err = p.EmitApprovalEvent(ctx, stateDB, from, spenderAddr, newAllowance); err != nil {
-		return nil, err
+	if isTransferFrom {
+		if err = p.EmitApprovalEvent(ctx, stateDB, from, spenderAddr, newAllowance); err != nil {
```
vladjdk, line 141: This isn't required by the ERC20 standard. Should we just remove this entirely? Moreover, the owner transferring their own tokens should return a max, but we removed that logic.

Imo this should just go.

Ref: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/56c07e59c4ddc95f77e71ddb54976f13b6863f90/contracts/token/ERC20/ERC20.sol#L129

precompiles/erc20/tx_test.go
```diff
@@ -198,15 +187,8 @@ func (s *PrecompileTestSuite) TestTransferFrom() {
...
-				err := s.network.App.AuthzKeeper.SaveGrant(
-					ctx,
-					spender.AccAddr,
-					owner.AccAddr,
-					&banktypes.SendAuthorization{SpendLimit: sdk.Coins{sdk.Coin{Denom: tokenDenom, Amount: math.NewInt(300)}}},
-					&expiration,
-				)
-				s.Require().NoError(err, "failed to save grant")
+				err := s.network.App.Erc20Keeper.SetAllowance(ctx, s.precompile.Address(), owner.Addr, spender.Addr, big.NewInt(300))
+				s.Require().NoError(err, "failed to set allowance")
```
vladjdk: different error message here, but the behaviour seems to be the same, but with setting allowances

precompiles/staking/integration_test.go
```diff
@@ -1928,319 +1491,42 @@ var _ = Describe("Calling staking precompile via Solidity", Ordered, func() {
 		})
```
vladjdk, line 1491: Not in scope, but I'm not understanding why the line right above this expects the error to be nil but the description says that we expect an error while calling the precompile. I ran this, and nothing is erroring out. Is this test even working properly?

precompiles/staking/tx.go
```diff
@@ -220,19 +175,12 @@ func (p *Precompile) Delegate(
...
-
 	// Emit the event for the delegate transaction
 	if err = p.EmitDelegateEvent(ctx, stateDB, msg, delegatorHexAddr); err != nil {
 		return nil, err
 	}
 
-	if !isCallerOrigin && msg.Amount.Denom == evmtypes.GetEVMCoinDenom() {
+	if contract.CallerAddress != origin && msg.Amount.Denom == evmtypes.GetEVMCoinDenom() {
 		// get the delegator address from the message
 		delAccAddr := sdk.MustAccAddressFromBech32(msg.DelegatorAddress)
```
vladjdk: Is this statement even necessary given the fact that we now enforce `contract.CallerAddress == delegatorHexAddr`?

precompiles/testutil/contracts/FlashLoan.sol
```diff
@@ -54,11 +55,12 @@ contract FlashLoan {
 
     function flashLoanWithRevert(
         address _token,
-        string memory _validator,
-        uint256 _amount
-    ) public returns (bool) {
+        string memory _validator
+    ) public payable returns (bool) {
```
vladjdk, line 59: Same here - why is this payable if we're using the ERC20?

x/erc20/keeper/allowance.go
vladjdk: Let's open a new issue for this yes

x/erc20/keeper/allowance_test.go
```diff
@@ -0,0 +1,494 @@
...
+				suite.network.App.Erc20Keeper.SetToken(ctx, pair)
+			},
+			true,
+			"",
+		},
+		{
+			"pass - positive value",
+			func() {
+				pair := types.NewTokenPair(erc20Addr, "coin", types.OWNER_MODULE)
+				suite.network.App.Erc20Keeper.SetToken(ctx, pair)
```
vladjdk: dont see the difference between this and the above test case

x/vm/statedb/integration_test.go
```diff
@@ -1,6 +1,7 @@
 package statedb_test
```
vladjdk, line 1: This test now passes when we remove the `.Copy()` method from the `MultiStoreSnapshot` method. I believe this test is now ineffective as contracts can no longer perform module actions on behalf of other users without approvals.

Do we still need this test, and moreover, do we want to bring back that functionality in the future within other precompiles? I don't think we need to add it within ICS20 and Staking again, but it would be nice to potentially have it for other modules in the future (i.e. if there's a dex module or a lending module that might make use of flashloans).

We should still find a way to make this test effective in testing the fix that was added in with `.Copy()` in the stateDB `MultiStoreSnapshot()` function.

precompiles/staking/tx_test.go
```diff
@@ -934,7 +859,7 @@ func (s *PrecompileTestSuite) TestUndelegate() {
 
 	testCases := []struct {
 		name                  string
-		malleate              func(delegator, grantee testkeyring.Key, operatorAddress string) []interface{}
+		malleate              func(delegato testkeyring.Key, operatorAddress string) []interface{}
```
vladjdk: nit: typo in delegato

precompiles/erc20/integration_test.go
```diff
@@ -1394,290 +1356,229 @@ var _ = Describe("ERC20 Extension -", func() {
...
+					is.ExpectAllowanceForContract(callType, contractsData, owner.Addr, spender.Addr, common.Big0)
 				},
 					Entry(" - direct call", directCall),
 					Entry(" - through erc20 contract", erc20Call),
 					Entry(" - through erc20 v5 contract", erc20V5Call),
 				)
 
-				When("the grantee is the same as the granter", func() {
+				When("the spender is the same as the owner", func() {
 					// NOTE: We differ in behavior from the ERC20 calls here, because the full logic for approving,
```
vladjdk: Is this comment still true?

precompiles/erc20/integration_test.go
```diff
@@ -1394,290 +1356,229 @@ var _ = Describe("ERC20 Extension -", func() {
...
-							grantee.Addr, granter.Addr, authzCoins,
+							owner.Addr, spender.Addr, allowance,
 						)
 					},
 						Entry(" - through erc20 contract", erc20Call),
 						Entry(" - through erc20 v5 contract", erc20V5Call),
 					)
 				})
-
-				DescribeTable("it should return an error if approving 0 and allowance only exists for other tokens", func(callType CallType) {
```
vladjdk, line 1638: We're removing a lot of these checks for other tokens, and I understand why, but I think we should open up a new issue for testing state behaviour when we have multiple ERC20s with their own approvals to ensure that there's no state overrides on any allowances changes. Since we have multiple ERC20s that all access a single module state in x/erc20, we should make sure that there's no way for one ERC20 to modify the allowances of another.

precompiles/erc20/integration_test.go
```diff
@@ -2267,147 +2164,87 @@ var _ = Describe("ERC20 Extension -", func() {
...
+					is.ExpectTrueToBeReturned(ethRes, erc20.IncreaseAllowanceMethod)
+					is.ExpectAllowanceForContract(callType, contractsData, contractAddr, spender.Addr, approveAmount)
 				},
 					Entry(" - contract call", contractCall),
 					Entry(" - through erc20 caller contract", erc20CallerCall),
 				)
 			})
 		})
 
-		When("an allowance exists for other tokens", func() {
```
vladjdk, line 2294: Same comment as above

vladjdk (changes requested):

vladjdk:
Gmgm, just finished reviewing this. First off, tremendous work on this. This was a big PR and overall I think it's a big step forward in the right direction. Thank you!

I left a few comments and a few requested changes. There's also a new merge conflict that needs to be resolved in the changelog.

x/vm/statedb/integration_test.go
cloudgray: I think this test is still in need.

Referencing the [comments below](https://github.com/cosmos/evm/blob/feat%2Fremove-authz-from-precompiles/x/vm/statedb/integration_test.go#L40-L45), the purpose of this test is verifying that state changes persisted after the precompile calls.

Approving FlashLoan contract to call precompile is not the target functionality of this test but the part of example scenario reported.

Though I'm not sure `Copy()` method of `MultiStoreSnapshot` would be removed after additional test or not, that is implementation of cosmos-sdk and this test is for state changes with EVM contract and precompiles. I think these two should be thought seperately.

precompiles/testutil/contracts/FlashLoan.sol
cloudgray: The reason I add `payable` to this function is to test existing test without x/authz functionality. It is same approach with modification of [precompiles/staking/testdata/StakingCaller.sol](https://github.com/cosmos/evm/blob/feat%2Fremove-authz-from-precompiles/precompiles/staking/testdata/StakingCaller.sol#L93) and precompiles/staking/integration_test.go

I think this is just example contract for vm/statedb/integration_test.go

Because the purpose of test is verifying that state chages after precompile call persists, it is not important what this contract do with ERC20 token. Borrowing and paying back ERC20 is meaningful action only in point of testing State change before and after precompile call. Even before modification, FlashLoan contract borrows ERC20 and delegate with native coin and pays back ERC20.

precompiles/testutil/contracts/FlashLoan.sol
cloudgray: I think [reply above](https://github.com/cosmos/evm/pull/62#discussion_r2062890619) could also applied to this question.

precompiles/erc20/approve.go
cloudgray: It's my fault. 
I really appreciate you for careful checking!

precompiles/erc20/approve_test.go
cloudgray: There were five instances of the same typo… I’ll fix them all.

x/erc20/keeper/allowance_test.go
cloudgray: I'll remove last test case.

precompiles/staking/tx_test.go
cloudgray: Thank you, I'll fix it.

precompiles/erc20/integration_test.go
cloudgray: No, it doesn't. I'll remove this comment.

Also, while looking for related Evmos PRs, I think that what I additionally have to do is removing restriction of `owner == spender` case. [ref: [precompiles/erc20/approve.go]](https://github.com/cosmos/evm/blob/feat%2Fremove-authz-from-precompiles/precompiles/erc20/approve.go#L173-L175)
This case was impossible with x/authz module because x/authz module does not allow `owner == spender` case. [ref: [evmos PR #2088](https://github.com/evmos/evmos/pull/2088/files#r1407646217)]
But ERC20 contract allow this case. [ref: [openzeppelin ERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/56c07e59c4ddc95f77e71ddb54976f13b6863f90/contracts/token/ERC20/ERC20.sol#L273-L284)]

And erc20 precompile `Allowance` also does not have to return `MaxUint256`. 
The reaseon erc20 precompile `Allowance` returns MaxUint256 for `owner == spender` case is because owner could not approve himself.
But if we remove restriction of `owner == spender`, `Allowance' does not have to care about it.

@vladjdk (cc. @zsystm @dongsam) How do you think about it?

precompiles/erc20/tx_test.go
cloudgray: I could not understand what you mean exactly.

Do you mean, setting `Allowance` in x/erc20 state is same with setting `Grant` in x/authz state, but error message is changed?

If my understanding is right, my reply is, 
I wanted to remove authz related naming as possible. Does it look awkward?

precompiles/erc20/tx.go
cloudgray: I agree with you.

I thought an Approval event was emitted when TransferFrom was called in an ERC20 contract, but after reviewing ERC20.sol you mentioned again, that’s not the case.
I think it would be better to just cleanly remove it.

And as you said, emission of Approval Event is not consistent with `Allowance` that returns MaxUint256 for `owner == spender` case. 

@zsystm @dongsam How do you think about removing emission of Approval event in case of TransferFrom?

precompiles/staking/integration_test.go
cloudgray: In my opinion, it is fault of who developed this test case.

`CallContractAndCheckLogs` checks not only whether tx succeeds, but also tx emits expected event, tx reverts with expected error. 

So, I suspect his intension was "unexpected error while calling contract" but because it could fails for other reason, I think general message like "error while calling the contract and checking logs" is better.

I'll fix wrong error verification related to this.

I appreciate you for careful review!

dudong2:
In the cosmos-evm context, naming convention is `interfaces.go`, not `expected_keepers.go`.
It's not critical, but I think it would be better to change it for consistency

evmd/allowance.go
```diff
@@ -0,0 +1,23 @@
...
+// ExampleAllowances creates a slice of allowance, that contains an allowance for the native denom of the example chain
+// implementation.
+var ExampleAllowances = []erc20types.Allowance{
+	{
+		Erc20Address: WEVMOSContractMainnet,
+		Owner:        ExampleOwner,
+		Spender:      ExampleSpender,
+		Value:        math.NewInt(100),
+	},
+}
```
dudong2: In my opinion, constants for tests are better added in `testutil/constants/constants.go` unless there is a clear reason for them to be other files.

proto/cosmos/evm/erc20/v1/erc20.proto
```diff
@@ -32,6 +32,27 @@ message TokenPair {
   Owner contract_owner = 4;
 }
 
+// Allowance is a token allowance only for erc20 precompile
+message Allowance {
+  option (gogoproto.equal) = false;
+
+  // erc20_address is the hex address of ERC20 contract
+  string erc20_address = 1;
```
dudong2, line 40:
```suggestion
  // address is the hex address of ERC20 contract
  string address = 1;
```

I think you can skip `erc20_` since it was added to the proto of ERC20.

precompiles/staking/tx.go
cloudgray: No, I think it is not. I'll use delegatorHexAddr stated above.

x/erc20/genesis.go
```diff
@@ -31,12 +34,26 @@ func InitGenesis(
...
+	for _, allowance := range data.Allowances {
+		erc20 = common.HexToAddress(allowance.Erc20Address)
+		owner = common.HexToAddress(allowance.Owner)
+		spender = common.HexToAddress(allowance.Spender)
+		value = allowance.Value.BigInt()
+		err := k.UnsafeSetAllowance(ctx, erc20, owner, spender, value)
+		if err != nil {
+			panic(fmt.Errorf("error setting allowance %s", err))
+		}
+	}
```
dudong2:
```suggestion
	for _, allowance := range data.Allowances {
		erc20 := common.HexToAddress(allowance.Erc20Address)
		owner := common.HexToAddress(allowance.Owner)
		spender := common.HexToAddress(allowance.Spender)
		value := allowance.Value.BigInt()
		err := k.UnsafeSetAllowance(ctx, erc20, owner, spender, value)
		if err != nil {
			panic(fmt.Errorf("error setting allowance %s", err))
		}
	}
```

x/erc20/genesis_test.go
```diff
@@ -148,6 +202,7 @@ func (suite *GenesisTestSuite) TestErc20ExportGenesis() {
 						ContractOwner: types.OWNER_MODULE,
 					},
 				},
+				[]types.Allowance{},
```
dudong2, line 202: How about to add some contents to test export data?

x/erc20/module.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 // consensusVersion defines the current x/erc20 module consensus version.
-const consensusVersion = 4
+const consensusVersion = 5
```
dudong2, line 30: This work doesn't cause breaking changes, so you don't need to increase the consensus version.

dudong2: I think this work doesn't cause breaking changes, so you don't need to increment the consensus version.
Or, if this work cause breaking changes, you need to serve the migration code.

x/erc20/keeper/allowance.go
```diff
@@ -0,0 +1,164 @@
...
+func (k Keeper) DeleteAllowance(
+	ctx sdk.Context,
+	erc20 common.Address,
+	owner common.Address,
+	spender common.Address,
+) error {
+	return k.setAllowance(ctx, erc20, owner, spender, common.Big0, false)
+}
+
+func (k Keeper) setAllowance(
```
dudong2, line 76: In my opinion, the `setAllowance` function should be defined purely as a function to set to store, and the other check logic should be left out as outer functions.

```
SetAllowance()
    - check token pair
    - call UnsafeSetAllowance

UnsafeSetAllowance
    - check owner, spender, value, ...
    - call setAllowance

setAllowance
    - set allowance to store
```

i think this architecture looks so good to me.

precompiles/erc20/integration_test.go
cloudgray: I'm not sure checking allowance for 

The reason this test was meaningful was because [SendAuthorization](https://github.com/cosmos/cosmos-sdk/blob/53dea340efc74cc6877530627cace6e7dc049db4/x/bank/types/authz.pb.go#L34-L41) of x/bank module has `SpendLimit` field as `sdk.Coins` type. 

Namely, when x/authz and x/bank SendAuthorization are used for erc20 precompile allowance, all allowances for same owner and spender has one key/value.

```text
{
    SpendLimit: []sdk.Coin{
        sdk.Coin{Denom: A, Amount: 1}
        sdk.Coin{Denom: B, Amount: 2}
        sdk.Coin{Denom: C, Amount: 3}
        sdk.Coin{Denom: D, Amount: 4}
    }, 
    AllowList: ...
}
```

If owner approves, increases, decreases allowance, regardless of whether denom is existing or not, only one state is accessed. So, verification whether only targeted SpendLimit was modified is important.

But now, the states of different allowances have different keys and values. So, this test case is not effective.

I think it is similar to test scenario below.
1. SendCoins Acct1 -> Acct2
2. Check balance of Acct3 is not changed

@vladjdk Even so, do you think we still need this test case?

cloudgray: Additionally, the reason why there is no test case that checks approving non-zero value for other denom is because approving zero value delege entire `SendAuthorization`.

precompiles/erc20/approve_test.go
cloudgray: ~~I think [this reply](https://github.com/cosmos/evm/pull/62#discussion_r2063258645) could serve as a replacement.~~

precompiles/erc20/approve_test.go
cloudgray: ~~I think [this reply](https://github.com/cosmos/evm/pull/62#discussion_r2063258645) could serve as a replacement.~~

precompiles/erc20/approve_test.go
cloudgray: ~~I think [this reply](https://github.com/cosmos/evm/pull/62#discussion_r2063258645) could serve as a replacement.~~

precompiles/erc20/integration_test.go
cloudgray: ~~I think [this reply](https://github.com/cosmos/evm/pull/62#discussion_r2063258645) could serve as a replacement.~~

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/d69151a339879164a657dd90dcce0a05bb832480

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/d69151a339879164a657dd90dcce0a05bb832480

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/d69151a339879164a657dd90dcce0a05bb832480

precompiles/staking/tx_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/d69151a339879164a657dd90dcce0a05bb832480

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/792ded157987de0a077f6731b1a1fdffb522109e

precompiles/erc20/approve.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/33d63f5fb47e9b3e0fbc93322d2350f7677d16c7

precompiles/staking/integration_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/43ca981823047ad08f5e7294c10f32e73b3a5d48

precompiles/staking/tx.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/aba5934d93a804254a8e847f4643c8848d797d66

precompiles/erc20/tx_test.go
vladjdk: Sorry, don't remember exactly what I meant with this. Seems to be fine at a second glance.

precompiles/erc20/integration_test.go
vladjdk: I think we should add one test case where we check that increasing/decreasing/setting the allowance for one user doesn't change it for another. Because this is a precompile, there may be areas where we can mess this up and we should test for that

precompiles/erc20/integration_test.go
vladjdk: Yeah, I think we should replicate behaviour in ERC20 as close as we can, and it looks like removing Authz frees us of another restriction

precompiles/testutil/contracts/FlashLoan.sol
vladjdk: I see, this is not an issue. See comment in the statedb integration test for the issue.

x/vm/statedb/integration_test.go
vladjdk: I actually see why the test isn't working now. It's subtle, but we're reverting in the wrong spot in the solidity test now. Because of our staking precompile changes and the removal of authz, we need to change how this test operates for the revert case as well.

In the code below, line 81 should be `address(this)` since we now expect the FlashLoan contract to do the staking. Because it's trying to stake on behalf of the sender, and it can't, it's reverting there. However, we want to test that there is in fact a valid attempted stake action. This part should actually succeed, *and then we force a revert*. Because we have `msg.sender` in the delegator field, we don't ever make it to `revert()`.

https://github.com/cosmos/evm/blob/f9d88877b7cb21bfbe505333f6765d590c4d1f79/precompiles/testutil/contracts/FlashLoan.sol#L79-L104

You'll notice by using `address(this)` instead, removing .Copy() from [this line](https://github.com/cosmos/evm/blob/f9d88877b7cb21bfbe505333f6765d590c4d1f79/x/vm/statedb/statedb.go#L114) breaks the test, as it should.

This is actually where a copy is needed - to make sure that we properly revert even handled reversions that use precompile, and don't still store the updated state.

x/erc20/keeper/allowance_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/792ded157987de0a077f6731b1a1fdffb522109e

x/vm/statedb/integration_test.go
cloudgray: refernce
- [CVE-2024-32644](https://github.com/advisories/GHSA-3fp5-2xwh-fxm6)
- [fix](https://github.com/evmos/evmos/blob/b196a522ba4951890b40992e9f97aa610f8b5f9c/x/evm/statedb/statedb.go#L460-L465)

precompiles/erc20/integration_test.go
cloudgray: Then, I'll add testcases that checks target token allowance and other allowances change after calling `Approve`, `IncreaseAllowance`, `DecreaseAllowance`.

cloudgray:
> In the cosmos-evm context, naming convention is `interfaces.go`, not `expected_keepers.go`. It's not critical, but I think it would be better to change it for consistency

applied in https://github.com/cosmos/evm/pull/62/commits/dc225a20b3b987d3ff2f4f993fcf80c62b20aee8

x/erc20/module.go
cloudgray: Ok, I'll lower consensus version to before modification and open issue.

cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/efc0ec02e492ddcce9b7fcc2eb061be1cd72cdeb

evmd/allowance.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/f30d239d3df40963389c07b071576137aa036b28

x/erc20/genesis_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/03a59a38c8ce8b8cc96082e1762e3d9de2f4729d

x/erc20/genesis.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/378a240d193df9cf033f62186747bc9bc80a6193

proto/cosmos/evm/erc20/v1/erc20.proto
cloudgray: Currently, the TokenPair also uses erc20_address as the field name for the token contract address, so I would prefer to maintain consistency if possible.
That said, do you still think it would be better to change it?

```ProtoBuf
// TokenPair defines an instance that records a pairing consisting of a native
// Cosmos Coin and an ERC20 token address.
message TokenPair {
  option (gogoproto.equal) = true;
  // erc20_address is the hex address of ERC20 contract token
  string erc20_address = 1;
  // denom defines the cosmos base denomination to be mapped to
  string denom = 2;
  // enabled defines the token mapping enable status
  bool enabled = 3;
  // contract_owner is the an ENUM specifying the type of ERC20 owner (0
  // invalid, 1 ModuleAccount, 2 external address)
  Owner contract_owner = 4;
}
```

x/erc20/keeper/allowance.go
cloudgray: @dudong2 I mostly agree with your opinion.

But there is one problem. 
If I check token pair external from UnsafeSetAllowance, checking existence of token pair in InitGenesis becomes difficult.

So, my proposal is as below.
1. `SetAllowance` calls `SetAllowanceWithValidation` with arg `allowDisabledTokenPair == false`
2. `SetAllowanceWithValidation` takes part of all validation and if allowDisabledTokenPair == true, it ignores `TokenPair.Enabled == false` and call `setAllowance`.
3. `setAllowance` only take part of setting store.

```golang
SetAllowance()
    - call SetAllowanceWithValidation()

SetAllowanceWithValidation()
    - check token pair
    - **check enabled/disabled tokenPair (optionally)**
    - check owner, spender, value, ...
    - call setAllowance()

setAllowance()
    - if value == nil, delete Allowance
    - set allowance to store

 ```

Comment (cloudgray) [reply x/erc20/keeper/allowance.go:76]: applied in https://github.com/cosmos/evm/pull/62/commits/c39ffaa7c6c6cb38cc5fa049e09ff6a538bc90b2

If you think the change is inappropriate, I'll modify again.

Comment (dudong2) [reply x/erc20/keeper/allowance.go:76]: ok. Then I think below architecture is better.

```
SetAllowance()
    - check token pair with enabled
    - call setAllowance

UnsafeSetAllowance()
    - check token pair w/o enabled
    - call setAllowance

setAllowance()
    - set allowance to store
```

File: x/erc20/module.go (hunk 28 above)
Comment (dudong2) [reply x/erc20/module.go]: I tested node starting and didn't cause breaking changes. test steps are below

```
1. start node with main branch
2. stop the node and build node on the `feat/remove-authz-from-precompiles` branch
3. start the node 
4. no breaking changes.
```

if you agree with this test, you don't need to open the issues.

File: precompiles/staking/tx.go (hunk 16 above)
Comment (vladjdk) [reply precompiles/staking/tx.go]: I meant that we make this check now:
```
if contract.CallerAddress != delegatorHexAddr {
	return nil, fmt.Errorf(ErrDifferentCallerFromDelegator, contract.CallerAddress.String(), delegatorHexAddr.String())
}
```

This means that the precompile will always need to be called by the delegator.

`if contract.CallerAddress != origin` is checking for a legacy scenario, where the contract can delegate on behalf of the user. Since we no longer support that, the above check, `if contract.CallerAddress != delegatorHexAddr`, covers all possible other scenarios - whether a contract is delegating or a user is delegating.

A contract delegating on behalf of another user (`contract.CallerAddress != origin`) is impossible with our changes to authz.

Therefore, we'll never need to update `delAccAddr := sdk.MustAccAddressFromBech32(msg.DelegatorAddress)`.

I think we can remove this entire if statement.

File: precompiles/erc20/tx.go (hunk 13 above)
Comment (cloudgray) [reply precompiles/erc20/tx.go:141]: @vladjdk 
Removing `Approval` event emission from `TransferFrom` should be performed in seperate PR with 
1. openzeppelin-contracts version bump (to v5.x)
2. removing `increaseAllowance` and `decreaseAllowance` from erc20 precompile.
I realized it during fixing broken test cases

Current cosmos/evm uses openzeppelin-contracts v4.9.6
In v4.9.6 ERC20.sol, 
- `transferFrom` emits `Approval` event. [[ref](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dc44c9f1a4c3b10af99492eed84f83ed244203f6/contracts/token/ERC20/ERC20.sol#L313)]
- `increaseAllowance`, `decreaseAllowance` method exists. [[ref](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dc44c9f1a4c3b10af99492eed84f83ed244203f6/contracts/token/ERC20/ERC20.sol#L165-L206)]

So, If I remove `Approval` event from `transferFrom` of erc20 precompile, it means erc20 precompile is compatible with openzeppelin ERC20.sol
In openzeppelin-contracts v5.3.x ERC20.sol
- `TransferFrom doesn't emit `Approval` event.
- `increaseAllowance`, `decreaseAllowance` method are removed. [[ref](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e4f70216d759d8e6a64144a9e1f7bbeed78e7079/contracts/token/ERC20/ERC20.sol#L287-L289)]

So, if I bump up openzeppelin-contracts to v5.3.0 and remove Approval event emission from TransferFrom, Many test contracts that inherits openzeppelin ERC20.sol lose `increaseAllowance`, `decreaseAllowance` methods. And it make integration test fail because it basically test compatibility between erc20 precompile and erc20 solidity contracts.

File: x/erc20/module.go (hunk 28 above)
Comment (cloudgray) [reply x/erc20/module.go]: I agree.

I think we should only open issue on whether initialize consensusVersion of x/modules to `1` or use current consensusVersion increased by evmos/evmos repo.

File: x/erc20/keeper/allowance.go (hunk 29 above)
Comment (cloudgray) [reply x/erc20/keeper/allowance.go:76]: In your new design, I think `check owner, spender, value, ...` is omitted.
I think you mean `check owner, spender, value, ...` is performed in setAllowance().
Is it right?

File: precompiles/erc20/tx.go (hunk 13 above)
Comment (cloudgray) [reply precompiles/erc20/tx.go:141]: Approval event emission was removed in https://github.com/cosmos/evm/pull/62/commits/bff9789fb9a48fc1fd543c689422ed275af13948 

And the change above was rollbacked in https://github.com/cosmos/evm/pull/62/commits/67c496bbdc17ed81547b93c354132226f2240fde.

File: precompiles/erc20/integration_test.go (hunk 21 above)
Comment (cloudgray) [reply precompiles/erc20/integration_test.go]: applied in https://github.com/cosmos/evm/pull/62/commits/bff9789fb9a48fc1fd543c689422ed275af13948 and https://github.com/cosmos/evm/pull/62/commits/67c496bbdc17ed81547b93c354132226f2240fde

File: x/vm/statedb/integration_test.go (hunk 19 above)
Comment (cloudgray) [reply x/vm/statedb/integration_test.go:1]: I fixed FlashLoan.sol in https://github.com/cosmos/evm/pull/62/commits/d2fe36cabfaef529a33f5a3cb6fc950ea9b275d2

File: x/erc20/keeper/allowance.go (hunk 29 above)
Comment (cloudgray) [reply x/erc20/keeper/allowance.go:76]: I modified as below for reducing logic redundancy. (https://github.com/cosmos/evm/pull/62/commits/aacfc4c035c594af9551289d34786054dffa9dc0)

```
SetAllowance()
    - call setAllowance w/o allowDisabledTokenPair

DeleteAllowance()
    - call setAllowance w/ allowDisabledTokenPair

UnsafeSetAllowance()
    - call setAllowance w/ allowDisabledTokenPair

setAllowance()
    - check token pair exists and enabled
    - check owner, spender, value, ...
    - set allowance to store
```

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/b7e8df109fcf0bf656055f88c097bfab98ddb9d5

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/b7e8df109fcf0bf656055f88c097bfab98ddb9d5

precompiles/erc20/approve_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/b7e8df109fcf0bf656055f88c097bfab98ddb9d5

precompiles/erc20/integration_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/b7e8df109fcf0bf656055f88c097bfab98ddb9d5

precompiles/erc20/integration_test.go
cloudgray: applied in https://github.com/cosmos/evm/pull/62/commits/b7e8df109fcf0bf656055f88c097bfab98ddb9d5

x/erc20/keeper/allowance.go
cloudgray: @vladjdk @zsystm @dudong2  @dongsam 
I created issue #117

precompiles/erc20/tx.go
vladjdk: Ah nice attention to detail. Let's do that all in one PR then. Good to skip in this one.

precompiles/staking/tx.go
vladjdk: Still approving but since I believe it's dead code, we should remove later

vladjdk (approved):
Lgtm, thanks for the work :)

Issue #48, opened 2025-04-03
Decimal handling (6 and 18)


vladjdk:
Gracefully handling the interoperability between 6 decimal Cosmos coins and 18-decimal ERC-20 tokens

dudong2:
## Test coverage:

- As mentioned in the Direction section, we plan to add integration and E2E tests before finalizing the work.
    - Fuzz tests with arbitrary 18-decimal values
    - Edge cases (e.g., 1 wei transfers, boundary conditions)
    - ERC20 precompile behavior (transfer, balanceOf, totalSupply)
    - IBC transfers with fractional amounts
    - Staking with a(n) values
    - Additional invariant check will be added:

## Invariants

- Existing [invariants](https://github.com/Kava-Labs/kava/blob/master/x/precisebank/keeper/invariants.go) are comprehensive. 
- But comparison with total supply is missing — we plan to include it

## Migration

- Migration plan is clear and simple:
    - Case 1: Chains already using 18-decimal denoms in both SDK and EVM (e.g., Injective with ainj)
        → No precision mismatch, x/bank can remain

    - Case 2: Chains using 6-decimal denoms in SDK (e.g., uatom)
        → All fractional balances are effectively zero
        → Migration is as simple as replacing x/bank with x/precisebank

vladjdk:
To summarize, this issue only exists in the gas token used in the network. If someone sends less than 1 `ugas` on the EVM side, it gets floored. Consequentially, it won't be reflected in the bank module, but the transfer within the EVM state still occurs. For example, in [this code](https://github.com/Uniswap/v2-periphery/blob/master/contracts/UniswapV2Router02.sol#L252-L261), if a swap occurs with a fractional amount of `ugas`, there ends up being a mismatch between the two states. On the other hand, ERC20s—whether they're single token represented with the precompile or EVM-native—do not face this issue, since the spec allows you to specify the precision of a token's decimals.

We concluded that in the short term, we'll use Kava's `x/precisebank` and upstream that to this repository in the form of an added module.

For existing chains, this means using `x/precisebank` as a [wrapper for bank](https://github.com/Kava-Labs/kava/blob/c9d0673a58f6f3e67492eabd055788b8de4aab42/app/app.go#L565) going forward. The module would then handle any fractional balances in its own state, but use existing `x/bank` state for the base token. Instructions on how to use it will be included in the docs.

[...]

Issue #33, opened 2025-03-28
Move tests using evmd into emvd


Eric-Warehime:
We should move all of the tests which use the example app (evmd) into the evmd module so that we can re-create a go.mod and make it easier for quick setup of a cosmos evm chain.

This will prevent the circular dependency between evm and evmd.

vladjdk:
@[**facs95**](https://github.com/facs95)** made this PR that allows you to use your App object with the integration tests.** [https://github.com/realiotech/realio-network/pull/241](https://github.com/realiotech/realio-network/pull/241) <br><br>Can be improved by separating interfaces because some modules needed to be mocked:

```
func (n *IntegrationNetwork) GetMintClient() minttypes.QueryClient {
	panic("Mint client not implemented")
}

func (n *IntegrationNetwork) GetPreciseBankClient() precisebanktypes.QueryClient {
	panic("Precise bank client not implemented")
```

PR #69, opened 2025-04-13
feat: x/precisebank module

dudong2:
# Description

**TL;DR:** Introduces `x/precisebank` to address EVM ↔ SDK precision mismatch and integrates it with `x/vm` and `precompiles`.

---

### Problem

- Cosmos SDK assumes a 6-decimal denomination across all modules.
- EVM uses 18 decimals, causing precision loss when transferring values or paying gas.
- Sub-6-decimal values are **floored**, which can lead to:
  - Loss of small token amounts
  - **Arbitrage vulnerabilities** in DEXs like Uniswap v2 navite coin -> erc20 senarios due to mismatched balance handling

---

### Purpose

- Enable precise handling of fractional balances in EVM-related modules.
- Preserve and manage sub-integer values (e.g., 0.0000001 (= 10^-7) tokens) without loss.
- Seamlessly integrate precise value tracking into `x/vm` and precompiles.

---

### Implemented Changes

- Introduced a new `x/precisebank` module based on Kava’s implementation.
- **Generalized to support 1~18 decimal assets**, not just 6↔18.
- Added **carry/borrow mechanics** for converting fractional to integer coins.
- Replaced direct use of `BankKeeper` with `PreciseBankKeeper` in:
  - x/vm
  - precompiles
- Added invariant check for total supply
- Added **CLI queries** for fractional balances and remainder state.
- Added **tests**:
  - Edge case handling
  - Fuzz tests (Mint, Burn, Send, Genesis validation)
  - Random value tests
  - Contract tests(simple wrap/unwarp)

---

### Notes

- For tokens with exactly 18 decimals, using `x/precisebank` is supported but not necessary — the standard `x/bank` module is sufficient and recommended.
- The `x/precisebank` module is designed to manage only the fractional (sub-integer) balances, while integer balances continue to be managed by the `x/bank` module..
- Staking, slashing, distribution, and other Cosmos SDK modules remain unchanged:
  - They all assume 6 decimals and are **out of scope** for this PR.

Closes: #48 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [x] added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [x] confirmed all author checklist items have been addressed
- [x] confirmed that this PR does not change production code
- [x] reviewed content
- [x] tested instructions (if applicable)
- [x] confirmed all CI checks have passed


evmd/app.go
```diff
@@ -495,13 +505,20 @@ func NewExampleApp(
...
+	)
+
 	// NOTE: it's required to set up the EVM keeper before the ERC-20 keeper, because it is used in its instantiation.
 	app.EVMKeeper = evmkeeper.NewKeeper(
 		// TODO: check why this is not adjusted to use the runtime module methods like SDK native keepers
 		appCodec, keys[evmtypes.StoreKey], tkeys[evmtypes.TransientKey],
 		authtypes.NewModuleAddress(govtypes.ModuleName),
 		app.AccountKeeper,
-		app.BankKeeper,
+		app.PreciseBankKeeper,
```
dongsam, line 503: - Depending on the situation of each chain, how about managing it with a variable so that we can atomically decide and assign whether to use BankKeeper or PreciseBankKeeper, or give good reference?
- e.g., If the SDK’s precision is already 18, there is no need to use PreciseBankKeeper.

```go
var bankKeeperForEvm evmtypes.BankKeeper
if decimal == 18 {
    bankKeeperForEvm = app.BankKeeper
} else {
    bankKeeperForEvm = PreciseBankKeeper
}
// ...
app.EVMKeeper = evmkeeper.NewKeeper(
    appCodec, keys[evmtypes.StoreKey], tkeys[evmtypes.TransientKey],
    authtypes.NewModuleAddress(govtypes.ModuleName),
    app.AccountKeeper,
    **bankKeeperForEvm**,
    app.StakingKeeper,
    app.FeeMarketKeeper,
    &app.Erc20Keeper,
    tracer,
)
// ...

app.EVMKeeper.WithStaticPrecompiles(
		NewAvailableStaticPrecompiles(
			*app.StakingKeeper,
			app.DistrKeeper,
                         **bankKeeperForEvm,**
			app.Erc20Keeper,
			app.AuthzKeeper,
			app.TransferKeeper,
			app.IBCKeeper.ChannelKeeper,
			app.EVMKeeper,
			app.GovKeeper,
			app.SlashingKeeper,
			app.EvidenceKeeper,
		),
	)
// ...	
```

x/precisebank/keeper/view.go
```diff
@@ -0,0 +1,101 @@
...
+	// Module balance should display as empty for extended denom. Module
+	// balances are **only** for the reserve which backs the fractional
+	// balances. Returning the backing balances if querying extended denom would
+	// result in a double counting of the fractional balances.
+	if denom == types.ExtendedCoinDenom && addr.Equals(k.ak.GetModuleAddress(types.ModuleName)) {
+		return sdk.NewCoin(denom, sdkmath.ZeroInt())
+	}
+
+	// Pass through to x/bank for denoms except ExtendedCoinDenom
+	if denom != types.ExtendedCoinDenom {
```
dongsam: `types.ExtendedCoinDenom` is hardcoded as `aatom`, which seems dangerous to use.
I suspect the original intention might have been to use `types.GetEVMCoinDenom()` instead?

If another app chain uses `precisebnak` module for EVM bank without properly configuring these unclear settings, it could lead to unintended behavior or errors.

https://github.com/cosmos/evm/blob/144cd489944081a765479b7518bfbd90cf5f5554/x/precisebank/types/fractional_balance.go#L14-L25

https://github.com/cosmos/evm/blob/144cd489944081a765479b7518bfbd90cf5f5554/x/vm/types/denom_config.go#L56-L59

https://github.com/cosmos/evm/blob/144cd489944081a765479b7518bfbd90cf5f5554/evmd/app.go#L140-L142

Currently, both assigning values to `precisebanktypes.*` in app.go’s init function and using `evmCoinInfo` coexist.
To prevent mistakes, it seems better to unify these into a single approach.

If this issue is confirmed, the above code is just one example, and we need to conduct a comprehensive review, refactor, and unify usage across the entire codebase, not just in this location.

Alternatively, if this was done due to dependencies between modules,
how about defining it at the type level in the root directory instead?

x/vm/types/scaling.go
```diff
@@ -9,23 +9,6 @@ import (
...
-
-	evmCoinDecimal := GetEVMCoinDecimals()
-	newAmount := coin.Amount.Mul(evmCoinDecimal.ConversionFactor())
-
-	return sdk.Coin{Denom: coin.Denom, Amount: newAmount}
-}
-
 // ConvertAmountToLegacy18Decimals convert the given amount into a 18 decimals
 // representation.
 func ConvertAmountTo18DecimalsLegacy(amt sdkmath.LegacyDec) sdkmath.LegacyDec {
```
dongsam, line 14: How about adding unit tests for `ConvertAmountTo18DecimalsLegacy` and `ConvertAmountTo18DecimalsBigInt`, although this is indirectly covered in the other tests, it’s still a core function and internal logic changed by `ConversionFactor`

x/vm/types/scaling.go
```diff
@@ -79,27 +59,10 @@ func ConvertCoinsFrom18Decimals(coins sdk.Coins) sdk.Coins {
 	convertedCoins := make(sdk.Coins, len(coins))
 	for i, coin := range coins {
 		if coin.Denom == evmDenom {
-			evmCoinDecimals := GetEVMCoinDecimals()
-
-			newAmount := coin.Amount.Quo(evmCoinDecimals.ConversionFactor())
-
-			coin = sdk.Coin{Denom: coin.Denom, Amount: newAmount}
+			evmExtendedDenom := GetEVMCoinExtendedDenom()
+			coin = sdk.Coin{Denom: evmExtendedDenom, Amount: coin.Amount}
```
dongsam: The logic for actual amount handling has been moved to `preciseBank`, so `ConvertEvmCoinFrom18Decimals` and `ConvertCoinsFrom18Decimals` now only perform denom changes.
However, the function names still use `Convert`, and since the other above `Convert*` functions actually perform amount conversions, this breaks consistency and could lead to confusion or mistakes.

x/precisebank/keeper/invariants.go
```diff
@@ -0,0 +1,253 @@
...
+	sdkmath "cosmossdk.io/math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// RegisterInvariants registers the x/precisebank module invariants
+func RegisterInvariants(
+	ir sdk.InvariantRegistry,
+	k Keeper,
+	bk types.BankKeeper,
```
dongsam: I think `bk types.BankKeeper` is unused

x/precisebank/module.go
```diff
@@ -0,0 +1,165 @@
...
+// ConsensusVersion defines the current module consensus version.
+const ConsensusVersion = 1
+
+var (
+	_ module.AppModule      = AppModule{}
+	_ module.AppModuleBasic = AppModuleBasic{}
+	_ module.HasABCIGenesis = AppModule{}
+
+	_ appmodule.AppModule   = AppModule{}
+	_ module.HasABCIGenesis = AppModule{}
```
dongsam, line 34: An invariant has been added to this module. Would it make sense to also add `HasInvariants`? [ref. bank](https://github.com/cosmos/cosmos-sdk/blob/98164408932ee74369b7ce18aa13bb1a1a6ba76b/x/bank/module.go#L44)

dongsam (commented):
Nice work 👍 

I haven’t completed a full review yet, but I’m leaving a note on the key comments so far, as other reviewers may also come across them.
I’ll continue with a second round of review and provide further comments afterward.

--- 

**About License/Copyright**
It may be necessary to review whether a notice is required to comply with the [Apache License or Copyright](https://github.com/Kava-Labs/kava?tab=Apache-2.0-1-ov-file#readme) for the file copied from Kava’s preciseBank module

**JSON-RPC Integration Test**
Although the code for the [GetBalance endpoint](https://github.com/cosmos/evm/blob/144cd489944081a765479b7518bfbd90cf5f5554/rpc/backend/account_info.go#L148) in JSON-RPC hasn’t changed, the internal logic now uses [precisebank as wrapped bank](https://github.com/cosmos/evm/blob/144cd489944081a765479b7518bfbd90cf5f5554/x/vm/keeper/keeper.go#L256).
Adding an integration test or including local test results could be helpful in verifying this behavior.
While the internal logic appears to be well covered in view_test.go, it would be beneficial to validate this change specifically in the context of JSON-RPC to catch any unexpected issues during integration.

x/precisebank/keeper/invariants.go
dudong2: I removed it.
ref. https://github.com/cosmos/evm/pull/69/commits/783f5407f6c174cfc4356e205039aac524f6bacd

x/precisebank/module.go
dudong2: I added it.
ref. https://github.com/cosmos/evm/pull/69/commits/783f5407f6c174cfc4356e205039aac524f6bacd

evmd/app.go
dudong2: I think it's enough to add a comments. Also, using precisebank module in the case of 18 decimals does not cause any problems. I added a comment to the precisebank keeper definition.
ref. https://github.com/cosmos/evm/pull/69/commits/619e6eccc03626f7183e8b4effe4487332007891

x/precisebank/keeper/view.go
dudong2: I removed that variables in https://github.com/cosmos/evm/pull/69/commits/ee976def9dc6befbb0b17a216016a5c41efd58bb.
But I'm not sure if this is the right way to do it, because then I would be setting evm coin info using the value in client.toml instead of the --chain-id option via the cli. I would like to hear your opinion.

x/vm/types/scaling.go
dudong2: I added tests for those util funcs.
ref. https://github.com/cosmos/evm/pull/69/commits/9516f672b609b39a32236b7ae1148f04cca8c48a

x/vm/types/scaling.go
dudong2: I renamed that funcs to `ChangeEvmCoinDenomFrom18Decimals` and `ChangeCoinsDenomFrom18Decimals`.
ref. https://github.com/cosmos/evm/pull/69/commits/9516f672b609b39a32236b7ae1148f04cca8c48a

x/precisebank/keeper/send.go
```diff
@@ -0,0 +1,410 @@
...
+	// If we do not return early here, the following issue occurs:
+	// - `senderNewFracBal` will be calculated by subtracting fractional `amt` from the sender's fractional balance.
+	// - `recipientNewFracBal` will be calculated by adding fractional `amt` to the recipient's fractional balance.
+	// - Since `from` and `to` are the same address, calling SetFractionalBalance(from, ...) and SetFractionalBalance(to, ...)
+	//   would overwrite the fractional balance with the recipientNewFracBal value.
+	// - As a result, the subtraction of fractional amount is lost, and it would *artificially inflate* the balance
+	//   by the fractional amount, effectively *duplicating* the fractional value.
+	//
+	// By returning early here, we ensure that no unintended state mutation happens when transferring to self.
+	if from.Equals(to) {
```
vladjdk, line 139: good catch

vladjdk: what did you do to find this?

x/precisebank/keeper/send.go
```diff
@@ -0,0 +1,410 @@
...
+	// Check balance is sufficient
+	bal := k.GetBalance(ctx, addr, types.ExtendedCoinDenom())
+	coin := sdk.NewCoin(types.ExtendedCoinDenom(), amt)
+
+	// TODO: This checks spendable coins and returns error with spendable
+	// coins, not full balance. If GetBalance() is modified to return the
+	// full, including locked, balance then this should be updated to deduct
+	// locked coins.
+
+	spendable := sdk.Coins{bal}
```
vladjdk, line 403: Why did we change this from sdk.NewCoins?

x/precisebank/keeper/send.go
```diff
@@ -0,0 +1,410 @@
...
+	if senderAddr == nil {
+		panic(errorsmod.Wrapf(sdkerrors.ErrUnknownAddress, "module account %s does not exist", senderModule))
+	}
+
+	recipientAcc := k.ak.GetModuleAccount(ctx, recipientModule)
+	if recipientAcc == nil {
+		panic(errorsmod.Wrapf(sdkerrors.ErrUnknownAddress, "module account %s does not exist", recipientModule))
+	}
+
+	if recipientModule == types.ModuleName {
```
vladjdk, line 372: differs from x/bank on the main branch, and this logic would also be covered in https://github.com/cosmos/evm/blob/9516f672b609b39a32236b7ae1148f04cca8c48a/x/precisebank/keeper/send.go#L130-L141

i think we can just remove this to match bank

x/precisebank/keeper/view.go
```diff
@@ -0,0 +1,101 @@
...
+		Add(fractionalAmount)
+
+	return sdk.NewCoin(types.ExtendedCoinDenom(), fullAmount)
+}
+
+func (k Keeper) IterateAccountBalances(ctx context.Context, account sdk.AccAddress, cb func(coin sdk.Coin) bool) {
+	k.bk.IterateAccountBalances(ctx, account, cb)
+}
+
+func (k Keeper) GetAllBalances(ctx context.Context, addr sdk.AccAddress) sdk.Coins {
```
vladjdk: We should add some tests for this since there's custom logic that isn't covered

x/precisebank/keeper/view_integration_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	for _, tt := range tests {
+		suite.Run(tt.name, func() {
+			suite.SetupTest()
+
+			addr := sdk.AccAddress([]byte("test-address"))
+
+			suite.MintToAccount(addr, tt.giveBankBal)
+
```
vladjdk, line 76: Noticed we get rid of these lines: https://github.com/Kava-Labs/kava/blob/c9d0673a58f6f3e67492eabd055788b8de4aab42/x/precisebank/keeper/view_integration_test.go#L101-L120

Should we add these back and instead include the locked vesting accounts in our network? Some networks have this, so it could still be important to test

x/precisebank/module.go
```diff
@@ -0,0 +1,166 @@
...
+	cdctypes "github.com/cosmos/cosmos-sdk/codec/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+)
+
+// ConsensusVersion defines the current module consensus version.
+const ConsensusVersion = 1
+
+var (
+	_ module.AppModule      = AppModule{}
```
vladjdk, line 29: Deprecated - let's remove this since we have `appmodule.AppModule` and `module.AppModuleBasic` already.

x/precisebank/client/cli/query.go
```diff
@@ -0,0 +1,117 @@
...
+		GetRemainderCmd(),
+		GetFractionalBalanceCmd(),
+	)
+	return cmd
+}
+
+// GetTotalFractionalBalancesCmd queries the total sum of all fractional balances
+func GetTotalFractionalBalancesCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "total-fractional-balances",
```
vladjdk: This seems to iterate over the entire store, so I imagine that this would be quite heavy computationally. I'm not sure that the TotalFractionalBalances is useful anywhere other than validation. I don't think we should expose this as a query at all.

x/precisebank/README.md
```diff
@@ -0,0 +1,676 @@
+# `x/precisebank`
+
```
vladjdk, line 2:
```suggestion
`Big thanks to the [Kava](https://www.kava.io/) team for their valuable contributions to this module.`
```

x/precisebank/keeper/integration_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+			// Check WATOM balance == wrapAmount
+			bal, err := utils.GetERC20Balance(suite.network, watomAddr, sender.Addr)
+			suite.Require().NoError(err)
+			suite.Require().Equal(amount.Cmp(bal), 0, "WATOM balance should match deposited amount (expected: %s, actual: %s)", amount, bal)
+
+			baseFeeRes, err = suite.network.GetEvmClient().BaseFee(suite.network.GetContext(), &evmtypes.QueryBaseFeeRequest{})
+			suite.Require().NoError(err)
+
+			fmt.Println(suite.network.App.PreciseBankKeeper.GetAllBalances(suite.network.GetContext(), watomAddr.Bytes()))
```
vladjdk: Nit - Residual print

x/vm/types/denom_config.go
```diff
@@ -49,16 +58,30 @@ func GetEVMCoinDenom() string {
...
+
 // setEVMCoinInfo allows to define denom and decimals of the coin used in the EVM.
 func setEVMCoinInfo(eci EvmCoinInfo) error {
 	if evmCoinInfo != nil {
 		return errors.New("EVM coin info already set")
 	}
 
+	if eci.Decimals == EighteenDecimals {
+		if eci.Denom != eci.ExtendedDenom {
+			return errors.New("EVM coin denom and extended denom must be the same for 18 decimals")
```
vladjdk, line 74: Doesn't seem like there's anything that enforces this in `x/precisebank`. Unless I somehow missed that, we should add in a check for genesis.

x/vm/types/scaling.go
vladjdk: Btw, do we still need the legacy function? I checked, and everywhere it's used, it gets converted to a big.Int anyways. Unless we need to retain the truncation that happens when it gets used, could we just remove it?

Example ref: https://github.com/cosmos/evm/blob/a828f95ffc3e8c636a07e657c380eaf26f2123e0/x/vm/wrappers/feemarket.go#L38

x/vm/types/scaling.go
```diff
@@ -58,48 +33,26 @@ func ConvertBigIntFrom18DecimalsToLegacyDec(amt *big.Int) sdkmath.LegacyDec {
...
+func ChangeEvmCoinDenomFrom18Decimals(coin sdk.Coin) (sdk.Coin, error) {
 	if coin.Denom != GetEVMCoinDenom() {
 		return sdk.Coin{}, fmt.Errorf("expected coin denom %s, received %s", GetEVMCoinDenom(), coin.Denom)
 	}
 
-	evmCoinDecimal := GetEVMCoinDecimals()
-	newAmount := coin.Amount.Quo(evmCoinDecimal.ConversionFactor())
-
-	return sdk.Coin{Denom: coin.Denom, Amount: newAmount}, nil
+	return sdk.Coin{Denom: GetEVMCoinExtendedDenom(), Amount: coin.Amount}, nil
```
vladjdk, line 43: Doesn't look like we're actually doing anything with the amount here, but the comment above says we are?

ante/evm/04_validate.go
```diff
@@ -110,11 +110,9 @@ func CheckTxFee(txFeeInfo *tx.Fee, txFee *big.Int, txGasLimit uint64) error {
 		return nil
 	}
 
-	convertedAmount := sdkmath.NewIntFromBigInt(evmtypes.ConvertAmountFrom18DecimalsBigInt(txFee))
-
-	baseDenom := evmtypes.GetEVMCoinDenom()
-	if !txFeeInfo.Amount.AmountOf(baseDenom).Equal(convertedAmount) {
-		return errorsmod.Wrapf(errortypes.ErrInvalidRequest, "invalid AuthInfo Fee Amount (%s != %s)", txFeeInfo.Amount, convertedAmount)
+	evmExtendedDenom := evmtypes.GetEVMCoinExtendedDenom()
```
vladjdk, line 117: Does this mean that we now have to set `aatom` as the gas token? What happens if a chain has `uatom` as their gas token, but this expects the txFeeInfo.Amount to be denominated in `aatom`?

Since most chains don't support multiple gas tokens, how would this affect dapps and wallets? Do they need to change the denom and decimals that nodes and applications use for transaction generation?

vladjdk: If this is the case, then we either need to work on a way for nodes to support multiple gas tokens (original udenom and new adenom) or we should keep this function as it was and floor gas. I don't think there's any vulnerability that can be abused here if we choose to floor.

ante/testutils/testutil.go
```diff
@@ -103,13 +103,19 @@ func (suite *AnteTestSuite) SetupTest() {
...
+	decimals := evmtypes.GetEVMCoinDecimals()           //nolint:staticcheck
+
 	configurator := evmtypes.NewEVMConfigurator()
 	configurator.ResetTestConfig()
 	err := configurator.
 		WithChainConfig(chainConfig).
-		WithEVMCoinInfo(denom, uint8(decimals)).
+		WithEVMCoinInfo(evmtypes.EvmCoinInfo{
+			Denom:         denom,
+			ExtendedDenom: extendedDenom,
```
vladjdk, line 116: To follow up on the above, it seems like we're making the assumption that chain's will be using their extendedDenom for gas now, which I don't think is necessarily what they'll want to do.

evmd/app.go
vladjdk: I think that this is fine for the sake of the example app. Other apps can just choose not to use the PreciseBank at all. It'll be sufficient to add comments and docs in the integration to explain the options.

contracts/solidity/WATOM.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MIT
```
vladjdk, line 1: If we only use this for testing, we should move this to a testing folder - reminder for #33

contracts/watom.go
```diff
@@ -0,0 +1,27 @@
+package contracts
```
vladjdk, line 1: Same as above

x/vm/types/scaling.go
```diff
@@ -58,48 +33,26 @@ func ConvertBigIntFrom18DecimalsToLegacyDec(amt *big.Int) sdkmath.LegacyDec {
...
 }
 
 // ConvertCoinsFrom18Decimals returns the given coins with the Amount of the evm
```
vladjdk: Like the above comment, seems like we're not converting the amount at all, and we're only changing the denom

x/vm/types/scaling.go
```diff
@@ -58,48 +33,26 @@ func ConvertBigIntFrom18DecimalsToLegacyDec(amt *big.Int) sdkmath.LegacyDec {
...
 // coin converted from the 18 decimals representation to the original one.
-func ConvertCoinsFrom18Decimals(coins sdk.Coins) sdk.Coins {
+func ChangeCoinsDenomFrom18Decimals(coins sdk.Coins) sdk.Coins {
```
vladjdk: Like I mentioned in the ante comments, I don't think this is the behaviour that we want for gas. The JSON-RPC uses this function, and if it doesn't convert the decimals down to `uatom`, then we basically won't support `uatom` anymore, and chains will have to switch to `aatom`, even for SDK transactions. We essentially don't have a way for supporting both tokens for gas when we don't convert the decimals back down to 6 here.

x/vm/wrappers/bank_test.go
```diff
@@ -21,27 +22,28 @@ import (
...
+		mockSetup func(*testutil.MockBankWrapper)
 	}{
 		{
-			name:        "success - convert 18 decimals amount to 6 decimals",
-			evmDenom:    TokenDenom,
-			evmDecimals: 6,
-			amount:      big.NewInt(1e18), // 1 token in 18 decimals
-			recipient:   sdk.AccAddress([]byte("test_address")),
-			expectErr:   "",
+			name:      "success - convert 18 decimals amount to 6 decimals",
```
vladjdk: I don't see how this test case is still converting 18 -> 6

x/vm/wrappers/bank_test.go
```diff
@@ -150,24 +128,25 @@ func TestMintAmountToAccount(t *testing.T) {
...
+		expectErr string
+		mockSetup func(*testutil.MockBankWrapper)
 	}{
 		{
-			name:        "success - convert 18 decimals amount to 6 decimals",
-			evmDenom:    TokenDenom,
-			evmDecimals: 6,
-			amount:      big.NewInt(1e18),
-			expectErr:   "",
+			name:      "success - convert 18 decimals amount to 6 decimals",
```
vladjdk: don't think we're converting anymore since we now have precisebank

x/vm/wrappers/bank_test.go
```diff
@@ -309,19 +286,18 @@ func TestSendCoinsFromModuleToAccount(t *testing.T) {
 			},
 		},
 		{
-			name:        "success - convert 18 decimals amount to 6 decimals with single token",
-			evmDenom:    TokenDenom,
-			evmDecimals: 6,
+			name:     "success - convert 18 decimals amount to 6 decimals with single token",
```
vladjdk: same here - seems like we're no longer converting

x/vm/wrappers/bank_test.go
vladjdk: Looks like w'ere converting the other way around. The denom switches from `u` to `a`

x/vm/wrappers/bank_test.go
```diff
@@ -361,20 +336,19 @@ func TestSendCoinsFromModuleToAccount(t *testing.T) {
 			},
 		},
 		{
-			name:        "success - convert 18 decimals amount to 6 decimals with multiple tokens",
-			evmDenom:    TokenDenom,
-			evmDecimals: 6,
+			name:     "success - convert 18 decimals amount to 6 decimals with multiple tokens",
```
vladjdk: Same as above

x/vm/wrappers/bank_test.go
```diff
@@ -473,19 +447,18 @@ func TestSendCoinsFromAccountToModule(t *testing.T) {
 			},
 		},
 		{
-			name:        "success - convert 18 decimals amount to 6 decimals with single token",
-			evmDenom:    TokenDenom,
-			evmDecimals: 6,
+			name:     "success - convert 18 decimals amount to 6 decimals with single token",
```
vladjdk: Same as above

x/vm/wrappers/bank_test.go
```diff
@@ -525,20 +497,19 @@ func TestSendCoinsFromAccountToModule(t *testing.T) {
 			},
 		},
 		{
-			name:        "success - convert 18 decimals amount to 6 decimals with multiple tokens",
-			evmDenom:    TokenDenom,
-			evmDecimals: 6,
+			name:     "success - convert 18 decimals amount to 6 decimals with multiple tokens",
```
vladjdk: Same as above, looks to be 6->18

x/precisebank/keeper/view.go
dongsam: I confirmed that the update has been applied and coin info is now unified and managed through a single variable 👍 

But I’m not fully clear on why this change would affect `client.toml` and `--chain-id`. From what I can see, `initClientCtx` internally calls `WithChainID(flags.FlagChainID)` through `ReadPersistentCommandFlags`, so is there an issue I’m missing or something fundamentally unsolvable here?

If there’s any concern, I think it’s a good idea to leave this comment open for now and get input from other reviewers as well.

vladjdk (changes requested):
Thanks for the work!

I left a few comments - few nits, and some questions.

My main concern is that we seem to no longer be converting the gas to the base chain's denom. e.g. If a chain is 6 decimals, it looks like we now require the gas to be 18 decimals. The txBuilder currently converts the decimals, but we changed the function to convert the denom to atto units. This means that we expect transactions to submit an atto unit tx fee. In this case, we'd also need to submit gas for other Cosmos transactions in the atto fee. For chains that currently run 6 decimals, the implications here are pretty big. They'd need to change their base fees on their nodes, inside wallets, and the way dapps generate/estimate fees. 

My recommendation would be to keep the conversion for gas as it was. Since we were originally implementing precisebank to solve for data loss in partial gas token sending on the EVM itself, it doesn't seem like gas would cause an issue. I don't agree with the implied assumption that chains will be willing to change their base gas token for everything that depends on it. I think we should instead open another issue where we support multiple gas tokens. Then, we can support atto token gas on the EVM side, but keep micro token gas support on the SDK side.

Please correct me if I'm understanding this wrong, and let me know your thoughts.

x/precisebank/keeper/send.go
vladjdk: I think we should also let Kava know that this is an issue in their repo

dudong2: I found it while working on the test code, I'll let the kava team know.

x/precisebank/keeper/send.go
dudong2: The kava code was based on cosmos-sdk v0.47 and used sdk.NewCoins to remove coins with amount 0. However, I checked cosmos-sdk v0.50 and it was modified to not remove coins with amount 0, so I changed it to use sdk.Coins.

ref. https://github.com/cosmos/cosmos-sdk/blob/v0.50.13/x/bank/keeper/send.go#L276-L283

x/precisebank/module.go
dudong2: i removed it, thanks!
ref. https://github.com/cosmos/evm/pull/69/commits/b92fb93dda4ee4217d86bd88609446849aeca21b

x/precisebank/README.md
dudong2: i added it.
ref. https://github.com/cosmos/evm/pull/69/commits/4ebbe5d320f946dc342c5ee4d45ff2b118d4c8e1

x/precisebank/keeper/integration_test.go
dudong2: i removed it!
ref. https://github.com/cosmos/evm/pull/69/commits/3569ed3078b365dd977d3bed002b33ade425368b

x/precisebank/keeper/send.go
dudong2: If we allow the transfer of funds to the precisebank module, it may affect the remainder logic in the mint, burn logic. The burn, mint is called after the transfer of assets to the evm module. So the check is correct and intended.

x/precisebank/client/cli/query.go
dudong2: yes, i agree. In addition to providing it through the CLI, it can always query it via GRPC, so i've removed it entirely. 
ref. https://github.com/cosmos/evm/pull/69/commits/c4c2a77656860a7fedce59d1817786288a1f6e0c

x/vm/types/denom_config.go
dudong2: If this assumption does not exist, then all converting denom logics would need an if condition for the case of 18 decimal. By adding a condition that requires denom and extended denom to be equal when 18 decimal, we can generalize all conversion operations.
For coin info, it is not appropriate to put it in genesis because it is not data that goes into genesis. i think the current location is appropriate.

contracts/solidity/WATOM.sol
dudong2: I agree. However, we are still unclear on how to structure the inside of tests. Also, since tests/solidity/* uses truffle, it would be pretty unnatural to move the contracts folder using hardhat. I think it would be better to add it as is in the current PR and move it all at once when we move forward with #33.

contracts/watom.go
dudong2: [ditto](https://github.com/cosmos/evm/pull/69#discussion_r2068115600)

ante/evm/04_validate.go
dudong2: When an evm tx comes in, it goes through the process of [converting](https://github.com/cosmos/evm/blob/c4c2a77656860a7fedce59d1817786288a1f6e0c/rpc/backend/call_tx.go#L126-L132) it to `MsgEthereumTx`, which is a cosmos tx. Here, the denom will be a `uatom`, not `aatom`. [BuildTx](https://github.com/cosmos/evm/blob/c4c2a77656860a7fedce59d1817786288a1f6e0c/x/vm/types/msg.go#L342-L347) then converts `uatom` to `aatom` meaning that logic that interacts with the user will use `uatom` and internal processing such as the ante handler will operate based on `aatom`.

ante/testutils/testutil.go
dudong2: As explained above, gas was designed to use uatom from user side. So I think the current design makes sense.

x/vm/types/scaling.go
dudong2: I chaned the comments and function name.
- ConvertEvmCoinDenomToExtendedDenom -> ConvertEvmCoinDenomToExtendedDenom
- ConvertCoinsDenomToExtendedDenom -> ConvertCoinsDenomToExtendedDenom

I think it is proper.

ref. https://github.com/cosmos/evm/pull/69/commits/7f651ae2929abcf095aa7e17e2fdc36d574931c7

x/vm/types/scaling.go
dudong2: please reference https://github.com/cosmos/evm/pull/69#discussion_r2068181132! I think that would be proper answer.

x/vm/wrappers/bank_test.go
dudong2: I renamed it.
convert 18 decimals amount to 6 decimals -> convert evm coin denom to extended denom
ref. https://github.com/cosmos/evm/pull/69/commits/9b9f9b1dae040f2af16f7b7741565cbc95b03f1f

x/vm/wrappers/bank_test.go
dudong2: [ditto](https://github.com/cosmos/evm/pull/69#discussion_r2068352936)

x/vm/wrappers/bank_test.go
dudong2: [ditto](https://github.com/cosmos/evm/pull/69#discussion_r2068352936)

x/vm/wrappers/bank_test.go
dudong2: [ditto](https://github.com/cosmos/evm/pull/69#discussion_r2068352936)

x/vm/wrappers/bank_test.go
dudong2: [ditto](https://github.com/cosmos/evm/pull/69#discussion_r2068352936)

x/vm/wrappers/bank_test.go
dudong2: [ditto](https://github.com/cosmos/evm/pull/69#discussion_r2068352936)

x/vm/types/scaling.go
dudong2: I chaned the comments and function name.
- ConvertEvmCoinDenomToExtendedDenom -> ConvertEvmCoinDenomToExtendedDenom
- ConvertCoinsDenomToExtendedDenom -> ConvertCoinsDenomToExtendedDenom

I think it is proper.

ref. https://github.com/cosmos/evm/pull/69/commits/7f651ae2929abcf095aa7e17e2fdc36d574931c7

x/precisebank/keeper/view_integration_test.go
dudong2: I added vesting module to app.go and restored that test!
ref. https://github.com/cosmos/evm/pull/69/commits/ac0e1d0ee54da13c6c8431d99d57687b93949100

x/precisebank/keeper/view.go
dudong2: I noticed that the `GetAllBalances` function, which was originally defined in the bank keeper interface of the x/vm module and implemented in precisebank, is not being used anywhere in the codebase. Since it's unused, I removed the GetAllBalances function from both precisebank and the bank keeper interface in x/vm.

ref. https://github.com/cosmos/evm/pull/69/commits/ca2a17b760a6c43dbd6f1babc32d05f9909542f0

x/precisebank/keeper/view.go
dudong2: The current change is reading the `client.toml` via `ReadFromClientConfig`, setting the chain id and calling EvmAppOptions. And if you look at the internal implementation of EvmAppOptions, once it is set via the seal variable, it cannot be set again. This means that after making this change, the option `--chain-id` will not be available when setting up the `evm configurator`.

ante/evm/04_validate.go
vladjdk: Okay, I did some testing, and it looks like atto gas and micro gas is converting properly on both ends. Tested both Ethereum transactions and Cosmos transactions with the Feemarket and base denom set up to be 0.0001utest.

However, in a future issue, we really need to set up E2E testing to test gas specifically. It was pretty hard to follow along with the code to see if both tokens are converting to the same gas fee amount in the end. 

Fine to do that when we start our testing phase.

vladjdk (approved):
Okay, I did some testing, and it looks like atto gas and micro gas are converting properly on both ends. Tested both Ethereum transactions and Cosmos transactions with the Feemarket and base denom set up to be 0.0001utest. Confirming that after a manual check, price/gas 

However, in a future issue, we really need to set up E2E testing to test gas from an external perspective. It was pretty hard to follow along with the code to see if both tokens are converting to the same gas fee amount in the end.

Nothing specific to your change, though. We're fine to make testing improvements when we start our testing phase.

Thanks for the work :)

x/precisebank/keeper/send.go
vladjdk: notified as well, resolving this convo

PR #84, opened 2025-04-16
feat!: permissionless erc20 registration

vladjdk:
ref: #17

depends on: #85 

reopen of #37


x/erc20/keeper/proposals_test.go
```diff
@@ -166,7 +166,7 @@ func (suite *KeeperTestSuite) TestRegisterERC20() {
 			tc.malleate()
 
 			_, err = suite.network.App.Erc20Keeper.RegisterERC20(ctx, &types.MsgRegisterERC20{
-				Authority:      authtypes.NewModuleAddress("gov").String(),
+				Signer:         authtypes.NewModuleAddress("gov").String(),
```
Eric-Warehime: Can we test this when the governance module isn't the signer so we actually validate the permissionless case?

Eric-Warehime (approved):

Issue #70, opened 2025-04-14
[Feature]: x/erc20 Allowance state for erc20 precompile


cloudgray:
### Summary

Introduce an `Allowance` state and related methods for the ERC20 precompile within the x/erc20 module.

### Problem Definition

The approach of implementing the ERC20 precompile’s allowance state by leveraging the Cosmos SDK `x/authz` module’s `Grant` state was not fully compatible from an interoperability perspective. Additionally, it imposed a requirement on chains that do not use the x/authz module to adopt it unnecessarily. (See: Issue #47)

### Key

```protobuf
KeyPrefixAllowance + erc20 + owner + spender
```

- `KeyPrefixAllowance`: []byte{4}
- `erc20`: erc20 precompile address
- `owner`: owner address
- `spender`: spender address

**Consideration**

- `AllowancePrefix`: The current last key prefix, `KeyPrefixSTRv2Addresses`, has a value of `[]byte{3}`, and the next value to be used is `[]byte{4}`.
- `erc20`: The reason erc20 comes first is to allow for easy retrieval and update of a specific contract’s Allowance state from the KVStore in case state migration is triggered by an upgrade handler for any reason.

### Value

```protobuf
//Allowance is a token allowance
message Allowance {
  option (gogoproto.equal) = false;

  // value specifies the maximum amount of tokens that can be spent
  // by this token allowance and will be updated as tokens are spent.
  string value = 1 [
    (cosmos_proto.scalar) = "cosmos.Int",
    (gogoproto.customtype) = "cosmossdk.io/math.Int",
    (gogoproto.nullable) = false
  ];
}
```

**Consideration**

- The erc20 precompile address, owner, and spender are all components used to construct the key, and since the denom can be derived from the erc20 precompile address, **only the necessary information — the value — is kept inside the Allowance type**.

## Keeper Methods

> These keeper methods are **only for erc20 precompile**. It is not exposed as msgServer or grpcQuery message on cosmos-sdk side. If there is a compelling reason to call these methods specifically as Cosmos SDK messages, modifications might be made in the future. However, at this point, it does not seem essential.

### GetAllowance

**Input**

| Name | Type | Description |
| --- | --- | --- |
| ctx | sdk.Context | - |
| erc20 | common.Address | evm address of erc20 precompile |
| owner | common.Address | evm address of owner |
| spender | common.Address | evm address of spender |

**Output**

| Type | Description |
| --- | --- |
| Allowance | - |
| error | - |

**Consideration**

[...]

PR #90, opened 2025-04-17
feat(x/erc20): add allowance state

cloudgray:
# Description

- Add `Allowance` state in x/erc20 module

Closes: #70 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


proto/cosmos/evm/erc20/v1/erc20.proto
```diff
@@ -32,6 +32,27 @@ message TokenPair {
...
+  option (gogoproto.equal) = false;
+
+  // erc20_address is the hex address of ERC20 contract
+  string erc20_address = 1;
+
+  // owner is the address of the owner account
+  string owner = 2;
+
+  // spender is the address that is allowed to spend the allowance
+  string spender = 3;
```
zsystm: @cloudgray 
Would it be possible to add a comment indicating the expected address format here — whether it's Bech32 or a 0x-prefixed hex address?

x/erc20/genesis.go
```diff
@@ -31,12 +33,23 @@ func InitGenesis(
...
 	}
+
+	var erc20, owner, spender common.Address
+	var value *big.Int
+	for _, allowance := range data.Allowances {
+		erc20 = common.HexToAddress(allowance.Erc20Address)
+		owner = common.HexToAddress(allowance.Owner)
+		spender = common.HexToAddress(allowance.Spender)
+		value = allowance.Value.BigInt()
+		k.SetAllowance(ctx, erc20, owner, spender, value)
```
zsystm: @cloudgray 
Could you add error handling here? If SetAllowance returns an error, we should panic to prevent inconsistent state.

x/erc20/genesis_test.go
```diff
@@ -29,7 +30,7 @@ type GenesisTestSuite struct {
 	genesis types.GenesisState
 }
 
-const osmoERC20ContractAddr = "0x5dCA2483280D9727c80b5518faC4556617fb19ZZ"
+const osmoERC20ContractAddr = "0x5D87876250185593977a6F94aF98877a5E7eD60E"
```
zsystm, line 33: @cloudgray 
Just curious — is there a specific reason this address was changed?

x/erc20/keeper/allowance.go
```diff
@@ -4,39 +4,160 @@ import (
...
+		var allowance types.Allowance
+		k.cdc.MustUnmarshal(iterator.Value(), &allowance)
+
+		if cb(allowance) {
+			break
+		}
+	}
+}
+
+func (k Keeper) checkAddressesNonZero(
```
zsystm: @cloudgray 
Would it be possible to move this function to a separate types or utils package?
Since it doesn’t rely on any keeper state, it might be cleaner to keep keeper-level functions focused on logic that requires state access. Stateless validations like this are generally better placed in a shared utility layer.

Also, since the verification logic is identical for each address and doesn’t vary by parameter, it might be worth considering accepting a variadic list of addresses (`...common.Address`) to reduce redundancy.

x/erc20/keeper/allowance_test.go
```diff
@@ -0,0 +1,58 @@
...
+	var (
+		ctx       sdk.Context
+		expRes    *big.Int
+		erc20Addr = utiltx.GenerateAddress()
+		owner     = utiltx.GenerateAddress()
+		spender   = utiltx.GenerateAddress()
+		value     = big.NewInt(100)
+	)
+
+	testCases := []struct {
```
zsystm, line 27: @cloudgray 
Could you also add some negative test cases that cover failure scenarios?
Adding some failing cases would help ensure the logic handles edge conditions correctly.

x/erc20/types/genesis.go
```diff
@@ -57,6 +59,25 @@ func (gs GenesisState) Validate() error {
 	if err := validatePrecompiles(gs.TokenPairs, gs.Params.NativePrecompiles); err != nil {
 		return fmt.Errorf("invalid native precompiles on genesis: %w", err)
 	}
+
+	// Check if allowances are valid
```
zsystm, line 63: @cloudgray 
The logic looks good overall.

One minor suggestion: consider using `map[string]struct{}` instead of `map[string]bool` for the seenAllowance map. Since we're only using the map for existence checks, using struct{} avoids allocating a bool value for each key, which can be slightly more memory-efficient in Go.

Not a blocking issue, but might be worth considering for optimization.

zsystm (commented):
@cloudgray 
Thanks for the draft — I’ve left a few comments that might be helpful as you continue working on this.

x/erc20/keeper/allowance_test.go
cloudgray: It is impossible. Because SetAllowance return error for negative allowance, I cannot add testcase for that case.

x/erc20/types/genesis.go
cloudgray: I think `map[string]struct{}` is not good for clarity and consistency.
I think memory-efficiency is less important than clarity and consistency in case of genesisState validation that is occasionally happened.

proto/cosmos/evm/erc20/v1/erc20.proto
cloudgray: applied in https://github.com/cosmos/evm/pull/90/commits/80036a6f2ebf3a98613389a3f13930ba937493f9

x/erc20/genesis.go
cloudgray: applied in https://github.com/cosmos/evm/pull/90/commits/80036a6f2ebf3a98613389a3f13930ba937493f9

x/erc20/genesis_test.go
cloudgray: Looking at the end of previous address, it contains `"Z"` that is not hex letter.
And test case breaks abnormaly because of it.
I think [testutil function](https://github.com/cosmos/evm/blob/feat%2Ferc20-allowance/x/erc20/genesis_test.go#L134-L136) that imports genState for test env does not check validity of genesis state.

x/erc20/keeper/allowance.go
cloudgray: I removed that helper function and just added [simple check conditional statement](https://github.com/cosmos/evm/blob/feat%2Ferc20-allowance/x/erc20/keeper/allowance.go#L77-L82).

cloudgray:
@zsystm I added [deleteAllowances](https://github.com/cosmos/evm/blob/feat%2Ferc20-allowance/x/erc20/keeper/allowance.go#L151-L159) method and let it called by [DeleteTokenPair](https://github.com/cosmos/evm/blob/feat%2Ferc20-allowance/x/erc20/keeper/token_pairs.go#L102) method.

It is not likely to happen that TokenPair for erc20 precompile is deleted and it really does not happen in cosmos/evm, 
because DeleteTokenPair method is public method, appchain that imports cosmos/evm can call it in somewhere.
Then, when that appchain import genesis, the situation where TokenPair does not exists but allowances exist happens.

So, I thought that all allowances that share same erc20 address are deleted when TokenPair is deleted.

I applied this change in https://github.com/cosmos/evm/pull/90/commits/80036a6f2ebf3a98613389a3f13930ba937493f9.
Could you check this change?

x/erc20/types/genesis.go
zsystm: @cloudgray 
Makes sense. I ran a heap profile diff just to quantify the difference between map[string]bool and `map[string]struct{}`, and the actual memory saving turned out to be around 1.6MB (~2.75%) for 1M entries.

Since most of the memory usage comes from the string keys themselves, the impact of using `struct{}` over `bool` is pretty minimal in practice.

Given that this logic only runs during genesis validation and clarity is a valid concern, I'm totally fine sticking with `map[string]bool` here.

For reference, I used the following code and command to test:


<details> <summary>Code</summary>

```go
package main

import (
	"fmt"
	"math/rand"
	"os"
	"runtime/pprof"
	"time"
)

const addressCount = 1_000_000

func randomAddress() string {
	b := make([]byte, 20)
	rand.Read(b)
	return fmt.Sprintf("0x%x", b)
}

func populateMap(useStruct bool) {
	if useStruct {
		m := make(map[string]struct{}, addressCount)
		for i := 0; i < addressCount; i++ {
			m[randomAddress()] = struct{}{}
		}
		_ = m
	} else {
		m := make(map[string]bool, addressCount)
		for i := 0; i < addressCount; i++ {
			m[randomAddress()] = true
		}
		_ = m
	}
}

func main() {
	rand.Seed(time.Now().UnixNano())

	// === Toggle this to switch test type ===
	useStruct := false // or false

	// Delay to stabilize heap
	time.Sleep(1 * time.Second)

	// Start profiling after warm-up
	f, err := os.Create("heap_bool.prof")
	if err != nil {
		panic(err)
	}
	defer f.Close()

	// Run target
	populateMap(useStruct)

	// Write heap profile after map allocation
	if err := pprof.WriteHeapProfile(f); err != nil {
		panic(err)
	}

	fmt.Println("heap profile written")
}
```
</details>

Command used for comparison: 
```bash
go tool pprof --diff_base=heap_bool.prof heap_struct.prof
(pprof) top
Showing nodes accounting for -1.59MB, 2.75% of 57.90MB total
Showing top 10 nodes out of 13
      flat  flat%   sum%        cum   cum%
   -3.09MB  5.34%  5.34%    -1.09MB  1.89%  main.populateMap
...
```

x/erc20/keeper/allowance_test.go
zsystm: @cloudgray 
Just to clarify — I was referring to adding test cases that trigger failure paths (not tests that would be invalid by construction). But looks like that’s already covered in this commit: https://github.com/cosmos/evm/pull/90/commits/04e0de816cdfae2df069feb5638bc1d0dbf01ded

So no worries — this looks good to me!

x/erc20/genesis.go
zsystm: @cloudgray 
It looks like disabled token pairs never get their allowances stored during `InitGenesis`, because `SetAllowance` calls `checkTokenPair` and returns an error if the token is disabled. However, `InitGenesis` is supposed to restore all state from the genesis file — including allowances for disabled pairs.

Right now, we simply continue when `types.ErrERC20TokenPairDisabled` is encountered, so those allowances don’t actually get written to the KVStore. If the module logic requires that disabled pairs still have their allowances recorded at genesis (so they might be re-enabled later or at least remain consistent with the imported state), we’ll need to revise this flow. One potential fix is to skip the `checkTokenPair` check during genesis, or handle the “disabled” error as a non-fatal condition that still writes the allowance. Otherwise, we lose a piece of state that was originally present in the genesis file.

zsystm:
> @zsystm I added [deleteAllowances](https://github.com/cosmos/evm/blob/feat%2Ferc20-allowance/x/erc20/keeper/allowance.go#L151-L159) method and let it called by [DeleteTokenPair](https://github.com/cosmos/evm/blob/feat%2Ferc20-allowance/x/erc20/keeper/token_pairs.go#L102) method.
> 
> It is not likely to happen that TokenPair for erc20 precompile is deleted and it really does not happen in cosmos/evm. However, because DeleteTokenPair method is public method, appchain that imports cosmos/evm can call it in somewhere. Then, when that appchain import genesis, the situation where TokenPair does not exists but allowances exist can happen.
> 
> So, I thought that all allowances that share same erc20 address are deleted when TokenPair is deleted.
> 
> I applied this change in [80036a6](https://github.com/cosmos/evm/commit/80036a6f2ebf3a98613389a3f13930ba937493f9). Could you check this change?

@cloudgray 
I agree that allowances should be removed when a token pair is deleted — especially since the deletion happens when the paired [ERC20 contract is self-destructed](https://github.com/cosmos/evm/blob/main/x/erc20/keeper/msg_server.go#L42-L46). In that case, keeping the allowance data serves no real purpose, so cleaning it up makes sense.

zsystm:
@cloudgray 
I have one last comment from my side — please take a look when you get a chance: https://github.com/cosmos/evm/pull/90#discussion_r2050923989

x/erc20/genesis.go
cloudgray: applied in https://github.com/cosmos/evm/pull/90/commits/fcae5bb7f6fce5dbceff7ecdae4a3350a8cbe1bc and https://github.com/cosmos/evm/pull/90/commits/80036a6f2ebf3a98613389a3f13930ba937493f9

x/erc20/genesis_test.go
```diff
@@ -120,6 +166,13 @@ func (suite *GenesisTestSuite) TestERC20InitGenesis() {
 		} else {
 			suite.Require().Len(tc.genesisState.TokenPairs, 0, tc.name)
 		}
+
+		allowances := nw.App.Erc20Keeper.GetAllowances(nw.GetContext())
+		if len(allowances) > 0 {
+			suite.Require().Equal(tc.genesisState.Allowances, allowances, tc.name)
+		} else {
+			suite.Require().Len(tc.genesisState.Allowances, 0, tc.name)
+		}
```
zsystm, line 175: @cloudgray 
I think it would strengthen this test if we explicitly check that the actual return values match the expected ones, rather than only verifying length or presence.

In particular, we could extend the test table with an expectedAllowances field and compare it against the result of `GetAllowances()` using `suite.Require().Equal(...)`. This would help ensure that the correct allowances are properly initialized and nothing unexpected sneaks in.

Could you restructure the test cases slightly to make this more explicit and robust?

zsystm (commented):
Final review from my side. Let's quickly resolve this and merge it. 
Thanks!

x/erc20/genesis_test.go
zsystm: Oh it is already handled.

zsystm (approved):
LGTM. Let's squash & merge it.
For lint failures, we can handle this when open PR for merging feat into main.

Issue #61, opened 2025-04-09
Replace erc20/ Prefix to erc20: in Native ERC20 Coin Denoms


zsystm:
### Background

1. **IBC v2 does not allow `/` in the base denomination.**
    
    IBC version 2 introduces [a restriction](https://github.com/cosmos/ibc-go/blob/e5baeaea3e549f64055f0a60f2617a3404ea777d/modules/apps/transfer/v2/ibc_module.go#L71-L73) that disallows slashes (`/`) in base denominations. As a r esult, any denom containing `/` cannot be transferred using the IBC v2 router. 
    
2. **Native ERC20 coins** currently use the `erc20/` prefix, making them **incompatible with IBC v2**.
    - **What is a native ERC20 coin?**
        
        A native ERC20 coin is a Cosmos bank coin that represents an existing ERC20 token contract. It is created when a governance proposal registers a **token pair** between the ERC20 contract and a new bank denom (which starts with `erc20/`).
        
        This regsitration sets the contract owner to OWNER_EXTERNAL, since the ERC20 contract was already deployed by an EoA or CA(not the module itself).
        
    - **Why do we pair an existing ERC20 with a bank denom?**
        
        Because IBC works only with coins stored in the Cosmos bank module, not EVM state. So, in order to send ERC20 tokens (stored in EVM) from a Cosmos EVM-based chain to another chain via IBC, we must first convert them a bank-native coin.
        
    - **When is the `erc20/` prefix added?**
        - It’s added when a token pair is manually registered via governance through the [RegisterERC20](https://www.notion.so/ERC20-Prefix-Discussion-1cec1352439b80ec9efefb9ea2390697?pvs=21) flow, which internally triggers:
            1. registerERC20
            2. CreateCoinMetadata
            3. [CreateDenom](https://github.com/cosmos/evm/blob/08eb3b7afb339a6376f998128360278b705ab4fe/x/erc20/types/proposal.go#L42-L45)
                resulting in a denom like `erc20/0x…`
                
    - **What is an ERC20 token with `OWNER_MODULE`?**
        
        This represents the “[Single Token Representation V2](https://medium.com/evmos/evmos-introduces-native-tokens-as-erc-20-for-cosmos-1a4c7de5c3e9)” model. In this case, when a native coin from an external chain is received through IBC, the chain automatically:
        
        - Deploys an ERC20 contract
        - Adds it to the active precompile list
        
[...]

PR #92, opened 2025-04-23
fix!: replace erc20 prefix

zsystm:
# Description

Closes: #61 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [x] reviewed content
- [x] confirmed all CI checks have passed

### DIFF
--- a/x/erc20/types/proposal.go
+++ b/x/erc20/types/proposal.go
@@ -41,19 +41,22 @@ func CreateDenomDescription(address string) string {
 
 // CreateDenom generates a string the module name plus the address to avoid conflicts with names staring with a number
 func CreateDenom(address string) string {
-	return fmt.Sprintf("%s/%s", ModuleName, address)
+	return fmt.Sprintf("%s%s", ModuleName, address)
 }
 
-// ValidateErc20Denom checks if a denom is a valid erc20/
-// denomination
+// ValidateErc20Denom checks if a denom is a valid erc20 denomination.
+// Example: "erc200xabc..."
 func ValidateErc20Denom(denom string) error {
-	denomSplit := strings.SplitN(denom, "/", 2)
-
-	if len(denomSplit) != 2 || denomSplit[0] != ModuleName {
-		return fmt.Errorf("invalid denom. %s denomination should be prefixed with the format 'erc20/", denom)
+	if strings.HasPrefix(denom, ModuleName) {
+		trimmed := strings.TrimPrefix(denom, ModuleName)
+		if len(trimmed) > 0 && trimmed[0] == '/' {
+			// backward compatibility with old denom format ("erc20/0xabc...")
+			return cosmosevmtypes.ValidateAddress(trimmed[1:])
+		}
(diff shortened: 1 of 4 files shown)


x/erc20/types/proposal_test.go
```diff
@@ -80,12 +80,17 @@ func (suite *ProposalTestSuite) TestValidateErc20Denom() {
 		expPass bool
 	}{
 		{
-			"- instead of /",
+			"fail: invalid address",
+			"0xdac17f958d2ee523a22",
```
dongsam, line 84: How about adding a test case for `0xdac17f958d2ee523a2206206994597c13d831ec7` as well?
Just by looking at the current test cases, it might not be clear whether an input like `0xdac17f958d2ee523a22` is considered invalid because of an incorrect address length or because it lacks a proper prefix.

x/erc20/types/proposal.go
```diff
@@ -41,19 +42,32 @@ func CreateDenomDescription(address string) string {
...
 func CreateDenom(address string) string {
-	return fmt.Sprintf("%s/%s", ModuleName, address)
+	return Erc20NativeCoinDenomPrefix + address
 }
 
-// ValidateErc20Denom checks if a denom is a valid erc20/
-// denomination
+// ValidateErc20Denom checks if a denom is a valid erc20 denomination.
+// Valid formats: "erc20:0xabc..." or "erc20/0xabc..." (legacy format)
 func ValidateErc20Denom(denom string) error {
```
dongsam, line 50: FYI: `ValidateErc20Denom` is currently not used outside of the test code.
If it gets used somewhere in the future, the current validation logic which accepts the legacy format as valid might unintentionally differ from the intended standard.

dongsam (approved):
I’ve left a few minor comments, but overall the PR looks good and achieves its intended purpose.

x/erc20/types/proposal.go
zsystm: @dongsam 
Hmm… I agree. 

Even though `ValidateErc20Denom` is only used in test code for now, it's probably safer to disallow the legacy `/` format—just in case it ends up being used in actual logic later on. 🤔

x/erc20/types/proposal_test.go
zsystm: Applied it in https://github.com/cosmos/evm/pull/92/commits/2a6692799c309017841031c9f51a258f2159d224

x/erc20/types/proposal.go
zsystm: Disallowed legacy format in https://github.com/cosmos/evm/pull/92/commits/2a6692799c309017841031c9f51a258f2159d224

MalteHerrmann:
Why was this not applied to e.g. the `SanitizeERC20Name` function?
https://github.com/cosmos/evm/blob/74596307656f5f62dc390408e6e630dd0592eacc/x/erc20/types/utils.go#L34-L39?

Issue #32, opened 2025-03-28
Split Cosmos and EVM Chain IDs


vladjdk:
<details><summary></summary>

Currently, Evmos requires a chain ID to be formatted in the following way: `prefix_####-X`, with # signifying the Ethereum chain ID, and `prefix-X` symbolizing the Cosmos chain ID. This should be decoupled

</details>

PR #138, opened 2025-05-13
(chore)!: Separate Cosmos and EVM Chain IDs

vladjdk:
closes: #32 

Splits out the Cosmos chain ID from the EVM chain ID, allowing for chains to have regular Cosmos chains while maintaining EIP-155 compatibility.

The chain ID currently gets hardcoded in `cmd/config/config.go`. It then gets brought into and initialized in the customAppConfig's EVMConfig in `cmd/root.go`.

Since the object is a pointer, it gets passed around and is made accessible within the RPC, modules, and the rest of the app.

Although I would like to move the config into `app.toml`, since there are already other configurations (like denom and bech32 prefix), I think it's best done in a separate issue.

Additionally, I added an antehandler and a check in the JSON-RPC to ensure that the Chain's ID is equal to the signed transaction's chain ID. These exist as two separate objects, because the chain enforces replay protection on the consensus level, but if turned off, the RPC can separately enforce it on a node level.


ante/cosmos/eip712.go
Eric-Warehime, line 1: nit: I think it would have been nice to remove this in a separate PR since it's deprecated and its removal is not strictly related to the chain ID changes.

ante/evm/05_signature_verification.go
```diff
@@ -68,11 +68,19 @@ func SignatureVerification(
 	allowUnprotectedTxs bool,
 ) error {
 	ethTx := msg.AsTransaction()
+	ethCfg := evmtypes.GetEthChainConfig()
```
Eric-Warehime, line 71: The EIP-155 chain ID is also being stored in the keeper [here](https://github.com/cosmos/evm/blob/89894286132d0883ce6b258288bc99594106d50f/proto/cosmos/evm/vm/v1/evm.proto#L211) which we have access to.

Maybe we can't make a keeper call in every place we need it though? I think we should remove it and use the config everywhere.

vladjdk: Fair enough, I'll see what I can do for that. Fwiw, I copied the `AnteHandle` function above with the cfg getter but I could just pass it in from there instead.

vladjdk: Removed in https://github.com/cosmos/evm/pull/138/commits/dba4cbce8873c5997d76f11e28d4cd62219aa104

ante/cosmos/eip712.go
vladjdk: Readded here: https://github.com/cosmos/evm/pull/138/commits/3a9f625d8ae3b5822bb8b15a2fdce3c6ff091e15

server/config/config.go
```diff
@@ -60,6 +60,8 @@ const (
 	// DefaultMaxTxGasWanted is the default gas wanted for each eth tx returned in ante handler in check tx mode
 	DefaultMaxTxGasWanted = 0
 
+	DefaultEVMChainID = 262144
+
```
Eric-Warehime, line 64: Shouldn't we just panic if this is not set? I doubt a default is ever a good value here in production.

Eric-Warehime (approved):
I still think there's more cleanup to be done around having configs scattered across the codebase, but this is a nice start.

server/config/config.go
vladjdk: It's just a placeholder for the chain ID that initializes the root.go cmd. I'l rename it to NoOpChainID

Issue #130, opened 2025-05-11
Precompile(gov): Missing methods


zsystm:
## Problem

The governance precompile is currently missing several Cosmos SDK governance features:

- Transaction methods: `submit-proposal`, `deposit`, and `cancel-proposal` are not implemented.
- Query method: `constitution` (added in Cosmos SDK v0.50.1) is absent.

This limits the ability of EVM-based dApps to interact fully with on-chain governance and raises the question of whether full feature parity with the SDK module is required.

## Closing criteria

- [x] Discuss and decide whether the precompile should aim for full Cosmos SDK parity or support only a subset of governance features.
- [ ] Based on the decision:
  - [ ] Implement any missing tx methods (`submit-proposal`, `deposit`, `cancel-proposal`) if required.
  - [ ] Add the `constitution` query method if deemed necessary.
mechanism.

cc: @dudong2

zsystm:
@vladjdk @Eric-Warehime 

## Context
In Cosmos-SDK gov module, a [proposal’s messages field](https://github.com/cosmos/cosmos-sdk/blob/73aa8721552295ff7b5336014277c6639d84d24c/x/gov/types/v1/tx.pb.go#L45) is defined as repeated google.protobuf.Any.

## Options considered

### 1. Raw `Any` (`typeUrl` + bytes)

- **Pros**
  - Perfect 1-for-1 match with the SDK interface – every existing and future message type “just works”.
  - No JSON parsing inside the precompile.

- **Cons**
  - Solidity devs must encode protobuf wire format – high tooling / DX barrier.
  - Wrong typeUrl or bytes → opaque failure.

### 2. Typed helper methods for popular messages

e.g. `submitBankSend(…)`, `submitDelegate(…)`, …

- **Pros**
  - Very friendly UX; ABI enforces correct fields.
- Cons
  - Needs a wrapper per message – custom modules not covered automatically.
  - Constant maintenance as new SDK msgs appear.

### 3. JSON-blob proposal (implemented in the #142 )

`submitProposal(bytes jsonBlob, Coin[] deposit)` where `jsonBlob` is exactly the [proto-JSON document used by the Cosmos CLI](https://github.com/cosmos/cosmos-sdk/blob/af29f5521d0ad83153e108e9b57b69ecfb6bb696/x/gov/client/cli/tx.go#L98-L115) (messages array, metadata, title, …).

- **Pros**
  - No protobuf required – any language can build JSON.
  - Custom modules work automatically (interface-registry).
[...]

PR #142, opened 2025-05-14
feat(precompiles/gov): add support for submit-proposal, deposit, cancel-proposal txs and constitution query

zsystm:
# Description

This PR enhances the `x/gov` module integration by:

- Implementing the following transaction methods:
  - `SubmitProposal`
  - `Deposit`
  - `CancelProposal`
- Adding the `Constitution` query method (introduced in Cosmos SDK v0.50.1)

Closes: #130 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content

### DIFF
--- a/precompiles/gov/integration_test.go
+++ b/precompiles/gov/integration_test.go
@@ -1813,7 +1813,7 @@ var _ = Describe("Calling governance precompile from another contract", Ordered,
 			}
 			txArgs.Amount = common.Big0
 			eventCheck = passCheck.WithExpEvents(gov.EventTypeDeposit)
-			_, evmRes, err = s.factory.CallContractAndCheckLogs(depositorKey1, txArgs, callArgs, eventCheck)
+			_, _, err = s.factory.CallContractAndCheckLogs(depositorKey1, txArgs, callArgs, eventCheck)
 			Expect(err).To(BeNil())
 			Expect(s.network.NextBlock()).To(BeNil())
 
@@ -1823,6 +1823,8 @@ var _ = Describe("Calling governance precompile from another contract", Ordered,
 			rate := math.LegacyMustNewDecFromStr(params.ProposalCancelRatio)
 			proposalDeposits, err := s.network.App.GovKeeper.GetDeposits(s.network.GetContext(), proposalID)
 			Expect(err).To(BeNil())
+			Expect(proposalDeposits).To(HaveLen(2))
+
 			cancelFees = make(map[string]math.Int)
 			remainingFees = make(map[string]math.Int)
 
@@ -1837,6 +1839,8 @@ var _ = Describe("Calling governance precompile from another contract", Ordered,
 					}
 				}
 			}
+			Expect(cancelFees).To(HaveLen(2))
+			Expect(remainingFees).To(HaveLen(2))
 
 			callArgs.MethodName = "testCancelWithTransfer"
 		})
@@ -1952,7 +1956,7 @@ var _ = Describe("Calling governance precompile from another contract", Ordered,
 			}
 			txArgs.Amount = common.Big0
 			eventCheck = passCheck.WithExpEvents(gov.EventTypeDeposit)
-			_, evmRes, err = s.factory.CallContractAndCheckLogs(depositorKey1, txArgs, callArgs, eventCheck)
+			_, _, err = s.factory.CallContractAndCheckLogs(depositorKey1, txArgs, callArgs, eventCheck)
 			Expect(err).To(BeNil())
 			Expect(s.network.NextBlock()).To(BeNil())
 
@@ -1962,6 +1966,8 @@ var _ = Describe("Calling governance precompile from another contract", Ordered,
 			rate := math.LegacyMustNewDecFromStr(params.ProposalCancelRatio)
 			proposalDeposits, err := s.network.App.GovKeeper.GetDeposits(s.network.GetContext(), contractProposalID)
 			Expect(err).To(BeNil())
+			Expect(proposalDeposits).To(HaveLen(2))
+
 			cancelFees = make(map[string]math.Int)
 			remainingFees = make(map[string]math.Int)
 
@@ -1976,6 +1982,8 @@ var _ = Describe("Calling governance precompile from another contract", Ordered,
 					}
 				}
 			}
+			Expect(cancelFees).To(HaveLen(2))
+			Expect(remainingFees).To(HaveLen(2))
 
 			callArgs.MethodName = "testCancelFromContractWithTransfer"
 		})


precompiles/common/types.go
```diff
@@ -1,7 +1,9 @@
 package common
 
 import (
+	"fmt"
 	"math/big"
+	"reflect"
```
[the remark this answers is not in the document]
zsystm: > The ToCoins helper (reflection slow-path) does not introduce non-determinism in consensus execution.

- The function iterates over a slice with an index-based `for i := 0; i < len; i++` loop. Slice element order is fully deterministic and identical on every node.
- It accesses two named fields (Denom, Amount) whose positions are fixed in the anonymous
struct generated by go-ethereum/abi; no map iteration or field enumeration is involved.
- There is no use of random sources, time, goroutines, or unsafe pointer conversions.
- All error paths immediately return; there is no panic/recover branch that could differ
between runtimes.

Because every node receives the same calldata and runs the same bytecode,
this code path yields identical `[]Coin` values (or identical errors) on every replica.

The reflection is only used to bridge the unnamed tuple type emitted by the ABI decoder to our
concrete `cmn.Coin`; it does not depend on implementation-defined ordering.

_If the warning is purely heuristic, it can be safely dismissed for this piece of code._

precompiles/gov/testdata/GovCaller.json
```diff
@@ -123,8 +123,8 @@
...
...ff8111156108d9576108d8610589565b5b6108e584828501610668565b60808301525060a082015167ffffffffffffffff81111561090957610908610589565b5b61091584828501610668565b60a08301525060c082015167ffffffffffffffff81111561093957610938610589565b5b61094584828501610668565b60c08301525060e082015167ffffffffffffffff81111561096957610968610589565b5b61097584828501610668565b60e08301525061010082015167ffffffffffffffff81111561099a57610999610589565b5b6109a684828501610668565b610100830152506101206109bc848285016105a5565b6101208301525061014082015167ffffffffffffffff8111156109e2576109e1610589565b5b6109ee84828501610668565b6101408301525061016082015167ffffffffffffffff811115610a1457610a13610589565b5b610a20848285016107b5565b61016083015250610180610a36848285016107fa565b610180830152506101a0610a4c848285016107fa565b6101a0830152506101c0610a62848285016107fa565b6101c0830152506101e082015167ffffffffffffffff811115610a8857610a87610589565b5b610a9484828501610668565b6101e08301525092915050565b600060208284031215610ab757610ab66104ff565b5b600082015167ffffffffffffffff811115610ad557610ad4610504565b5b610ae18482850161080f565b9150509291505056fea2646970667358221220ef70f6cc09cfd814baf88d1053e659c26a578f375ef1dbf3319576c02d3a28e164736f6c63430008140033",
```
vladjdk, line 126: I can't seem to find the .sol file for the bytecode

vladjdk (changes requested):
Overall, I don't mind the direction with JSON instead of protobuf. I'd need to see the solidity code for the function to see how it's called. Will finish the review after that's included in the diff

precompiles/gov/errors.go
```diff
@@ -21,6 +21,12 @@ const (
...
 	// ErrInvalidWeightedVoteOptionWeight invalid weighted vote option weight.
 	ErrInvalidWeightedVoteOptionWeight = "invalid weighted vote option weight %s "
-	// ErrInvalidDepositor invalid depositor.
-	ErrInvalidDepositor = "invalid depositor %s "
+	// ErrInvalidProposalJSON invalid proposal json.
+	ErrInvalidProposalJSON = "invalid proposal json %s "
+	// ErrInvalidProposer invalid proposer.
+	ErrInvalidProposer = "invalid proposer %s"
+	// ErrInvalidDepositor invalid firstAddr address.
+	ErrInvalidDepositor = "invalid firstAddr address: %s"
```
vladjdk: Nit: I think it's more clear to keep the wording as invalid depositor

precompiles/gov/integration_test.go
```diff
@@ -747,5 +935,54 @@ var _ = Describe("Calling governance precompile from EOA", func() {
...
+// -----------------------------------------------------------------------------
+// Helper functions (test‑only)
+// -----------------------------------------------------------------------------
+
+func minimalDeposit(denom string) []cmn.Coin {
+	return []cmn.Coin{{Denom: denom, Amount: big.NewInt(1)}}
+}
+
+// minimalBanSendProposalJSON returns a valid governance proposal encoded as UTF‑8 bytes.
+func minimalBanSendProposalJSON(to sdk.AccAddress, denom, amount string) []byte {
```
vladjdk: Nit: Ban -> Bank

precompiles/gov/integration_test.go
```diff
@@ -77,11 +91,185 @@ var _ = Describe("Calling governance precompile from EOA", func() {
 			To: &precompileAddr,
 		}
 		txArgs.GasLimit = 200_000
+
+		firstKey = s.keyring.GetPrivKey(0)
```
vladjdk: Nit: can we rename this to be depositor/proposer/etc? Kind of confusing when just reading the tests

precompiles/gov/integration_test.go
```diff
@@ -77,11 +91,185 @@ var _ = Describe("Calling governance precompile from EOA", func() {
...
+				gov.ErrDifferentOrigin,
+				notProposerAddr.String(),
+				firstAddr.String(),
+			)
+
+			_, _, err := s.factory.CallContractAndCheckLogs(notProposerKey, txArgs, callArgs, errCheck)
+			Expect(err).To(BeNil())
+		})
+
+		It("cancels a live proposal and emits event", func() {
```
vladjdk, line 246: Let's test out that canceling returns the deposit less the cancellation fee in a separate test

precompiles/gov/tx.go
```diff
@@ -12,12 +12,111 @@ import (
...
+	}
+
+	if _, err = govkeeper.NewMsgServerImpl(&p.govKeeper).Deposit(ctx, msg); err != nil {
+		return nil, err
+	}
+
+	if err = p.EmitDepositEvent(ctx, stateDB, depositorHexAddr, msg.ProposalId, msg.Amount); err != nil {
+		return nil, err
+	}
+
```
vladjdk, line 110: Do we need to add something like this?

https://github.com/cosmos/evm/blob/89894286132d0883ce6b258288bc99594106d50f/precompiles/staking/tx.go#L183-L191

precompiles/gov/tx.go
```diff
@@ -12,12 +12,111 @@ import (
...
+	if _, err = govkeeper.NewMsgServerImpl(&p.govKeeper).CancelProposal(ctx, msg); err != nil {
+		return nil, err
+	}
+
+	if err = p.EmitCancelProposalEvent(ctx, stateDB, proposerHexAddr, msg.ProposalId); err != nil {
+		return nil, err
+	}
+
```
vladjdk, line 172: Same as above, but for cancellation deposit refunds

vladjdk (changes requested):
Left a couple small comments and questions

vladjdk:
Overall, I think the JSON direction is fine here for the sake of easing the development experience. I'll discuss this with the team for consensus, but in the meantime, you can proceed with this!

precompiles/gov/errors.go
zsystm: Good catch! Applied in https://github.com/cosmos/evm/pull/142/commits/24bcf5dff5fa45d070ab18be33e2cb68b63685f3

precompiles/gov/integration_test.go
zsystm: Applied in https://github.com/cosmos/evm/pull/142/commits/24bcf5dff5fa45d070ab18be33e2cb68b63685f3

precompiles/gov/integration_test.go
zsystm: Applied in https://github.com/cosmos/evm/pull/142/commits/24bcf5dff5fa45d070ab18be33e2cb68b63685f3

precompiles/gov/integration_test.go
zsystm: Added in https://github.com/cosmos/evm/pull/142/commits/13fe300077bbb194ebaf33a12f848720c0e5d710

vladjdk: Let's also add one more test that checks for multiple depositors on a single proposal. After a cancellation, each of them should get their deposits back.

vladjdk (changes requested):
I'd like to see one last test, where we have multiple depositors on a single proposal. Each of those depositors should get their deposit returned after a proposal has been canceled.

zsystm:
> I'd like to see one last test, where we have multiple depositors on a single proposal. Each of those depositors should get their deposit returned after a proposal has been canceled.

Added in https://github.com/cosmos/evm/compare/99b1431...7d8be9f3e19b30b87f599374ca6afe2ef778744e

vladjdk (approved):

Issue #126, opened 2025-05-11
Precompile(distribution): FundCommunityPool Denom Handling (Needs Discussion)


zsystm:
## Problem

The [FundCommunityPool](https://github.com/cosmos/evm/blob/main/precompiles/distribution/tx.go#L243-L251) method is currently hard-coded for a single denom, which does not match the SDK’s flexible Coins approach.  
It needs to be discussed whether we should support multiple denoms or keep the implementation simplified for `evmDenom` only.

## Closing criteria

- [X] Reach a consensus on whether `FundCommunityPool` should accept multiple denoms.
- [ ] Update the implementation accordingly (support multi-denom or document the single-denom restriction).

cc: @cloudgray

PR #145, opened 2025-05-15
feat(precompiles/distribution): update FundCommunityPool interface

zsystm:
# Description

Closes: #126 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


precompiles/common/types.go
```diff
@@ -1,7 +1,9 @@
 package common
 
 import (
+	"fmt"
 	"math/big"
+	"reflect"
```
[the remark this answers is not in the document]
zsystm: > The ToCoins helper (reflection slow-path) does not introduce non-determinism in consensus execution.

- The function iterates over a slice with an index-based `for i := 0; i < len; i++` loop. Slice element order is fully deterministic and identical on every node.
- It accesses two named fields (Denom, Amount) whose positions are fixed in the anonymous
struct generated by go-ethereum/abi; no map iteration or field enumeration is involved.
- There is no use of random sources, time, goroutines, or unsafe pointer conversions.
- All error paths immediately return; there is no panic/recover branch that could differ
between runtimes.

Because every node receives the same calldata and runs the same bytecode,
this code path yields identical `[]Coin` values (or identical errors) on every replica.

The reflection is only used to bridge the unnamed tuple type emitted by the ABI decoder to our
concrete `cmn.Coin`; it does not depend on implementation-defined ordering.

_If the warning is purely heuristic, it can be safely dismissed for this piece of code._

precompiles/distribution/events.go
```diff
@@ -164,31 +165,39 @@ func (p Precompile) EmitWithdrawValidatorCommissionEvent(ctx sdk.Context, stateD
...
-	topics := make([]common.Hash, 2)
 
-	// The first topic is always the signature of the event.
-	topics[0] = event.ID
-
-	var err error
-	topics[1], err = cmn.MakeTopic(depositor)
-	if err != nil {
-		return err
+	for _, coin := range coins {
```
vladjdk, line 175: Could we add a test for the changes in the events emitted/logs?

vladjdk (changes requested):
Overall lgtm, let's just add a test for the emitted events

precompiles/distribution/setup_test.go
```diff
@@ -75,6 +75,7 @@ func (s *PrecompileTestSuite) SetupTest() {
 	}
 
 	nw := network.NewUnitTestNetwork(
+		network.WithOtherDenoms([]string{"foo", "bar"}),
```
cloudgray: How about making "foo", "bar" as constant in testutil/constant pacakge?

In #146, I declared those variables in the testutil/constant package and assigned them to member variables in PrecompileTestSuite, setting them using WithOtherDenoms. [ref](https://github.com/cosmos/evm/pull/146/files#diff-1d0af9d98a35a54d486b152f5d984229a624ff603be298800babc769d4adc9d2R34-R85)

I’m bringing this up in advance as it’s likely to cause a conflict.

precompiles/distribution/events.go
zsystm: Applied in https://github.com/cosmos/evm/pull/145/commits/8b562a3e17a6cbb265f577bbdfd728a07f9e18a3

precompiles/distribution/setup_test.go
zsystm: Applied in https://github.com/cosmos/evm/pull/145/commits/8b562a3e17a6cbb265f577bbdfd728a07f9e18a3

vladjdk:
We need to fix some conflicts here

zsystm:
> We need to fix some conflicts here

@vladjdk Fixed conflicts.

vladjdk (approved):

PR #150, opened 2025-05-16
EVM Callbacks

AdityaSripal:
# Description

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


x/vm/statedb/statedb.go
```diff
@@ -111,7 +111,7 @@ func (s *StateDB) MultiStoreSnapshot() storetypes.CacheMultiStore {
 	// the cacheCtx multi store is already a CacheMultiStore
 	// so we need to pass a copy of the current state of it
 	cms := s.cacheCtx.MultiStore().(storetypes.CacheMultiStore)
-	snapshot := cms.Copy()
+	snapshot := cms.CacheMultiStore()
```
AdityaSripal: Had to change this to fix the SDK upgrade. Not sure the correct way to fix this

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if !isCbPacket {
+		return nil
+	}
+
+	ethReceiver := common.BytesToAddress(receiver)
+	contractAddr := common.HexToAddress(contractAddress)
+
+	// TODO: Do something with the response
+	_, err = k.evmKeeper.CallEVMWithData(cachedCtx, ethReceiver, &contractAddr, cbData.Calldata, true)
```
AdityaSripal: `msg.sender` here should be the module address

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,157 @@
...
+	version string,
+) error {
+	data, err := transfertypes.UnmarshalPacketData(packet.GetData(), version, "")
+	if err != nil {
+		return err
+	}
+
+	// can only call callback if the receiver is the module address
+	receiver := sdk.MustAccAddressFromBech32(data.Receiver)
+	if !receiver.Equals(k.authKeeper.GetModuleAddress(types.ModuleName)) {
```
AdityaSripal: Only continue callback logic on `recvPacket` if the receiver of the tokens is the module account address. This is because we set the msg.sender on the contract callback to the module address

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if !isCbPacket {
+		return nil
+	}
+
+	sender := common.BytesToAddress(sdk.MustAccAddressFromBech32(packetSenderAddress))
+	contractAddr := common.HexToAddress(contractAddress)
+
+	// TODO: Do something with the response
+	_, err = k.evmKeeper.CallEVMWithData(cachedCtx, sender, &contractAddr, cbData.Calldata, true)
```
AdityaSripal: Here the contract is getting the calldata in the packet regardless of what the acknowledgment is. Question: Should we allow packet senders to provide different calldata for successful and unsuccessful acknowledgements

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// TODO: Do something with the response
+	_, err = k.evmKeeper.CallEVMWithData(cachedCtx, ethReceiver, &contractAddr, cbData.Calldata, true)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (k ContractKeeper) IBCOnAcknowledgementPacketCallback(
```
AdityaSripal, line 271: Note: Here we can use the `packetsenderAddress` safely as `msg.sender` in `AcknowledgePacket` and `TimeoutPacket`

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	if !isCbPacket {
+		return nil
+	}
+
+	sender := common.BytesToAddress(sdk.MustAccAddressFromBech32(packetSenderAddress))
+	contractAddr := common.HexToAddress(contractAddress)
+
+	// TODO: Do something with the response
+	_, err = k.evmKeeper.CallEVMWithData(cachedCtx, sender, &contractAddr, cbData.Calldata, true)
```
AdityaSripal: If we support different success and failure calldata for acknowledgement, we can use the failure calldata here as well.

x/ibc/callbacks/keeper/keeper.go
Eric-Warehime: To clarify, we can use `msg.sender` here because we assume there will be no logic around who is sending the ack or timeout?

x/ibc/callbacks/keeper/keeper.go
Eric-Warehime: It makes sense to require the packet to be sent to the module address as a way to avoid footguns. Are we concerned with the security around executing all callbacks from a single account? For example, any approvals which were not cleaned up could be exploited by anyone invoking a callback. Maybe there's no way to exploit that, but it feels risky. 
[CW Hooks](https://github.com/cosmos/ibc-apps/tree/main/modules/ibc-hooks#cosmwasm-contract-execution-format) has some hashing logic here.
> Sender: IBC packet senders cannot be explicitly trusted, as they can be deceitful. Chains cannot risk the sender being confused with a particular local user or module address. To prevent this, the sender is replaced with an account that represents the sender prefixed by the channel and a Wasm module prefix. This is done by setting the sender to Bech32(Hash("ibc-wasm-hook-intermediary" || channelID || sender)), where the channelId is the channel id on the local chain.

I'm wondering if the same or similar. should be used here.

x/vm/statedb/statedb.go
Eric-Warehime: This should fixed [on main](https://github.com/cosmos/evm/blob/3977168b066e13dda869721bf0b7bab8b057f6ad/x/vm/statedb/statedb.go#L108) already

ante/interfaces/cosmos.go
```diff
@@ -18,6 +19,9 @@ type AccountKeeper interface {
 	GetParams(ctx context.Context) (params authtypes.Params)
 	GetSequence(ctx context.Context, addr sdk.AccAddress) (uint64, error)
 	AddressCodec() addresscodec.Codec
+	UnorderedTransactionsEnabled() bool
+	RemoveExpiredUnorderedNonces(ctx sdk.Context) error
+	TryAddUnorderedNonce(ctx sdk.Context, sender []byte, timestamp time.Time) error
```
Eric-Warehime: Why do we now depend on unordered transactions? I don't see it being used in this PR.

x/ibc/callbacks/keeper/keeper.go
Eric-Warehime: I assume people will need different behaviors to either finalize or unwind application logic here, but we should get some more product feedback around exactly what is required since I don't have a clear picture of what that would look like.

dhfang: What currently sets the calldata used here? Is it the contract that handles the receive packet?

AdityaSripal: This is all being set by the original packet sender at the start of the packet lifecycle

x/ibc/callbacks/keeper/keeper.go
AdityaSripal: Yes I thought about this. We could follow the convention set by CW hooks as well. I didn't do this at first because this was moving in the direction of the gmp work that @srdtrk is doing anyway. But perhaps it is important to have individual accounts even in this case.

x/ibc/callbacks/keeper/keeper.go
AdityaSripal: Not sure I understand this question

Eric-Warehime: I don't think there's a question here. It seems like `msg.sender` is not a "trustable" field. So we can't actually use this field for anything, which is why it seems prudent to have that field be the hashed value of something.

x/ibc/callbacks/keeper/keeper.go
dhfang: Hmm maybe I'm missing something, but if the source calldata is chosen prior to packet creation, how does a contract determine for which packet this source callback is associated with? For example, our entrypoint contract has a usecase where when handling a timed out or failed transfer, it will send the recovered assets to an address that is stored in state with an association to the packet seq no / channel of the transfer. How would our entrypoint contract determine what packet is failing in the evm callbacks case?

contracts/solidity/evmd/eips/testdata/Counter.sol
AdityaSripal, line 1: All of these files get deleted if you call:

`make contracts-all`

If you want to add them back, you can just call:

`make contracts-compile`

evmd/app.go
```diff
@@ -545,19 +550,27 @@ func NewExampleApp(
 
 		transfer stack contains (from bottom to top):
 			- ERC-20 Middleware
+			- IBC Callbacks Middleware (with EVM ContractKeeper)
 			- IBC Transfer
 
 		SendPacket, since it is originating from the application to core IBC:
-		 	transferKeeper.SendPacket -> erc20.SendPacket -> channel.SendPacket
+		 	transferKeeper.SendPacket -> callbacks.SendPacket -> erc20.SendPacket -> channel.SendPacket
```
AdityaSripal: The SendPacket flow doesn't happen this way because the ICS4 stack is not wired correctly.

ERC20Keeper must implement ICS4Wrapper

and then we can wire this correctly.

It doesn't have an effect at the moment, but we can make this fix later

evmd/app.go
```diff
@@ -545,19 +550,27 @@ func NewExampleApp(
...
 		RecvPacket, message that originates from core IBC and goes down to app, the flow is the other way
-			channel.RecvPacket -> erc20.OnRecvPacket -> transfer.OnRecvPacket
+			channel.RecvPacket -> erc20.OnRecvPacket -> callbacks.OnRecvPacket -> transfer.OnRecvPacket
 	*/
 
 	// create IBC module from top to bottom of stack
 	var transferStack porttypes.IBCModule
 
 	transferStack = transfer.NewIBCModule(app.TransferKeeper)
+	maxCallbackGas := uint64(1_000_000)
```
AdityaSripal, line 566: Not used in EVM contractKeeper

precompiles/callbacks/ICallbacks.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: LGPL-3.0-only
+pragma solidity >=0.8.18;
+
+interface ICallbacks {
```
AdityaSripal, line 4: Here is the ABI interface for source side callbacks. Looking for feedback here

precompiles/callbacks/callbacks.go
AdityaSripal, line 1: This file ended up being unnecessary. It was generated by `abigen`. We can remove if desired

x/ibc/callbacks/README.md
```diff
@@ -0,0 +1,180 @@
...
+    function onPacketTimeout(
+        string memory channelId,
+        string memory portId,
+        uint64 sequence,
+        bytes memory data
+    ) external;
+}
+```
+
+## Limitations
```
AdityaSripal, line 200: NOTE THE LIMITATIONS

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	ethReceiver := common.BytesToAddress(receiver)
+	contractAddr := common.HexToAddress(contractAddress)
+
+	// TODO: Approve the ERC20 tokens in the transfer packet for the contract on behalf of the receiver
+	// before calling the callback.
+
+	// TODO: Do something with the response
+	_, err = k.evmKeeper.CallEVMWithData(cachedCtx, ethReceiver, &contractAddr, cbData.Calldata, true)
```
AdityaSripal: I'm not sure how to pass in the gas limit for the EVM call with the provided function

x/ibc/callbacks/keeper/keeper.go
AdityaSripal: Here we do not want the packet sender to spoof a msg.sender on the receiving side

But on the source side, we can allow the packet sender to approve contract executions on behalf of himself. So there is no need for a hashed value here imo

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return err
+	}
+	if !isCbPacket {
+		return nil
+	}
+
+	// can only call callback if the receiver is the isolated address for the packet sender on this chain
+	receiver := sdk.MustAccAddressFromBech32(data.Receiver)
+	isolatedAddr := types.GenerateIsolatedAddress(packet.GetDestChannel(), data.Sender)
```
AdityaSripal: This should be converted into an ETH address but not sure exactly how this should work.

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if !receiver.Equals(isolatedAddr) {
+		return errorsmod.Wrapf(types.ErrInvalidReceiverAddress, "expected %s, got %s", isolatedAddr.String(), receiver.String())
+	}
+
+	ethReceiver := common.BytesToAddress(receiver)
+	contractAddr := common.HexToAddress(contractAddress)
+
+	// TODO: Approve the ERC20 tokens in the transfer packet for the contract on behalf of the receiver
```
AdityaSripal: Wasn't sure how this should happen in cosmos/evm so I left as a TODO now

x/ibc/callbacks/keeper/keeper_test.go
```diff
@@ -0,0 +1,277 @@
...
+	)
+	testCases := []struct {
+		name     string
+		malleate func()
+		expErr   error
+	}{
+		{
+			"success",
+			func() {},
+			nil, // TODO: This is unexpected. Not sure why this is passing without a valid contract address
```
AdityaSripal: I don't understand why the EVM didn't throw an error here

x/ibc/callbacks/keeper/keeper_test.go
```diff
@@ -0,0 +1,277 @@
...
```
AdityaSripal: I don't understand why this doesn't throw an error in EVM

AdityaSripal (commented):
Self-review to create TODOs and further explanations

x/ibc/callbacks/keeper/keeper.go
AdityaSripal: TODO: Modify this function to take in a custom gas cap which can be taken from the callback data

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	ethReceiver := common.BytesToAddress(receiver.Bytes())
+	contractAddr := common.HexToAddress(contractAddress)
+
+	// TODO: Approve the ERC20 tokens in the transfer packet for the contract on behalf of the receiver
+	// before calling the callback.
+
+	// TODO: Do something with the response
+	_, err = k.evmKeeper.CallEVMWithData(cachedCtx, ethReceiver, &contractAddr, cbData.Calldata, true)
```
AdityaSripal: TODO: Check if EVM reversion matches SDK transaction expectations:

1. If the EVM tx reverts, does the EVM state also revert.
2. If the SDK transaction fails at a later point after a successful EVM call, does the EVM state revert along with the Cosmos state

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,182 @@
...
+
+	"github.com/cosmos/evm/x/ibc/callbacks/types"
+)
+
+// ContractKeeper implements callbacktypes.ContractKeeper
+var _ callbacktypes.ContractKeeper = (*ContractKeeper)(nil)
+
+type ContractKeeper struct {
+	authKeeper            types.AccountKeeper
+	evmKeeper             types.EVMKeeper
```
AdityaSripal, line 34: TODO: We need to fail if the contract address has no code before calling into EVM since this will fail silently in the EVM side

x/ibc/callbacks/keeper/keeper_test.go
vladjdk: If you send a transaction with packed data to an address that has no code, the transaction will still succeed, but the data will be ignored.

Added a check that fails if there's no code at the contract account.

x/ibc/callbacks/keeper/keeper_test.go
vladjdk: If you send a transaction with packed data to an address that has no code, the transaction will still succeed, but the data will be ignored.

Added a check that fails if there's no code at the contract account.

x/ibc/callbacks/keeper/keeper.go
vladjdk: Added a check that fails if there's no code at the contract account.

x/ibc/callbacks/keeper/keeper.go
vladjdk: Added in https://github.com/cosmos/evm/pull/150/commits/8345eea8f922f7e03853dc8b4c633075a50758e4. Ended up using the remaining context gas as the gas cap. Need confirmation that this was the right move or if we should use the max callback gas value itself.

x/ibc/callbacks/keeper/keeper.go
vladjdk: Added: https://github.com/cosmos/evm/pull/150/commits/5b015912ed0827ff920deab879757d44ed878221

x/ibc/callbacks/keeper/keeper.go
vladjdk: Added: https://github.com/cosmos/evm/pull/150/commits/5b015912ed0827ff920deab879757d44ed878221

x/erc20/keeper/ibc_callbacks.go
```diff
@@ -111,7 +111,7 @@ func (k Keeper) OnRecvPacket(
 	// IsNativeFromSourceChain will check if the coin is native from the source chain.
 	// If the coin denom starts with `factory/` then it is a token factory coin, and we should not convert it
 	// NOTE: Check https://docs.osmosis.zone/osmosis-core/modules/tokenfactory/ for more information
-	case !found && strings.HasPrefix(coin.Denom, "ibc/") && ibc.IsBaseDenomFromSourceChain(data.Denom):
+	case !found && strings.HasPrefix(coin.Denom, "ibc/"):
```
zsystm, line 114: @vladjdk @AdityaSripal 
Could you clarify why this condition was removed?
If we're removing it intentionally, we should also update the comments in Case 1 to reflect the change.

x/ibc/callbacks/README.md
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+```
+
+An ICS-20 packet is formatted correctly for EVM callbacks on destination chain if all of the following are true:
+
+- The `memo` is not blank.
+- The`memo` is valid JSON.
+- The `memo` has at least one key, with the value `"dest_callbacks"`.
+- The `memo["dest_callbacks"]` has these two entries, `"contract"` and `"calldata"`.
```
zsystm:
```suggestion
- The `memo["dest_callback"]` has these three entries, `"address"`, `"gas_limit"` and `"calldata"`.
```

x/ibc/callbacks/README.md
```diff
@@ -0,0 +1,216 @@
...
+	commit bool,
+    gasCap *big.Int,
+) (*types.MsgEthereumTxResponse, error)
+```
+
+For use with EVM `recvPacket callbacks, the message fields above can be derived from the following:
+
+- `Sender`: IBC packet senders cannot be explicitly trusted, as they can be deceitful. Chains cannot
+risk the sender being confused with a particular local user or module address. To prevent this, the
+`sender` is replaced with an account that represents the sender prefixed by the channel and a Wasm module
```
Eric-Warehime:
```suggestion
`sender` is replaced with an account that represents the sender prefixed by the channel and a VM module
```

x/ibc/callbacks/README.md
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+}
+```
+
+An ICS-20 packet is formatted correctly for EVM callbacks on destination chain if all of the following are true:
+
+- The `memo` is not blank.
+- The`memo` is valid JSON.
+- The `memo` has at least one key, with the value `"dest_callbacks"`.
```
Eric-Warehime:
```suggestion
- The `memo` has at least one key, with the value `"dest_callback"`.
```

precompiles/erc20/erc20.go
```diff
@@ -22,16 +22,24 @@ const (
...
+	GasDecreaseAllowance = 3_620
 	GasName              = 3_421
 	GasSymbol            = 3_464
 	GasDecimals          = 427
-	GasTotalSupply       = 2_477
-	GasBalanceOf         = 2_851
-	GasAllowance         = 3_246
+	GasTotalSupply       = 2_480
+	GasBalanceOf         = 2_870
+	GasAllowance         = 3_225
```
Eric-Warehime, line 42: Why are we using more gas for these now?

precompiles/erc20/erc20.go
```diff
@@ -22,16 +22,24 @@ const (
...
+	// necessary tests for this use case.
+	//
+	// The results can be inspected here:
+	// https://github.com/evmos/evmos/blob/malte/erc20-gas-tests/precompiles/erc20/plot_gas_values.ipynb
+
+	GasTransfer          = 9_000
+	GasTransferFrom      = 30_500
+	GasApprove           = 8_100
+	GasIncreaseAllowance = 8_580
+	GasDecreaseAllowance = 3_620
```
Eric-Warehime, line 36: Why are these values changing so much?

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,429 @@
...
+	// `ProcessCallback` in IBC-Go overrides the infinite gas meter with a basic gas meter,
+	// so we need to generate a new infinite gas meter to run the EVM executions on.
+	// Skipping this causes the EVM gas estimation function to deplete all Cosmos gas.
+	// We re-add the actual EVM call gas used to the original context after the call is complete
+	// with the gas retrieved from the EVM message result.
+	cachedCtx, writeFn := ctx.CacheContext()
+	cachedCtx = evmante.BuildEvmExecutionCtx(cachedCtx).
+		WithGasMeter(types2.NewInfiniteGasMeterWithLimit(cbData.CommitGasLimit))
+
+	receiver := sdk.MustAccAddressFromBech32(data.Receiver)
```
Eric-Warehime: Why is it ok to panic here?

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,429 @@
...
+	// with the gas retrieved from the EVM message result.
+	cachedCtx, writeFn := ctx.CacheContext()
+	cachedCtx = evmante.BuildEvmExecutionCtx(cachedCtx).
+		WithGasMeter(types2.NewInfiniteGasMeterWithLimit(cbData.CommitGasLimit))
+
+	if len(cbData.Calldata) != 0 {
+		return errorsmod.Wrap(types.ErrInvalidCalldata, "timeout callback data should not contain calldata")
+	}
+
+	sender := common.BytesToAddress(sdk.MustAccAddressFromBech32(packetSenderAddress))
```
Eric-Warehime: Also panics here?

x/ibc/callbacks/keeper/keeper.go
Eric-Warehime: Maybe because `data.Receiver` has already been validated somewhere before this in the stack?

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,429 @@
...
+
+	var allowance *big.Int
+	err = erc20.ABI.UnpackIntoInterface(&allowance, "allowance", res.Ret)
+	if err != nil {
+		return errorsmod.Wrapf(types.ErrAllowanceFailed, "failed to unpack allowance: %v", err)
+	}
+
+	if allowance.Cmp(big.NewInt(1)) != 0 {
+		return errorsmod.Wrapf(types.ErrAllowanceFailed, "failed to set allowance")
+	}
```
Eric-Warehime: I think this is really confusing to read. It's not clear to me that the `allowance` value being returned here is actually a boolean for success and not the actual allowance value being set by the `approve` call.

Are we not able to actually unpack to a bool here instead?

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,429 @@
...
+
+	// Check that the contract has received all the tokens.
+	// NOTE: contracts must implement an IERC20(token).transferFrom(msg.sender, address(this), amount)
+	// for the total amount, or the callback will fail.
+	// This check is here to prevent funds from getting stuck in the isolated address,
+	// since they would become irretrievable.
+	contractTokenBalance := k.erc20Keeper.BalanceOf(ctx, erc20.ABI, tokenPair.GetERC20Contract(), contractAddr) // here, we can use the original ctx and skip manually adding the gas
+	if contractTokenBalance.Cmp(amountInt.BigInt()) != 0 {
+		return errorsmod.Wrapf(erc20types.ErrEVMCall, "contract balance %d does not equal sent amount %d", contractTokenBalance, amountInt.BigInt())
+	}
```
Eric-Warehime: How does this work if the callback contract already had tokens of this denom before the callback, or if the callback has already forwarded the tokens out to another account? I think you just need to check that the sender account has no tokens in it rather than assuming anything about the state of the callback contract.

x/ibc/callbacks/keeper/keeper.go
```diff
@@ -0,0 +1,429 @@
...
+	// This check is required because if there is no code, the call will still pass on the EVM side,
+	// but it will ignore the calldata and funds may get stuck.
+	if !contractAccount.IsContract() {
+		return errorsmod.Wrapf(types.ErrCallbackFailed, "provided contract address is not a contract: %s", contractAddr)
+	}
+
+	abi, err := callbacksabi.LoadABI()
+	if err != nil {
+		return err
+	}
```
Eric-Warehime, line 324: This is mostly the same across multiple callbacks right? Can we just refactor this into a function and call it in each one?

Eric-Warehime: Nevermind, I tried it and it's an ugly function with too many return variables.

precompiles/erc20/erc20.go
vladjdk: Based on this: https://github.com/evmos/evmos/pull/2142

Malte was saying that he missed this in Evmos/OS when he was cherry picking commits

x/ibc/callbacks/keeper/keeper.go
vladjdk: Not ok - we should error out instead

x/ibc/callbacks/keeper/keeper.go
vladjdk: Same as above

x/ibc/callbacks/keeper/keeper.go
vladjdk: Let's unpack a bool and rename it to success. The naming is an artifact of me initially believing that you got an allowance amount back, and then changing the check after I realized it was a success/fail

x/ibc/callbacks/keeper/keeper.go
vladjdk: Good call - current code would only allow callbacks on a contract one time and would break afterwards

x/erc20/keeper/ibc_callbacks.go
vladjdk: Let's update the docs here. I don't think it makes sense to only accept single hop tokens, given the fact that Eureka routes through the Hub and recipient chains may be using Eureka.

Eric-Warehime (approved):

PR #179, opened 2025-05-23
Fix compile issue in start.go

NotJeremyLiu:
## Summary
- remove undefined error check in `start.go`

## Testing
- `go fmt server/start.go`
- `go build ./server...` *(fails: proxyconnect tcp: no route to host)*

### DIFF
--- a/server/start.go
+++ b/server/start.go
@@ -145,10 +145,6 @@ which accepts a path for the resulting pprof file.
 			}
 
 			serverCtx.Logger.Debug("received quit signal")
-			// TODO: why is this check here? Should not make sense since err is checked above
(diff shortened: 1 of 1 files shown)


server/start.go
```diff
@@ -145,10 +145,6 @@ which accepts a path for the resulting pprof file.
 			}
 
 			serverCtx.Logger.Debug("received quit signal")
-			// TODO: why is this check here? Should not make sense since err is checked above
```
davidterpay, line 147: why is this check here? doesnt make sense since the error is checked above. can likely remove it

aljo242:
@NotJeremyLiu plz fix

PR #116, opened 2025-04-29
Bump geth to v1.13

vladjdk:
requires: https://github.com/cosmos/go-ethereum/pull/2
closes: #115

PR #181, opened 2025-05-23
(feat)!: upgrade to geth 1.13

vladjdk:
forks #116


precompiles/distribution/integration_test.go
```diff
@@ -2614,21 +2614,20 @@ var _ = Describe("Calling distribution precompile from another contract", Ordere
...
 			Expect(err).To(BeNil(), "error while calling the smart contract: %v", err)
-			Expect(txRes).NotTo(BeNil(), "expected tx to succeed")
 			Expect(s.network.NextBlock()).To(BeNil())
 
 			balRes, err := s.grpcHandler.GetBalanceFromBank(s.keyring.GetAccAddr(0), s.bondDenom)
 			Expect(err).To(BeNil())
 			finalBalance := balRes.Balance
-			fees := math.NewIntFromUint64(txRes.GasUsed).Mul(math.NewIntFromBigInt(txArgs.GasPrice))
-			Expect(finalBalance.Amount.Equal(initialBalance.Amount.Sub(fees))).To(BeTrue(), "expected final balance must be initial balance minus any gas spent")
+			expectedGasCost := math.NewInt(79_416_000_000_000)
```
vladjdk, line 2629: Gas increased from 79_415_000_000_000 to 79_416_000_000_000. My theory is it's a small opcode gas diff with the new geth, but not 100% sure

x/vm/statedb/state_object.go
```diff
@@ -100,27 +108,29 @@ func (s *stateObject) markSuicided() {
 
 // AddBalance adds amount to s's balance.
 // It is used to add funds to the destination account of a transfer.
-func (s *stateObject) AddBalance(amount *big.Int) {
+func (s *stateObject) AddBalance(amount *uint256.Int) {
+	// amount = types.AdjustExtraDecimalsUint256(amount)
```
Eric-Warehime: Remove this

x/vm/statedb/state_object.go
```diff
@@ -100,27 +108,29 @@ func (s *stateObject) markSuicided() {
...
 	}
-	s.SetBalance(new(big.Int).Add(s.Balance(), amount))
+	s.SetBalance(new(uint256.Int).Add(s.Balance(), amount))
 }
 
 // SubBalance removes amount from s's balance.
 // It is used to remove funds from the origin account of a transfer.
-func (s *stateObject) SubBalance(amount *big.Int) {
+func (s *stateObject) SubBalance(amount *uint256.Int) {
+	// amount = types.AdjustExtraDecimalsUint256(amount)
```
Eric-Warehime: Not needed.

x/vm/statedb/statedb.go
```diff
@@ -413,46 +407,102 @@ func (s *StateDB) SetState(addr common.Address, key, value common.Hash) {
...
-	stateObject.account.Balance = new(big.Int)
+	stateObject.account.Balance = new(uint256.Int)
+}
+
+func (s *StateDB) Selfdestruct6780(addr common.Address) {
+	stateObject := s.getStateObject(addr)
+	if stateObject == nil {
+		return
+	}
+	s.SelfDestruct(addr)
```
Eric-Warehime, line 431: This is not equivalent to [upstream](https://github.com/ethereum/go-ethereum/blob/c5ba367eb6232e3eddd7d6226bfd374449c63164/core/state/statedb.go#L465). The `SelfDestruct6780` call should only run on contracts which were created in the same txn.

Eric-Warehime (approved):

x/vm/statedb/statedb.go
Eric-Warehime: I think we'll need `CreateContract` added in here--we should track that against 1.15 upgrade and leave as is.

vladjdk: Added this thread in the todos

Issue #131, opened 2025-05-11
Precompile authentication: tx.origin usages


zsystm:
## Problem

The `Vote` and `VoteWeighted` methods in the governance precompile rely on `tx.origin` to determine the caller's address.  

This pattern may have security or compatibility implications, particularly in the context of smart contract-based wallets or proxy contracts. It needs to be reviewed and either confirmed safe or replaced with a more appropriate mechanism.

## Closing criteria

- [ ] Review the implications of using `tx.origin` in vote-related methods.
- [ ] Decide whether to:
  - (a) Keep `tx.origin` with proper documentation, or
  - (b) Replace it with `msg.sender` or an explicitly passed address.
- [ ] Update implementation and tests accordingly if changes are needed.

cc: @dudong2

PR #183, opened 2025-05-24
feat!: update precompile authentication

zsystm:
# Description

Only EOA→precompile calls are now allowed. Any proxy in the call path will always revert.

Closes: #131 

### Authentication hardening

All precompiles now require `msg.sender == requester`.

1. **Distribution precompile**
  - `ClaimRewards`, `SetWithdrawAddress`, `WithdrawDelegatorReward`, `WithdrawValidatorCommission`, `FundCommunityPool`, `DepositValidatorRewardsPool` → direct-only
2. **Evidence precompile**
  - `SubmitEvidecne` → direct-only
    - **BREAKING CHANGE:** First argument becomes submitter address.
3. **Governance precompile**
  - `SubmitProposal`, `Deposit`, `CancelProposal`, `Vote`, `VoteWeighted`  → direct-only
4.  **ICS20 precompile**
  - `Transfer`  → direct-only
5. **Slashing precompile**
  - `Unjail`  → now enforces `msg.sender == validatorAddress`

### Balance change journal

Removed the if contract.CallerAddress != origin { … } guard so that `SetBalanceChangeEntries` runs for all direct calls as well. This simplifies the logic and avoids confusion.

### Staking Precompile Contract Check
Previously, CreateValidator and EditValidator forbade contract accounts by checking msgSender != origin. Now we detect “contract” by the presence of on-chain bytecode at the caller address.

Rationale: under EIP-7702, an EOA could carry code, so bytecode-presence is a more reliable proxy for “isContract.”

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


precompiles/distribution/integration_test.go
```diff
@@ -1664,28 +1718,33 @@ var _ = Describe("Calling distribution precompile from another contract", Ordere
...
 					if tc.withdrawer != nil {
-						expDelFinalBalance = delInitialBalance.Amount.Sub(fees).Add(contractTransferredAmt)
 						expWithdrawerFinalBalance := withdrawerInitialBalance.Amount.Add(expRewards)
 						// withdrawer balance should have the rewards
 						balRes, err = s.grpcHandler.GetBalanceFromBank(tc.withdrawer.Bytes(), s.bondDenom)
 						Expect(err).To(BeNil())
 						withdrawerFinalBalance := balRes.Balance
 						Expect(withdrawerFinalBalance.Amount).To(Equal(expWithdrawerFinalBalance), "expected final balance to be greater than initial balance after withdrawing rewards")
+					} else { // contract is the withdrawer
+						// contract balance be updated according to the transferred amountn and rewards
```
vladjdk: nit: typo in amount

precompiles/distribution/integration_test.go
```diff
@@ -2795,54 +2461,16 @@ var _ = Describe("Calling distribution precompile from another contract", Ordere
...
-				fees := gasPrice.Mul(math.NewInt(res.GasUsed))
-				balRes, err := s.grpcHandler.GetBalanceFromBank(s.keyring.GetAccAddr(0), s.bondDenom)
-				Expect(err).To(BeNil())
-				signerFinalBalance := balRes.Balance
-				Expect(signerFinalBalance.Amount).To(Equal(signerInitialBalance.Amount.Sub(math.NewIntFromBigInt(depositAmt)).Sub(fees)))
-			})
-		})
-
-		When("depositor is different from signer", func() {
+		When("depositor is not a contract", func() {
```
vladjdk: Let's reword this to be "depositor is different from the depositing contract", because this implies that we should fail out if the depositor is not a contract

precompiles/staking/tx.go
```diff
@@ -223,9 +221,9 @@ func (p Precompile) Undelegate(
 		),
 	)
 
-	// The provided delegator address should always be equal to the contract caller address.
-	if contract.CallerAddress != delegatorHexAddr {
-		return nil, fmt.Errorf(ErrDifferentCallerFromDelegator, contract.CallerAddress.String(), delegatorHexAddr.String())
+	msgSender := contract.CallerAddress
```
vladjdk, line 223: Remove the unused tx.origin in the function params for all the functions here.

vladjdk (changes requested):

precompiles/testutil/contracts/DistributionCaller.sol
cloudgray, line 1: Due to the changes introduced in this PR, a significant portion of the existing tests has become unnecessary, but it seems that only some of them have been updated.

In particular, the methods with and without the `FromContract` postfix have effectively become functionally identical.
For example, `testSetWithdrawAddressFromContract` on line 12 and `testSetWithdrawAddress` on line 32 are now essentially the same method.

It might be a good idea to clean up and consolidate these parts accordingly.

This also applies to GovCaller.sol.

precompiles/distribution/integration_test.go
cloudgray, line 1: Entirely, keyword "origin" is included in test description.
I think it should be modified.

cloudgray: This also applies to other integration tests.

cloudgray (commented):
I left 2 comments!

precompiles/distribution/integration_test.go
zsystm: applied in https://github.com/cosmos/evm/pull/183/commits/2558c9ad651931bb7f9733fdf28b677ccb0103a3

precompiles/distribution/integration_test.go
zsystm: applied in https://github.com/cosmos/evm/pull/183/commits/2558c9ad651931bb7f9733fdf28b677ccb0103a3

precompiles/testutil/contracts/DistributionCaller.sol
zsystm: Thanks for pointing this out! While the methods may appear functionally identical at a glance, they are actually testing slightly different behaviors.

In `testSetWithdrawAddressFromContract`, we explicitly pass `address(this)` as the delegator address, which simulates a direct call from the contract itself. On the other hand, `testSetWithdrawAddress` takes an arbitrary address as input, allowing us to verify whether the precompile enforces direct calls only (i.e., the delegator must be the caller) or supports proxy-style calls.

This distinction is intentional, as we're trying to validate the authentication logic in the precompile under different call scenarios. So while they look similar, I believe both are still valuable and shouldn't be consolidated at this point.

precompiles/distribution/integration_test.go
zsystm: applied in https://github.com/cosmos/evm/pull/183/commits/3a02e0a37f6d780d5d8f5f10e1ee842be112554f

precompiles/testutil/contracts/DistributionCaller.sol
cloudgray: agree! I didn't consider test function for failure case.

precompiles/staking/tx.go
zsystm: applied in https://github.com/cosmos/evm/pull/183/commits/1284fbde2ba3fd2c6655896cdf4ec98705f55a4d

PR #188, opened 2025-05-26
fix: add IBC-Go Transfer App Module Migraton for V6

vladjdk:
closes: #187


x/ibc/transfer/module.go
```diff
@@ -55,4 +55,8 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(types.ModuleName, 4, m.MigrateDenomMetadata); err != nil {
 		panic(fmt.Errorf("failed to migrate transfer app from version 4 to 5 (set denom metadata migration): %v", err))
 	}
+
+	if err := cfg.RegisterMigration(types.ModuleName, 5, m.MigrateDenomTraceToDenom); err != nil {
+		panic(fmt.Errorf("failed to migrate transfer app from version 5 to 6 (migrate DenomTrace to Denom): %v", err))
+	}
```
Eric-Warehime, line 61: I think we should just remove all of these for v1--there's a bunch of breaking changes that make no sense in the context of an existing chain. Ok to ship this for now though--there's pretty much no harm.

PR #198, opened 2025-06-04
tests: refactoring

zsystm:
## Refactor Integration Tests and Modularize evmd Package

This PR introduces a major restructuring of the Cosmos EVM testing infrastructure to improve modularity, reusability, and maintainability.

### Summary
- Modularized `evmd` into a standalone package (`github.com/cosmos/evm/evmd`)
- Refactored integration test suites to depend on an abstract `EvmApp` interface
- Reorganized test structure for easier consumption by downstream client chains
- Removed cyclic dependencies and cleaned up unused files

### Key Changes
`evmd` **Modularization**
- `evmd` is now a standalone package that imports from `evm`, but never the other way around.
- This guarantees a clean and unidirectional dependency graph, ensuring that core logic remains encapsulated within `evm`.

**Interface-Based Testing**
- Introduced an `EvmApp` interface to represent the application under test.
- All integration test suites now depend on this interface rather than directly referencing `evmd`.
- This allows any client chain to reuse the test suites by implementing `EvmApp`.

**Test Structure Restructuring**
- Integration tests that use both `network` and `evmd` were moved to `evm/tests/integration`.
- Test file names were changed from `*_test.go` to `test_*.go` to allow proper compilation and import from other packages.
  - Go ignores `*_test.go` files when importing, so this rename enables downstream reuse.

**Internal Refactoring**
- Changed `backend` fields to public for easier access during tests.
  - If security is a concern, we may introduce getters.
- Removed unused test files and redundant code.

**Outcome**
- Enables downstream chains to reuse Cosmos EVM integration tests with minimal changes
- Ensures modular and decoupled architecture between `evm` and `evmd`
- Improves test suite maintainability and extensibility moving forward

Closes: #33 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


tests/integration/ante/test_evm_unit_04_validate.go
```diff
@@ -250,12 +250,12 @@
...
 	} {
 		for _, tc := range testCases {
-			suite.Run(fmt.Sprintf("%s, %s", chainID.ChainID, tc.name), func() {
+			s.Run(fmt.Sprintf("%s, %s", chainID.ChainID, tc.name), func() {
 				// Call the configurator to set the EVM coin required for the
 				// function to be tested.
 				configurator := evmtypes.NewEVMConfigurator()
 				configurator.ResetTestConfig()
-				suite.Require().NoError(configurator.WithEVMCoinInfo(testconstants.ExampleChainCoinInfo[chainID]).Configure())
+				s.Require().NoError(configurator.WithEVMCoinInfo(testconstants.ExampleChainCoinInfo[chainID]).Configure())
```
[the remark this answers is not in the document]
zsystm: It is reachable when you run the test with `-tags=test` flag.
proof:
image

evmd/go.mod
```diff
@@ -0,0 +1,280 @@
+module cosmosevm.io/evmd
```
Eric-Warehime: I prefer to maintain the github based url.

interfaces.go
Eric-Warehime, line 1: Ideally we can just declare the interface wherever we need to use it. Are you doing it here because it's just used everywhere?

go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/cosmos/evm
 go 1.23.8
 
 require (
+	cosmosevm.io/evmd v0.0.0
```
Eric-Warehime: I thought the idea was to remove this?
> evmd is now a standalone package that imports from evm, but never the other way around.

evmd/go.mod
zsystm: @Eric-Warehime 
How about using `github.com/cosmos/evm/evmd` instead?
Happy to hear any other suggestions as well if you have something else in mind!

go.mod
zsystm: @Eric-Warehime 
You're right — the ultimate goal is to remove evmd from evm entirely.
However, there are a few more changes needed before we can fully eliminate it.

For example, we'll need to:
- Move `evm/cmd` to `evm/evmd/cmd`/
- Relocate shared data structures currently under `evmd` into a more appropriate place within `evm`.

I considered including these changes in the testing refactoring PR, but it felt like too much to bundle into a single PR. That's why I thought it would be cleaner to address the removal of `evmd` in a follow-up PR.

Would you prefer we fully remove the `evmd` dependency as part of this PR?
Or does it make more sense to merge the testing refactoring first, and then handle the cleanup/removal in a separate one?

Happy to align either way.

interfaces.go
zsystm: @Eric-Warehime 
That’s a fair point.

Currently, `EvmApp` is used across quite a few directories. While it’s technically possible to define the interface locally wherever it's needed, I thought it made sense to keep it in `evm/interfaces.go` since it serves as a kind of reference interface for chains that want to integrate with `cosmos/evm`, especially when running the tests under `tests/integration`.

Putting it directly under `cosmos/evm` felt like a clear signal that it’s the standard interface contract for client chains implementing `EvmApp`.

That said, I’m not strongly opinionated here — if you still feel it’s better to colocate the interface with each usage site, I’m happy to revisit that direction.

evmd/go.mod
Eric-Warehime: Yeah that's fine. I think we just don't like the confusion that the `github.com/cosmos/cosmos-sdk` and `cosmossdk.io` naming has caused, so it's best to stick with the github prefixes here.

go.mod
Eric-Warehime: Ok, happy to do things in separate PRs since this is already large in terms of lines of code.

evmd/go.mod
zsystm: @Eric-Warehime 
Applied in https://github.com/cosmos/evm/pull/198/commits/e7377b4b6ffb1182b907b3d02e8713a450e4bb58

go.mod
zsystm: Created separate issue for following up this: https://github.com/cosmos/evm/issues/211

Issue #223, opened 2025-06-17
Precompile Errors Not Propagated to the Caller Contract


cloudgray:
## Problem

- When the caller contract calls the precompile, the error message isn’t propagated back to it.
- Unlike the geth opcode `opRevert`, which returns the error message in its return data and `ErrExecutionReverted` as the error, the Cosmos/EVM precompile currently returns `nil` as return data and a custom error type as the error.
- We need to make precompile errors behave like `opRevert`: return the error message in the return data and use `ErrExecutionReverted` as the error.

## Closing criteria

- [ ] When a contract calls a precompile, the returned error is `“execution reverted”` and the EthereumTxResponse.Ret should contain the byte-encoded revert reason.

### Expected Behavior

![Image](https://github.com/user-attachments/assets/284ae68b-9553-43c1-ab1c-1ce378ebf1c4)

- In general, when one Solidity contract calls another, error messages generated in the callee contract are propagated back to the caller contract.
- The example above illustrates this behavior: when a caller contract invokes the `permit` method of a Token Contract (inheriting from [ERC2612Permit.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/token/ERC20/extensions/ERC20Permit.sol)), the “invalid signature” error thrown by `permit` is propagated to the caller contract.

### Actual Behaviour

<img width="906" alt="Image" src="https://github.com/user-attachments/assets/212bb9a0-c927-4244-8fa5-c7c56b51d140" />

- However, the situation is different when the callee contract is a Cosmos/EVM precompile. Developers expect to see the actual error message in the response when a revert is caused by a precompile error, but in reality only the `"execution revert"` message is returned.
- This issue does not occur when an EOA calls the precompile directly—it only happens when the call is made via a caller contract.

### How to Reproduce Revert

```solidity
function testClaimRewards(
    address _delAddr,
    uint32 _maxRetrieve
) public returns (bool success) {
    return
        distribution.DISTRIBUTION_CONTRACT.claimRewards(
        _delAddr,
        _maxRetrieve
    );
}
```

- When calling `testClaimRewards` in precompiles/testutil.contracts/DistributionCall.sol, passing `senderAddr` instead of `contractAddr` to `_delAddr` triggers an `ErrRequesterIsNotMsgSender` error.
- However, in the returned `EthereumTxResponse`, you cannot retrieve the error message from `ErrRequesterIsNotMsgSender`.

### Root Cause: `opCall` doesn't return error

[...]

PR #224, opened 2025-06-17
fix(precompiles): return data for revert

cloudgray:
# Description

## Problem

- When the caller contract calls the precompile, the error message isn’t propagated back to it.
- Unlike the geth opcode opRevert, which returns the error message in its return data and ErrExecutionReverted as the error, the Cosmos/EVM precompile currently returns nil as return data and a custom error type as the error.
- We need to make precompile errors behave like opRevert: return the error message in the return data and use ErrExecutionReverted as the error.

## Solution

1. Modify return values on precompile error
    - Always return `ErrExecutionReverted` as the error.
    - Encode the original error message with the Revert selector (`"Error(string)"`) as the return data.
    - Use `evm.Interpreter.SetReturnData()` to set that encoded data into `EVMInterpreter.returnData`.
2. Modify return values in `ApplyMessageWithConfig`
    - If err is `ErrExecutionReverted`, set res.Ret to `evm.Interpreter.ReturnData()`.
3.  Update the test util function
    - Previously, on error the code logged the revert reason via emitted events (which never occur when an error is thrown).
    - Now, pull the `EthereumTxResponse.Ret` returned in step 2, decode it with `abi.UnpackRevert`, and print that as the reason.

### As-Is

err_msg_empty
In the original code, errors thrown by the precompile were not propagated to the caller contract.

### To-Be

revert_with_reason
Now, errors returned by the precompile are returned in the caller contract’s return data, and the test utilities have been updated so that the test code can verify them.

Closes: #223 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content

### DIFF
--- a/precompiles/common/revert.go
+++ b/precompiles/common/revert.go
@@ -1,0 +1,26 @@
+package common
+
+import (
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+var (
+	revertSelector = crypto.Keccak256([]byte("Error(string)"))[:4]
+)
+
+// RevertReasonBytes converts a message to ABI-encoded revert bytes.
+func RevertReasonBytes(reason string) ([]byte, error) {
+	typ, err := abi.NewType("string", "", nil)
+	if err != nil {
+		return nil, err
+	}
+	packed, err := (abi.Arguments{{Type: typ}}).Pack(reason)
+	if err != nil {
+		return nil, err
+	}
+	bz := make([]byte, 0, len(revertSelector)+len(packed))
+	bz = append(bz, revertSelector...)
+	bz = append(bz, packed...)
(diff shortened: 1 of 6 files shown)


tests/integration/precompiles/staking/test_integration.go
```diff
@@ -2001,12 +2012,19 @@ func TestPrecompileIntegrationTestSuite(t *testing.T, create network.CreateEvmAp
 							s.keyring.GetAddr(0), valAddr.String(), true, true,
 						}
 
-						txArgs.Amount = delAmt.BigInt()
+						txArgs.To = &contractTwoAddr
+
+						reverReasonCheck := execRevertedCheck.WithErrContains(
```
vladjdk, line 2017: nit: revert typo

vladjdk (approved):
This is great, thank you!

PR #226, opened 2025-06-17
feat: preinstalls

gjermundgaraba:
I want to discuss how we add this, and if we want any temporary stuff or how we should do this.

In the current code I have the following:
- A method on the vm keeper that takes in a list of preinstalls and sets them
- Calling said method from InitGenesis with a list of types.DefaultPreinstalls (this could be changed to making it actually configurable in Genesis, as well)
- A store migration from 9 to 10 that also adds the same default list (I've not added tests for this yet, because I want to answer the questions below first)

The questions I have are:
1. What mechanism do we want to use for this?
  a. Eric mentioned maybe wanting a cli command to add pre-installs. I could do that, but it seems like temporary code we might now want in the evm repo
  b. If this is primarily to add them to the running testnet, I would suggest just changing the gaia binary being used to call the `keeper.Add
2. How configurable do we need this to be right now?

### DIFF
--- a/tests/integration/x/vm/test_iterate_contracts.go
+++ b/tests/integration/x/vm/test_iterate_contracts.go
@@ -52,6 +52,7 @@ func TestIterateContracts(t *testing.T, create network.CreateEvmApp, options ...
 	var (
 		foundAddrs  []common.Address
 		foundHashes []common.Hash
+		addrToHash  = make(map[common.Address]common.Hash)
 	)
 
 	network.App.GetEVMKeeper().IterateContracts(network.GetContext(), func(addr common.Address, codeHash common.Hash) bool {
@@ -62,12 +63,20 @@ func TestIterateContracts(t *testing.T, create network.CreateEvmApp, options ...
 
 		foundAddrs = append(foundAddrs, addr)
 		foundHashes = append(foundHashes, codeHash)
+		addrToHash[addr] = codeHash
 		return false
 	})
 
-	require.Len(t, foundAddrs, 2, "expected 2 contracts to be found when iterating")
+	require.Len(t, foundAddrs, 6, "expected 6 contracts to be found when iterating (4 preinstalled + 2 deployed)")
 	require.Contains(t, foundAddrs, contractAddr, "expected contract 1 to be found when iterating")
 	require.Contains(t, foundAddrs, contractAddr2, "expected contract 2 to be found when iterating")
-	require.Equal(t, foundHashes[0], foundHashes[1], "expected both contracts to have the same code hash")
-	require.NotEqual(t, types.EmptyCodeHash, foundHashes[0], "expected store code hash not to be the keccak256 of empty code")
+
+	// Get the code hashes for our deployed contracts
+	hash1, exists1 := addrToHash[contractAddr]
+	require.True(t, exists1, "expected to find code hash for contract 1")
+	hash2, exists2 := addrToHash[contractAddr2]
+	require.True(t, exists2, "expected to find code hash for contract 2")
+
+	require.Equal(t, hash1, hash2, "expected both deployed contracts to have the same code hash")
+	require.NotEqual(t, types.EmptyCodeHash, hash1, "expected store code hash not to be the keccak256 of empty code")
 }


x/vm/types/preinstall.go
```diff
@@ -0,0 +1,27 @@
+package types
+
+import "github.com/ethereum/go-ethereum/common"
+
+type Preinstall struct {
+	Address common.Address
```
vladjdk: Could we add names here to keep track of what these are?

x/vm/keeper/migrations.go
```diff
@@ -0,0 +1,24 @@
...
+
+// NewMigrator returns a new Migrator.
+func NewMigrator(keeper *Keeper) Migrator {
+	return Migrator{keeper: keeper}
+}
+
+// Migrate9to10 migrates the x/vm module state from the consensus
+// version 9 to version 10. Specifically, it deploys a set of predefined
+// preinstall contracts with a fixed addresses and code.
+func (m Migrator) Migrate9to10(ctx sdk.Context) error {
```
vladjdk: We're gonna reset all module versions to 1, so let's remove this or move it to some example upgradehandler. I think we should keep the preinstalls keeper method though

x/vm/module.go
```diff
@@ -127,6 +127,11 @@ func (AppModule) Name() string {
 func (am AppModule) RegisterServices(cfg module.Configurator) {
 	types.RegisterMsgServer(cfg.MsgServer(), am.keeper)
 	types.RegisterQueryServer(cfg.QueryServer(), am.keeper)
+
+	m := keeper.NewMigrator(am.keeper)
+	if err := cfg.RegisterMigration(types.ModuleName, 9, m.Migrate9to10); err != nil {
```
vladjdk: Same as above - I want to avoid having migrations for modules before v1

x/vm/genesis.go
```diff
@@ -57,6 +57,10 @@ func InitGenesis(
 		}
 	}
 
+	if err := k.AddPreinstalls(ctx, types.DefaultPreinstalls); err != nil {
```
vladjdk: Could we add this into the evmd instead? I'm thinking that we could pass in the default preinstalls as a default there, but allow anyone to specify their own precompiles upon genesis inside their app

vladjdk (changes requested):

x/vm/types/preinstall.go
gjermundgaraba: Yeah, I think that is a good idea. I will be doing that!

x/vm/keeper/migrations.go
gjermundgaraba: Even better, will do that 🚀

x/vm/genesis.go
gjermundgaraba: Yeah for sure, I also felt there was something slightly off about this, so I will take another look to make this more reasonable.

x/vm/types/preinstall.go
gjermundgaraba: Done!

x/vm/keeper/migrations.go
gjermundgaraba: Done!

x/vm/module.go
gjermundgaraba: Done!

x/vm/genesis.go
gjermundgaraba: I've made this change as well, let me know if it is what you had in mind.

I haven't really tested it yet, is there a good way to test evmd genesis that I should be doing beyond what I have added unit tests for?

gjermundgaraba:
I believe I've made the required changes, so putting this in ready-for-review.
I also linted the proto files, so got some extra changes from that. I hope that's OK, if not I can revert that commit.

x/vm/keeper/preinstalls.go
```diff
@@ -0,0 +1,50 @@
...
+		if !types.IsEmptyCodeHash(existingCodeHash.Bytes()) && !bytes.Equal(existingCodeHash.Bytes(), codeHash) {
+			return errorsmod.Wrapf(types.ErrInvalidPreinstall, "preinstall %s already has a code hash with a different code hash", preinstall.Address)
+		}
+
+		// check that the account is not already set
+		if exists := k.accountKeeper.HasAccount(ctx, accAddress); !exists {
+			// return errorsmod.Wrapf(types.ErrInvalidPreinstall, "preinstall %s already has an account in account keeper", preinstall.Address.String())
+			// create account with the account keeper
+			account := k.accountKeeper.NewAccountWithAddress(ctx, accAddress)
+			k.accountKeeper.SetAccount(ctx, account)
```
Eric-Warehime, line 40: Why do we not want to error when an account exists--like you have in the comment?

gjermundgaraba: Oh yeah, this was actually something I hadn't fully finished thinking through, but the initial thinking was that this would be problematic these already existed. After the changes, this should not really happen, I think.

I'll have another look at this part!

gjermundgaraba: It is also an option to check here if this already exists, and then also check if the given account and code all match, and then just ignore and continue. Not sure if that is a legit scenario though 🤔

gjermundgaraba: Oh yeah, I might have done this because all the integration tests fail with this. It seems like they don't reset the account keeper or something.

I'm a bit lost with this integration test suite. Any pointers from you or @vladjdk on this?

x/vm/module.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 // consensusVersion defines the current x/evm module consensus version.
-const consensusVersion = 9
+const consensusVersion = 10
```
Eric-Warehime, line 30: Let's leave this--I think we're going to reset this for the production release since I don't think anyone is actually running module version 9.

x/vm/keeper/preinstalls.go
Eric-Warehime: Which tests? All integ tests are passing with this diff
Screenshot 2025-06-26 at 12 05 32 PM

Eric-Warehime (approved):
I'm going to merge this and follow up on a couple of things in a separate PR.
Thanks @gjermundgaraba !

Issue #211, opened 2025-06-12
Complete Decoupling of evm and evmd by Relocating Remaining Dependencies


zsystm:
## Description

The recent #198 introduced a major refactoring of the test infrastructure and modularized the evmd package. 
However, a full decoupling between `evm` and `evmd` has not yet been completed. This issue tracks the remaining work needed to fully eliminate the dependency from `evm` to `evmd`.

## Goal

Ensure a **clean and intentional one-way dependency**: evmd → evm, and **not the reverse**.
This design reflects the intended layering:
- `evm`: a library providing core Cosmos EVM functionality.
- `evmd`: a consumer of evm, showcasing a concrete example application and testbed.

By removing all remaining `evm → evmd` dependencies, we:
- Reinforce modular boundaries
- Make `evm` reusable as a standalone library
- Avoid cyclical or unintended coupling
- Simplify maintenance and auditing

## Remaining Tasks

To achieve a clean unidirectional dependency graph, the following actions are required:

- [ ] Move CLI entry point: Relocate evm/cmd into evm/evmd/cmd/ to encapsulate all executable logic within evmd.
- [ ] Relocate shared data structures: Move any structs or logic currently defined in `evmd` that are used by `evm` into a more appropriate location within `evm`, or abstract into a shared internal package if necessary.

PR #229, opened 2025-06-18
refactor: fully decouple evm from evmd

zsystm:
## Description

This PR finalizes the decoupling between the `evm` and `evmd` packages by ensuring a strict one-way dependency from `evmd → evm`, and eliminating all reverse dependencies.

## Key Changes
- **Moved shared logic**: Files previously located in evmd but used by evm were moved into appropriate evm subpackages.
- **Relocated evmd-specific tests**: tests that directly rely on `evmd` internals and not using Network harness which is used from integraion tests were moved into the evmd package.
- **Interface-based refactoring**: Tests that unnecessarily depended on `evmd` were updated to use the `EvmApp` interface, allowing them to remain within `evm` while preserving abstraction.
- **Updated Makefile**: Adjusted build and install targets to reflect the relocation of cmd/ under evmd.

## Motivation

- Enforces a clean architectural boundary between `evm` (library) and `evmd` (application).
- Enables `evm` to be consumed independently by downstream chains.
- Reduces coupling, simplifies maintainability, and aligns with long-term modularization goals.

Closes: #211 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


config/chain_id.go
vladjdk, line 1: I would prefer that we have two copies of these (evm_app_options, constants, config, activators). One in the test files, and one in evmd. The reason is that these are app-specific config options and it should be clear that any chain implementation needs to implement their own. If we include it in EVMD, then evmd-specific configs will be available via the package, which is confusing imo

config/evmd_config.go
vladjdk, line 1: Likewise, an evmd config shouldn't be in the cosmos/evm files

evmd/test_helpers.go
```diff
@@ -43,23 +43,23 @@ func init() {
...
+func setup(withGenesis bool, invCheckPeriod uint, chainID string, evmChainID uint64) (*EVMD, evmconfig.GenesisState) {
 	db := dbm.NewMemDB()
 
 	appOptions := make(simtestutil.AppOptionsMap, 0)
-	appOptions[flags.FlagHome] = DefaultNodeHome
+	appOptions[flags.FlagHome] = defaultNodeHome
 	appOptions[server.FlagInvCheckPeriod] = invCheckPeriod
 
-	app := NewExampleApp(log.NewNopLogger(), db, nil, true, appOptions, evmChainID, EvmAppOptions, baseapp.SetChainID(chainID))
+	app := NewExampleApp(log.NewNopLogger(), db, nil, true, appOptions, evmChainID, evmconfig.EvmAppOptions, baseapp.SetChainID(chainID))
```
vladjdk: App options should be within the evmd files

ante/ante.go
vladjdk, line 1: I think these ante handlers should also remain in evmd - they're custom to every chain and I don't think they should be set as a default

vladjdk (changes requested):
Overall, I would prefer to have app-specific configs (and the ante, somewhat) contained within EVMD. If we need to duplicate them for testing, we should do that. I mostly don't feel comfortable having the configs and options as dependencies that people could use within Cosmos EVM. They likely won't be entirely conforming to the specs each chain have.

config/chain_id.go
zsystm: Totally agree — that's a great point, thanks for highlighting it.
I've placed the test-related configs under `testutil/config`, and the ones used by evmd under `cmd/evmd/config`.
This should help prevent any confusion for client chains integrating with evm.

config/evmd_config.go
zsystm: Absolutely — I’ve moved the evmd-specific configs out of `cosmos/evm` and into `cmd/evmd/config`, where they belong.
This separation should help maintain a clean boundary between the core evm package and app-specific configurations.

evmd/test_helpers.go
zsystm: Yes — I’ve moved the app options into `cmd/evmd/config` as suggested.
Thanks again for pointing this out!

ante/ante.go
zsystm: Good point — I’ve moved the ante handlers under evmd/ante since they’re specific to the app and shouldn’t be treated as defaults. Thanks for the suggestion!

PR #230, opened 2025-06-18
chore: clean up deps

aljo242:
- clean up minor deps
- fix lint issue with deprecation of crisis
- no longer rely on `x/exp` and instead use packages native from `Go` stdlib


testutil/integration/evm/network/setup.go
```diff
@@ -157,7 +155,7 @@ func createBalances(
 
 // createStakingValidator creates a staking validator from the given tm validator and bonded
 func createStakingValidator(val *cmttypes.Validator, bondedAmt sdkmath.Int, operatorAddr *sdktypes.AccAddress) (stakingtypes.Validator, error) {
-	pk, err := cryptocodec.FromTmPubKeyInterface(val.PubKey) //nolint:staticcheck
+	pk, err := cryptocodec.FromCmtPubKeyInterface(val.PubKey)
```
aljo242, line 158: FromTmPubKeyInterface just calls FromCmtPubKeyInterface under the hood

Issue #235, opened 2025-06-20
[Bug]: opBlockhash is not aligned with geth v1.15


mmsqe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

return empty [opBlockhash](https://github.com/ethereum/go-ethereum/blob/release/1.15/core/vm/instructions.go#L431) when `block protocol is incorrect` error occurs, [empty](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/c9f1e115b388595975d97bb954a29970a34c08bc/integration_tests/test_basic.py#L501) block header provided from [staking](https://github.com/cosmos/evm/blob/0e511d32206b1ac709a0eb0ddb1aa21d29e833b8/x/vm/keeper/state_transition.go#L113) module

### Cosmos EVM Version

main

### How to reproduce?

[compare](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/c9f1e115b388595975d97bb954a29970a34c08bc/integration_tests/test_basic.py#L501) blockhash result with current block results fails

technicallyty:
what causes "block protocol is incorrect"? cc @yihuang

yihuang:
> what causes "block protocol is incorrect"? cc [@yihuang](https://github.com/yihuang)
> 
> why is the block version not matching what comet asserts?

sorry for the confusion.

the root cause is the header stored by staking module is come from the context, which is initialized [here](https://github.com/cosmos/cosmos-sdk/blob/908df9d4e2e07c2cd923d7b35f18b1e008c5106c/baseapp/abci.go#L722), it's not complete, and will cause error in this [conversion here](https://github.com/cosmos/evm/blob/0e511d32206b1ac709a0eb0ddb1aa21d29e833b8/x/vm/keeper/state_transition.go#L119). That's why we have to explicitly store the block hash in evm states.

PR #236, opened 2025-06-20
fix: block header hash not available in staking module

yihuang:
Closes: #235

Solution:
- store in evm module by itself

# Description

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content

### DIFF
--- a/x/vm/keeper/abci.go
+++ b/x/vm/keeper/abci.go
@@ -10,6 +10,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+const DefaultHeaderHashNum = int64(10000)
+
 // BeginBlock emits a base fee event which will be adjusted to the evm decimals
 func (k *Keeper) BeginBlock(ctx sdk.Context) error {
 	logger := ctx.Logger().With("begin_block", "evm")
@@ -31,6 +33,13 @@ func (k *Keeper) BeginBlock(ctx sdk.Context) error {
 			),
 		})
 	}
+
+	// store block header hash and rotate the old ones.
+	k.SetHeaderHash(ctx)
+	if i := ctx.BlockHeight() - DefaultHeaderHashNum; i > 0 {
+		k.DeleteHeaderHash(ctx, i)
+	}
+
 	return nil
 }
 
--- a/x/vm/keeper/keeper.go
+++ b/x/vm/keeper/keeper.go
@@ -351,3 +351,21 @@ func (k Keeper) AddTransientGasUsed(ctx sdk.Context, gasUsed uint64) (uint64, er
 	k.SetTransientGasUsed(ctx, result)
 	return result, nil
 }
+
+// SetHeaderHash stores the hash of the current block header in the store.
+func (k Keeper) SetHeaderHash(ctx sdk.Context) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetHeaderHashKey(ctx.BlockHeight()), ctx.HeaderHash())
+}
+
+// GetHeaderHash retrieves the hash of a block header from the store by height.
+func (k Keeper) GetHeaderHash(ctx sdk.Context, height int64) []byte {
+	store := ctx.KVStore(k.storeKey)
+	return store.Get(types.GetHeaderHashKey(height))
+}
+
+// DeleteHeaderHash removes the hash of a block header from the store by height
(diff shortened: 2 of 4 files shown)


x/vm/keeper/abci.go
```diff
@@ -10,6 +10,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
+const DefaultHeaderHashNum = int64(10000)
```
vladjdk: Can we add this as a param?

vladjdk (changes requested):
Mostly lgtm, let's just add the default max storage as a parameter

x/vm/keeper/abci.go
yihuang: done

vladjdk:
@yihuang tests failing

vladjdk:
@yihuang now lints are failing

yihuang:
> @yihuang now lints are failing

I tried both `gofmt -w` `gofumpt -w`, don't show any updates 🤔

yihuang:
> @yihuang now lints are failing

fixed

aljo242:
@technicallyty to review

x/vm/keeper/keeper.go
```diff
@@ -357,6 +359,25 @@ func (k Keeper) AddTransientGasUsed(ctx sdk.Context, gasUsed uint64) (uint64, er
 	return result, nil
 }
 
+// SetHeaderHash stores the hash of the current block header in the store.
+func (k Keeper) SetHeaderHash(ctx sdk.Context) {
+	store := ctx.KVStore(k.storeKey)
+	fmt.Println("HeaderHash", hex.EncodeToString(ctx.HeaderHash()))
```
technicallyty:
```suggestion
```

technicallyty:
i thought about this some more, and i think we may want to consider updating the SDK to store historical block hashes. imo it seems a little out of bounds for x/vm to be doing this work. checking in with team on this and will get back to you

yihuang:
> i thought about this some more, and i think we may want to consider updating the SDK to store historical block hashes. imo it seems a little out of bounds for x/vm to be doing this work. checking in with team on this and will get back to you

https://eips.ethereum.org/EIPS/eip-2935

another idea is to adopt the new method as go-ethereum, it's stored in a contract storage, so it's still kind of in the `x/evm` storage.

technicallyty:
> > i thought about this some more, and i think we may want to consider updating the SDK to store historical block hashes. imo it seems a little out of bounds for x/vm to be doing this work. checking in with team on this and will get back to you
> 
> https://eips.ethereum.org/EIPS/eip-2935
> 
> another idea is to adopt the new method as go-ethereum, it's stored in a contract storage, so it's still kind of in the `x/evm` storage.
> 
> This will requires to allow replaying non-eip-155 tx though, see: #314

ahh thanks for linking the EIP. im okay with proceeding here then.

x/vm/keeper/state_transition.go
```diff
@@ -86,47 +84,13 @@ func (k Keeper) GetHashFn(ctx sdk.Context) vm.GetHashFunc {
...
-				k.Logger(ctx).Error("failed to cast tendermint header from proto", "error", err)
-				return common.Hash{}
-			}
-
-			return common.BytesToHash(header.Hash())
-		default:
-			// Case 3: heights greater than the current one returns an empty hash.
-			return common.Hash{}
 		}
+		return common.BytesToHash(k.GetHeaderHash(ctx, h))
```
technicallyty, line 93: in geth, it only ever services the hashes in the range of 256 blocks. otherwise it would return 0. is this assertion important to hold here? 

https://github.com/ethereum/go-ethereum/blob/5cc9137c9cec8a977b845a025d3deced9457fd48/core/vm/instructions.go#L431-L459

yihuang: https://github.com/cosmos/evm/pull/236/files#diff-434db4099af9e9801db7c89d9f43095fb59e2d02dbfd899b525c27190d109468R34-R42

the rotation is handled at the store time.

but if we go with [eip-2935](https://eips.ethereum.org/EIPS/eip-2935), these logics will be in the contract.

yihuang: I opened PR to implement EIP-2935, I think this PR can be adapted to that, see https://github.com/cosmos/evm/issues/406

Issue #225, opened 2025-06-17
Research Possibility of Keeping CMS instead of Copy()


vladjdk:

yihuang:
https://github.com/cosmos/cosmos-sdk/pull/14444

Try to sell our previous work again here, it's designed specific to support statedb snapshot/revert scenario.

PR #244, opened 2025-06-24
feat(x/vm): apply stack-based StateDB snapshot mechamism for precompile call 

cloudgray:
# Description

## Summary
- Applying the stack-based snapshot mechanism achieves the same effect as a deep copy of the original store.
- Benchmark results show that the stack-based snapshot mechanism uses about 1.4× more memory than CacheMultiStore, but far less than a deep copy—and the gap grows larger as the load increases.

## Stack-based Snapshot Mechanism

- When executing a precompile, a new store layer is added to the top of the stack, and state writes are performed on this new layer.
- If a revert occurs during execution, the invalid store layers are popped off the stack to ensure that a valid store layer remains at the top.
- When committing changes, the writes are propagated sequentially from the top of the stack down to the bottom and applied to the initialStore.
- The `Snapshot()` method of the new store uses approximately 1.4 times more memory than the previously used `CacheMultiStore()`, but it provides a more robust structure that avoids edge cases.

## Benchmark

memory_usage_full_stack_based_updated

The benchmark results show that the new Snapshot() method uses significantly less memory than a deep copy, and the gap widens as the number of iterations increases.

cms_vs_snapshot_stack_based_updated

Like CacheMultiStore(), the new Snapshot() method exhibits linear memory growth with increasing iterations, using approximately 1.4× more memory than CacheMultiStore()

Closes: #225 

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


x/vm/statedb/statedb.go
```diff
@@ -163,29 +168,17 @@ func (s *StateDB) GetCacheContext() (sdk.Context, error) {
...
-func (s *StateDB) RevertMultiStore(cms storetypes.CacheMultiStore, events sdk.Events) {
-	s.cacheCtx = s.cacheCtx.WithMultiStore(cms)
+func (s *StateDB) RevertMultiStore(snapshot int, events sdk.Events) {
+	s.snapshotter.RevertToSnapshot(snapshot)
 	s.writeCache = func() {
-		// rollback the events to the ones
-		// on the snapshot
 		s.ctx.EventManager().EmitEvents(events)
-		cms.Write()
+		s.cacheCtx.MultiStore().(storetypes.CacheMultiStore).Write()
```
zsystm, line 181:
```suggestion
		s.cacheCtx.MultiStore().(*snapshotmulti.Store).Write()
```

x/vm/statedb/statedb.go
```diff
@@ -194,7 +187,21 @@ func (s *StateDB) cache() error {
...
+	cms := s.cacheCtx.MultiStore().(storetypes.CacheMultiStore)
+	storeKeys := s.keeper.KVStoreKeys()
+
+	// Create and set snapshot store to stateDB
+	snapshotStore := snapshotmulti.NewStore(cms, storeKeys)
+	s.snapshotter = snapshotStore
+	s.cacheCtx = s.cacheCtx.WithMultiStore(snapshotStore)
+	s.writeCache = func() {
+		s.ctx.EventManager().EmitEvents(s.cacheCtx.EventManager().Events())
+		s.cacheCtx.MultiStore().(storetypes.CacheMultiStore).Write()
```
zsystm, line 202:
```suggestion
		snapshotStore.Write()
```

zsystm (commented):
@cloudgray 
Since the actual type of `s.cacheCtx.MultiStore()` at this point is `*snapshotmulti.Store`, I think it's clearer to cast it explicitly rather than going through the `storetypes.CacheMultiStore` interface.

While `*snapshotmulti.Store` does implement `CacheMultiStore`, making that implicit assumption here could lead to confusion for future maintainers who aren’t aware of the underlying implementation. Being explicit improves readability and reduces ambiguity in understanding what store we’re operating on.

x/vm/store/snapshotmulti/store_bench_test.go
```diff
@@ -0,0 +1,70 @@
...
+	for i := 0; i < b.N; i++ {
+		for tx := 0; tx < txs; tx++ {
+			for j := 0; j < 10; j++ {
+				for _, key := range selected {
+					kv := cms.GetKVStore(key)
+					kv.Get([]byte(fmt.Sprintf("%s-key-%d", key.Name(), j%20)))
+					kv.Set([]byte(fmt.Sprintf("%s-tx-%d-%d", key.Name(), tx, j)), genBytes(32))
+				}
+				snapshot := cms.Snapshot()
+				benchSink = snapshot
```
zsystm, line 62: Just curious — what’s the purpose of benchSink here?

x/vm/store/snapshotmulti/store_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func setupStore() (*snapshotmulti.Store, *storetypes.KVStoreKey) {
+	key := storetypes.NewKVStoreKey("store")
+	kv := cachekv.NewStore(dbadapter.Store{DB: dbm.NewMemDB()})
+	stores := map[*storetypes.KVStoreKey]storetypes.CacheWrap{key: kv}
+	return snapshotmulti.NewStoreWithKVStores(stores), key
+}
+
+func TestSnapshotMultiIndexing(t *testing.T) {
+	cms, _ := setupStore()
```
zsystm: Minor naming suggestion: would it be possible to rename the `cms` variable to something like `ss` or `snapshotStore`?

Since `cms` often refers to `CacheMultiStore` in the SDK context, it might be a bit misleading here, especially if it’s actually a `snapshotmulti.Store`. Renaming it could help reduce confusion for readers familiar with the SDK naming conventions.

zsystm:
@cloudgray 
Left a couple of nits, but overall LGTM — nice work!

x/vm/statedb/statedb.go
cloudgray: This is the matter of meaning.

s.cacheCtx.ms and snapshotStore points to same reference, so as you said, if we use snapshotStore instead of `s.cacheCtx.MultiStore().(storetypes.CacheMultiStore)` the result will be same.

However, the reason I did this is because, I wanted to clearly write the the meaning that `CacheMultiStore` interface, `Write()` occurs.
If we use, `snapshotStore.Write()`, the developer who first read this code can think snapshot is written.
That is the situation I wanted to avoid.

x/vm/statedb/statedb.go
cloudgray: *snapshotmulti.Store is the type that we defined in cosmos/evm level.
But I want to conserve the original meaning that `CacheMultiStore interface`, `Write()` is called from `s.cacheCtx.MultiStore()`

x/vm/store/snapshotmulti/store_bench_test.go
cloudgray: This variable is a common technique used to prevent compiler optimizations during benchmarking.

```go
benchSink = snapshot
```

By assigning the snapshot result to the global variable `benchSink`, it prevents the compiler from treating the `cms.Snapshot()` call as a no-op and eliminating it. Since Go benchmarks run a loop b.N times to measure performance, if the result of an operation isn’t used, the Go compiler (GC) may optimize it away entirely.

Therefore, benchSink serves two purposes:

- It ensures the Snapshot operation is not omitted by actually using the result.
- It helps maintain benchmark accuracy under the same conditions as real-world code.

x/vm/store/snapshotmulti/store_test.go
cloudgray: applied! (https://github.com/cosmos/evm/pull/244/commits/df311f3c09760c9cca2a1c99084a5fb6392102a2)

tests/integration/precompiles/distribution/test_distribution.go
```diff
@@ -497,13 +497,18 @@ func (s *PrecompileTestSuite) TestCMS() {
 			if tc.expPass {
 				s.Require().NoError(err, "expected no error when running the precompile")
 				s.Require().NotNil(resp.Ret, "expected returned bytes not to be nil")
-				testutil.ValidateWrites(s.T(), cms, 2)
+				// NOTES: After stack-based snapshot mechanism is added for precopile call,
```
vladjdk: typo in precopile

tests/integration/precompiles/distribution/test_distribution.go
```diff
@@ -497,13 +497,18 @@ func (s *PrecompileTestSuite) TestCMS() {
...
+				// It is because CacheMultiStore() is not called when creating snapshot for MultiStore,
+				// Count of Write() is not accumulated.
+				testutil.ValidateWrites(s.T(), cms, 1)
 			} else {
 				s.Require().Error(err, "expected error to be returned when running the precompile")
 				s.Require().Nil(resp.Ret, "expected returned bytes to be nil")
 				s.Require().ErrorContains(err, tc.errContains)
-				// Writes once because of gas usage
-				testutil.ValidateWrites(s.T(), cms, 1)
+				// NOTES: After stack-based snapshot mechanism is added for precopile call,
```
vladjdk: same as above

vladjdk: Another nit: this test is called TestCMS but we're now using a Cache Stack

tests/integration/precompiles/staking/test_integration.go
```diff
@@ -1834,6 +1834,54 @@ func TestPrecompileIntegrationTestSuite(t *testing.T, create network.CreateEvmAp
...
+						callArgs := testutiltypes.CallArgs{
+							ContractABI: stakingReverterContract.ABI,
+							MethodName:  "callPrecompileBeforeAndAfterRevert",
+							Args: []interface{}{
+								big.NewInt(5), s.network.GetValidators()[0].OperatorAddress,
+							},
+						}
+
+						delegateCheck := passCheck.WithExpEvents(staking.EventTypeDelegate, staking.EventTypeDelegate)
+						// Tx should be successful, but no state changes happened
```
vladjdk: What do you mean by no state changes? Isn't there a delegation being tested here?

vladjdk: After running through the tests, I'm going to assume you mean no unintended state changes from the reversions

x/vm/keeper/keeper_test.go
```diff
@@ -38,6 +55,16 @@ func TestKeeperTestSuite(t *testing.T) {
 }
 
 func (suite *KeeperTestSuite) SetupTest() {
+	keys := storetypes.NewKVStoreKeys(
```
vladjdk, line 58: Is this a requirement for every module?

x/vm/store/snapshotmulti/store.go
```diff
@@ -0,0 +1,161 @@
...
+func (s *Store) GetStoreType() storetypes.StoreType {
+	return storetypes.StoreTypeMulti
+}
+
+// Implements CacheWrapper.
+func (s *Store) CacheWrap() storetypes.CacheWrap {
+	return s.CacheMultiStore().(storetypes.CacheWrap)
+}
+
+// CacheWrapWithTrace implements the CacheWrapper interface.
```
vladjdk, line 91: Let's add in the docs that this doesn't do anything

tests/integration/precompiles/distribution/test_distribution.go
cloudgray: fixed in https://github.com/cosmos/evm/pull/244/commits/264a8fe47b1bff3cc598e7af72dc04065e1a5fe2

tests/integration/precompiles/distribution/test_distribution.go
cloudgray: fixed in https://github.com/cosmos/evm/pull/244/commits/264a8fe47b1bff3cc598e7af72dc04065e1a5fe2

tests/integration/precompiles/staking/test_integration.go
cloudgray: That comment was left unchanged during the process of copying and modifying an existing test case. 😅
Fixed in https://github.com/cosmos/evm/pull/244/commits/a881f290b73e9ea39bc0e86b121dd2433ec27948

x/vm/store/snapshotmulti/store_bench_test.go
vladjdk: big brain

x/vm/store/snapshotmulti/store_test.go
```diff
@@ -0,0 +1,149 @@
...
+	kv.Set([]byte("a"), []byte("1"))
+
+	idx0 := snapshotStore.Snapshot()
+	snapshotStore.GetKVStore(key).Set([]byte("b"), []byte("2"))
+
+	idx1 := snapshotStore.Snapshot()
+	snapshotStore.GetKVStore(key).Set([]byte("c"), []byte("3"))
+
+	snapshotStore.RevertToSnapshot(idx1)
+	require.Nil(t, snapshotStore.GetKVStore(key).Get([]byte("c")))
```
vladjdk, line 49: Could we test this using the same key for the KVs (i.e. use "a" each time with a different value)? It would simulate someone overwriting the same address store for something like a delegation over multiple different snapshots.

x/vm/store/snapshotmulti/store_test.go
```diff
@@ -0,0 +1,149 @@
...
+	require.Equal(t, snapshotStore, wrap)
+
+	idx := snapshotStore.Snapshot()
+	require.Equal(t, 1, idx)
+}
+
+func TestSnapshotMultiCacheWrapWithTrace(t *testing.T) {
+	snapshotStore, _ := setupStore()
+
+	wrap := snapshotStore.CacheWrapWithTrace(nil, nil)
```
vladjdk, line 144: Document that this is just a regular cache wrap

x/vm/store/snapshotmulti/store_test.go
vladjdk: Maybe as an added test case - I do like this one as well.

x/vm/store/snapshotmulti/store.go
cloudgray: applied! https://github.com/cosmos/evm/pull/244/commits/ae1c9fc0e0204c9c30e3f55f1804e44ac1c9058b

x/vm/store/snapshotmulti/store_test.go
cloudgray: applied! https://github.com/cosmos/evm/pull/244/commits/33f939aba5cbbadead44ddf97184ab6409f4ef76

x/vm/store/snapshotmulti/store_test.go
cloudgray: applied! https://github.com/cosmos/evm/pull/244/commits/33f939aba5cbbadead44ddf97184ab6409f4ef76

PR #245, opened 2025-06-24
Problem: diff mempool max-txs trigger apphash mismatch

mmsqe:
# Description

`tx must have at least one signer` error trigger by different [mempool.max-txs](https://github.com/MANTRA-Chain/mantrachain-e2e/commit/abc8c27388c424ce508f688a970cf18ffe149974) 

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


ante/evm/04_validate.go
```diff
@@ -24,10 +24,6 @@ func ValidateMsg(
 	txData evmtypes.TxData,
 	from sdktypes.AccAddress,
 ) error {
-	if from != nil {
```
mmsqe, line 27: @Vvaradinov We need signer recovery and extractor to make PriorityMempool [work](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/d4bc921dce85f69e8aa060e6d844d0ba67a636eb/integration_tests/test_priority.py#L41)

ante/evm/05_signature_verification.go
```diff
@@ -83,16 +83,8 @@ func SignatureVerification(
...
-			errortypes.ErrorInvalidSigner,
-			"couldn't retrieve sender address from the ethereum transaction: %s",
-			err.Error(),
-		)
+	if err := msg.VerifySender(ethCfg.ChainID); err != nil {
+		return errorsmod.Wrapf(errortypes.ErrorInvalidSigner, "signature verification failed: %s", err.Error())
 	}
-
-	// set up the sender to the transaction field if not already
-	msg.From = sender.Hex()
```
vladjdk, line 96: Why was this ever allowed to be nil?

evmd/signer_test.go
```diff
@@ -0,0 +1,73 @@
...
+	ethMsg := types.NewTx(evmTx)
+	ethMsg.From = ethAddr.Bytes()
+	txWithEth := &mockHasExtOptions{
+		msg: ethMsg,
+	}
+	fallback := &mockFallback{}
+	adapter := evmd.NewEthSignerExtractionAdapter(fallback)
+	signers, err := adapter.GetSigners(txWithEth)
+	require.NoError(t, err)
+	require.Equal(t, []mempool.SignerData{{Signer: signers[0].Signer, Sequence: signers[0].Sequence}}, signers)
```
vladjdk: This doesn't seem like an effective comparison. We're basically saying require(signers==signers), which is always going to be true.

proto/cosmos/evm/vm/v1/tx.proto
```diff
@@ -46,10 +46,11 @@ message MsgEthereumTx {
   // hash of the transaction in hex format
   string hash = 3
       [ (gogoproto.moretags) = "rlp:\"-\"", (amino.dont_omitempty) = true ];
-  // from is the ethereum signer address in hex format. This address value is
-  // checked against the address derived from the signature (V, R, S) using the
+  string deprecated_from = 4 [ deprecated = true ];
```
vladjdk, line 49: For V1, we're going to be using new protos, with the assumption that the first version is a fresh start. Let's just keep using this field and remove the deprecated tag.

x/vm/types/msg.go
```diff
@@ -207,6 +224,39 @@ func (msg *MsgEthereumTx) GetMsgsV2() ([]protov2.Message, error) {
...
+}
+
+// GetSender convert the From field to common.Address
+// From should always be set, which is validated in ValidateBasic
+func (msg *MsgEthereumTx) GetSender() common.Address {
+	return common.BytesToAddress(msg.From)
+}
+
+// GetSenderLegacy fallbacks to old behavior if From is empty, should be used by json-rpc
+func (msg *MsgEthereumTx) GetSenderLegacy(chainID *big.Int) (common.Address, error) {
```
vladjdk: Could we rename this from GetSenderLegacy to GetSenderFromSignature? Legacy implies that this is an old method that shouldn't be used. As far as I understand, it injects the sender bytes into the msg.From and should always be used by the JSON-RPC before it hits the antehandler.

x/vm/types/msg.go
```diff
@@ -207,6 +224,39 @@ func (msg *MsgEthereumTx) GetMsgsV2() ([]protov2.Message, error) {
...
+func (msg *MsgEthereumTx) GetSigners() []sdk.AccAddress {
+	if len(msg.From) == 0 {
+		return nil
+	}
+	return []sdk.AccAddress{msg.GetFrom()}
+}
+
+// GetSender convert the From field to common.Address
+// From should always be set, which is validated in ValidateBasic
+func (msg *MsgEthereumTx) GetSender() common.Address {
```
vladjdk, line 238: Like the below comment, let's rename this to GetSenderFromMsg, which will make it clear that msg.From cannot be nil

tests/integration/x/vm/state_transition_benchmark.go
```diff
@@ -146,14 +150,12 @@ func newNativeMessage(
...
 ) (*core.Message, error) {
-	msgSigner := ethtypes.MakeSigner(cfg, big.NewInt(blockHeight), 10000000)
-
 	msg, baseFee, err := newEthMsgTx(nonce, address, krSigner, ethSigner, txType, data, accessList)
 	if err != nil {
 		return nil, err
 	}
 
-	m, err := msg.AsMessage(msgSigner, baseFee)
+	m, err := msg.AsMessage(baseFee)
```
vladjdk, line 155: Rename msg to signedMsg here

evmd/cmd/evmd/cmd/root.go
```diff
@@ -315,10 +316,19 @@ func newApp(
 
 	// Set up the required mempool and ABCI proposal handlers for Cosmos EVM
 	baseappOptions = append(baseappOptions, func(app *baseapp.BaseApp) {
-		mempool := sdkmempool.NoOpMempool{}
-		app.SetMempool(mempool)
-
-		handler := baseapp.NewDefaultProposalHandler(mempool, app)
+		var mpool mempool.Mempool
+		if maxTxs := cast.ToInt(appOpts.Get(server.FlagMempoolMaxTxs)); maxTxs >= 0 {
```
vladjdk, line 321: I'm testing this with a Foundry script that uses transaction batching on this branch, and it seems like this is still failing to process out-of-order nonces.

Getting these errors:
```
- server returned an error response: error code -32000: invalid nonce; got 24, expected 23: invalid sequence: invalid sequence
```

Afaik, this PR is supposed to somewhat mitigate that. Is that not the case, or am I doing something wrong here?

vladjdk (changes requested):
My understanding of this PR is that we're moving the VRS signer extraction from the antehandler to the JSON-RPC. I think that's a good idea overall. I have a few small nits regarding the changes in that domain.

However, when it came to using the priority mempool, I couldn't actually get it to work properly. I'm still getting the same errors that I would get using the default mempool.

I don't mind using the priority nonce as an interim to [this proposal](https://skip-protocol.notion.site/EXT-Cosmos-EVM-Appside-Mempool-21e8c17818cb80028f28cfa440d7a7e7?source=copy_link), because the mempool doesn't hold transactions, and only sorts them, so there's no risk of p2p layer spam.

However, since it seems to behave the same way as the `noop` mempool with batched transactions, I don't really see a difference in setting it over noop in this case.

To be clear, I agree with the rest of the PR.

ante/evm/05_signature_verification.go
mmsqe: `From` would be extracted from the signed msg inside transaction, so the [From](https://github.com/evmos/ethermint/pull/1479/files#diff-38bc4d76a75796a20c4e171ba855f67d7eea4a006fccb34423da3da429758a22R155) passed in is obsolete.

evmd/cmd/evmd/cmd/root.go
mmsqe: Do you mind share the script? but we use different [senders](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/d4bc921dce85f69e8aa060e6d844d0ba67a636eb/integration_tests/test_priority.py#L53) for priority test.

vladjdk (approved):

proto/cosmos/evm/vm/v1/tx.proto
vladjdk: I thought about this again, and the audited version of the code will still have the string version of `from`. Looks like we're going to still have to support this/deprecate it since this PR won't yet be in effect for the audited release. Apologies for my mistake here!

evmd/cmd/evmd/cmd/root.go
vladjdk: Anything that batch submits transactions using Forge would successfully replicate. Example: https://gist.github.com/vladjdk/cef7503d2dfe4c44eb3dbb1183947ec8

mmsqe: Got it, since need specify nonce from same sender to avoid out of order in priority mempool, I could try sender nonce mempool later.

yihuang: `PriorityNonceMempool` don't handle out of order nonce, it only handles priority.

vladjdk: @yihuang based on the [ADR](https://github.com/cosmos/cosmos-sdk/blob/main/types/mempool/priority_nonce_spec.md) it should order by nonce:
>The follow rules are strictly observed while iterating the mempool to select transactions:
> 1. For a given sender their txs must be selected in nonce order.
> 2. A transaction with a higher priority is always selected before a transaction with a lower priority except when to do so would violate sender-nonce order.

Based on a quick look of the code, it looks like it orders by sender and nonce as a secondary key:
https://github.com/cosmos/cosmos-sdk/blob/main/types/mempool/priority_nonce.go#L241-L276.

@mmsqe sounds good. Priority nonce, however, should function the same as a sender nonce with an added priority dimension.

yihuang: @vladjdk 

the problem is check-tx state, if the txs comes in the order of nonces, the check-tx logic can run on the same check-tx state. but if out of order, the check-tx logic will fail and they won't get into mempool in the first place.

aljo242:
@vladjdk to revist

vladjdk:
Will merge when you fix the lints, tests and fix the merge conflict.

skosito:
@mmsqe sorry for ping, i am curious is this issue that I opened in cosmos-sdk valid? https://github.com/cosmos/cosmos-sdk/issues/24848

it seemed once i checked that this extractor is not used in proposal handler, so my question is does priority nonce mempool work like this?

PR #255, opened 2025-07-01
Revert "(chore): move feemarket update BaseFee to EndBlock (#190)"

mmsqe:
This reverts commit b2b79472927ad254ece593f1abdb2e36314eb087.

# Description

for test [info](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/04a456e9f446bc311a76451f77df693d7913d65e/integration_tests/test_eip1559.py)

---

## Author Checklist

**All** items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.

I have...

- [ ] tackled an existing issue or discussed with a team member
- [ ] left instructions on how to review the changes
- [ ] targeted the `main` branch

## Reviewers Checklist

**All** items are required.
Please add a note if the item is not applicable
and please add your handle next to the items reviewed
if you only reviewed selected items.

I have...

- [ ] added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] confirmed all author checklist items have been addressed
- [ ] confirmed that this PR does not change production code
- [ ] reviewed content
- [ ] tested instructions (if applicable)
- [ ] confirmed all CI checks have passed

PR #246, opened 2025-06-25
fix: align eth_feeHistory with geth

mmsqe:
# Description

1.  To align [fee_history](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.fee_history) in  [test_basic](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/e398ccd6250450badbb12e16368e31f149a358ef/integration_tests/test_fee_history.py#L31) 
  - fixes broken `earliest`/`latest` block tag handling
  - aligns `blockCount` as `hexOrDecimal64` for EIP-1559 compatibility that breaks contract deploy in Foundry
  - ensures proper `baseFeePerGas` array length calculations
  - corrects `oldestBlock` return value (-1 discrepancy) with go-ethereum
2. correct handling of next block fee for historical block should not change based on latest block in [test_change](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/e398ccd6250450badbb12e16368e31f149a358ef/integration_tests/test_fee_history.py#L79)
3. rejects requests beyond current head block in [test_beyond_head](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/e398ccd6250450badbb12e16368e31f149a358ef/integration_tests/test_fee_history.py#L107)
4. error handling for invalid reward percentile in [test_percentiles](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/e398ccd6250450badbb12e16368e31f149a358ef/integration_tests/test_fee_history.py#L114)
5. ensures consistent results across concurrent block fetches in [test_concurrent](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/e398ccd6250450badbb12e16368e31f149a358ef/integration_tests/test_fee_history.py#L151)
6. ensures update parameter changes not breaking historical data consistency in [test_param_change](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/e398ccd6250450badbb12e16368e31f149a358ef/integration_tests/test_fee_history.py#L213)

Closes: #XXXX

---

## Author Checklist

**All** items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.

I have...

- [ ] tackled an existing issue or discussed with a team member
- [ ] left instructions on how to review the changes
- [ ] targeted the `main` branch

## Reviewers Checklist

**All** items are required.
Please add a note if the item is not applicable
and please add your handle next to the items reviewed
if you only reviewed selected items.

I have...

[...]

PR #190, opened 2025-05-29
(chore): move feemarket update BaseFee to EndBlock

vladjdk:
Moves the feemarket base fee updates to the end of the block for easier gas fee estimation.

Closes: #123

PR #256, opened 2025-07-01
fix(rpc): don't return negative rewards from eth_feeHistory

Pitasi:
# Description

Hi, I have a corner case where I'm doing something funny with gas meters during the execution of ethereum transactions, this leads to the `eth_feeHistory` RPC to return negative rewards in the first ~100 blocks after genesis.

Here is an example of what `eth_feeHistory` RPC returns when this happen:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "oldestBlock": "0x24",
    "reward": [
      [
        "-0x7caf0c"
      ],
      [
        "-0x6d192a"
      ],
      [
        "-0x5f7605"
      ],
      [
        "-0x538744"
      ],
      [
        "-0x49165c"
      ],
      [
        "-0x3ff390"
      ],
      [
        "-0x37f51e"
      ],
      [
        "-0x30f67a"
      ],
      [
        "-0x2ad7ab"
      ],
      [
        "-0x257cb6"
      ]
    ],
    "baseFeePerGas": [
      "0x7caf0c",
      "0x6d192a",
      "0x5f7605",
      "0x538744",
      "0x49165c",
      "0x3ff390",
      "0x37f51e",
      "0x30f67a",
      "0x2ad7ab",
      "0x257cb6",
      "0x257cb6"
    ],
    "gasUsedRatio": [
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974
    ]
  }
}
```

_Numbers_ like `-0x7caf0c` make most parser/eth clients cry 😅 This PR clips the reward to zero.

Honestly, I'm not 100% sure this is the right thing to do, but it fixes my issues with those clients (e.g. `forge`) that try to estimate EIP1552 fees using this RPC call (it's possible that they don't even use these rewards number, but at least they can now parse the response). 

As I said initially, this issue goes away after a few blocks, as rewards tend to return to 0 over time and baseFee decreases. Still, it's a bit annoying when developing and restarting the chain often.

### DIFF
--- a/rpc/backend/utils.go
+++ b/rpc/backend/utils.go
@@ -179,7 +179,7 @@ func (b *Backend) processBlock(
 			}
 			tx := ethMsg.AsTransaction()
 			reward := tx.EffectiveGasTipValue(blockBaseFee)
-			if reward == nil {
+			if reward == nil || reward.Sign() < 0 {
 				reward = big.NewInt(0)
 			}
 			sorter = append(sorter, txGasAndReward{gasUsed: txGasUsed, reward: reward})


cloudgray:
Following the FeeHistory Query logic, you'll find a line in the code where it queries the gasFeeCap of transactions included in blocks and assigns gasFeeCap - baseFee to the reward. **This is where negative values occur.**

[go-ethereum/core/types/transactions.go](https://github.com/cosmos/go-ethereum/blob/baa6eeda43ac9c40de082103017bb25b1ea1474c/core/types/transaction.go#L366)
```go
// EffectiveGasTip returns the effective miner gasTipCap for the given base fee.
// Note: if the effective gasTipCap is negative, this method returns both error
// the actual negative value, _and_ ErrGasFeeCapTooLow
func (tx *Transaction) EffectiveGasTip(baseFee *big.Int) (*big.Int, error) {
	if baseFee == nil {
		return tx.GasTipCap(), nil
	}
	var err error
	gasFeeCap := tx.GasFeeCap()
	if gasFeeCap.Cmp(baseFee) < 0 {
		err = ErrGasFeeCapTooLow
	}
	gasFeeCap = gasFeeCap.Sub(gasFeeCap, baseFee)     // <----This line

	gasTipCap := tx.GasTipCap()
	if gasTipCap.Cmp(gasFeeCap) < 0 {
		return gasTipCap, err
	}
	return gasFeeCap, err
}
```

**Under normal circumstances,` gasFeeCap > baseFee` should hold true**, 

However, currently, there's an issue where BaseFee is being set in the EndBlocker instead of the BeginBlocker, causing Block(N) to store the baseFee of Block(N+1). 

**In this situation where Block(N) uses baseFee(N+1), cases where `gasFeeCap < baseFee` become possible.**

We've almost finished identifying this problem, and changing the logic to be handled in the `BeginBlocker` should resolve the current FeeHistory issue.

Pitasi:
ah I see! In that case I'll wait for your EndBlocker->BeginBlocker fix, and will be more than happy to confirm if my problem is solved as well.

cloudgray:
> Hey @Pitasi, check if the recent PR (#190) fixed your issue!

@Pitasi 
I think @vladjdk meant PR #255 whice reverts PR #190

Pitasi:
Thanks guys, I tested with commit 679305dc8ea3, but I still have my problem plus a new one 😁 

1. eth_feeHistory no longer works correctly:

before, this correctly worked:

```
curl localhost:8545 -H content-type:application/json --data '{"method":"eth_feeHistory","params":["0xa", "latest", [20]],"id":1,"jsonrpc":"2.0"}'
```

after, the first param only works if it's a decimal number:
```
 curl localhost:8545 -H content-type:application/json --data '{"method":"eth_feeHistory","params":[10, "latest", [20]],"id":1,"jsonrpc":"2.0"}' | jq
```

otherwise you get this reponse:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32602,
    "message": "invalid argument 0: json: cannot unmarshal string into Go value of type uint64"
  }
}
```

2. i still have negative fee rewards:

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "oldestBlock": "0xa",
    "reward": [
      [
        "-0xfae36f8"
      ],
      [
        "-0xdb87019"
      ],
      [
        "-0xc016216"
      ],
      [
        "-0xa8135d3"
      ],
      [
        "-0x9310f18"
      ],
      [
        "-0x80aed35"
      ],
      [
        "-0x7098f8f"
      ],
      [
        "-0x6285d9d"
      ],
      [
        "-0x56351e9"
      ],
      [
        "-0x4b6e7ac"
      ]
    ],
    "baseFeePerGas": [
      "0xfae36f8",
      "0xdb87019",
      "0xc016216",
      "0xa8135d3",
      "0x9310f18",
      "0x80aed35",
      "0x7098f8f",
      "0x6285d9d",
      "0x56351e9",
      "0x4b6e7ac",
      "0x4b6e7ac"
    ],
    "gasUsedRatio": [
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974,
      0.000023283064370807974
    ]
  }
}
```

but this might be just me doing some "gas refund" logic inside a precompile call, do the block gas meter no longer matches the actual gas used by the transactions or something like this

Pitasi:
I just saw that 1. is already fixed in https://github.com/cosmos/evm/pull/246, but my original problem remain.

cloudgray:
@Pitasi 
https://github.com/cosmos/evm/commit/679305dc8ea3cdba1ec127bb872f23a07b6dd898 is before PR #255 . Could you test with latest commit? 

When I tested, the test cmd you mentioned works and negative value is not presented.

```bash
curl localhost:8545 -H content-type:application/json --data '{"method":"eth_feeHistory","params":["0xa", "latest", [20]],"id":1,"jsonrpc":"2.0"}'
```

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "oldestBlock": "0x2a",
    "reward": [
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ],
      [
        "0x0"
      ]
    ],
    "baseFeePerGas": [
      "0x381e0c",
      "0x311a4a",
      "0x2af701",
      "0x259821",
      "0x20e51d",
      "0x1cc879",
      "0x192f6a",
      "0x16097d",
      "0x13484d",
      "0x10df43",
      "0xec35b",
      "0xceaef",
      "0xb4d91",
      "0x9e3df",
      "0x8a763",
      "0x79277",
      "0x79277"
    ],
    "gasUsedRatio": [
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0,
      0
    ]
  }
}
```

Pitasi:
oof, I don't know where I got that commit hash from 😓 

re-tested with af6f8d436a410a76a14f414729abc8302ad9ff46 (`github.com/cosmos/evm v1.0.0-rc2.0.20250707012801-af6f8d436a41`), still having my issue

mmsqe:
> * During PrepareProposal I'm injecting an artificial MsgEthereumTx embedding a LegacyTx with gas price 0 at the top of every block ([code](https://github.com/warden-protocol/wardenprotocol/blob/main/warden/x/sched/keeper/abci.go#L133))

seems you bypass [ErrGasFeeCapTooLow](https://github.com/ethereum/go-ethereum/blob/v1.15.11/core/types/transaction.go#L363) when using 0 GasPrice for your precompile.

aljo242:
Closing this for now as we have resolved this in [this PR](https://github.com/cosmos/evm/pull/255).  Please feel free to open an issue if this was not resolved

Pitasi:
my "special" is use case has not been solved by #255, the comments

https://github.com/cosmos/evm/pull/256#issuecomment-3050203684

and 

https://github.com/cosmos/evm/pull/256#issuecomment-3051530798

contain more info on the use cases that are still affected

PR #263, opened 2025-07-07
test: separate test suites for dedicated component 

mmsqe:
# Description

* avoids using single KeeperTestSuite that share state
* ensures `_test.go` suffix to make test IDE friendly

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


tests/integration/x/feemarket/eip1559_test.go
zsystm, line 1: Hey @mmsqe, thanks for your interest in improving the test structure — really appreciate the effort.

I have a question regarding the changes in your PR.

One of the main reasons I originally renamed the test files from `*_test.go` to `test_*.go` was to make these test suites reusable by other applications implementing the `EvmApp` interface (ref: https://github.com/cosmos/evm/pull/198). With this structure, client chains can write tests like the following, by simply providing their own CreateApp implementation:
```go
func TestFeeMarketKeeperTestSuite(t *testing.T) {
	s := feemarket.NewTestKeeperTestSuite(CreateEvmd)
	suite.Run(t, s)
}
```

However, I noticed that in your PR, the `x/feemarket` tests were renamed back to `*_test.go`, which limits them to being runnable only within our own evmd application. Since Go does not expose `*_test.go` files as importable packages, other client chains won't be able to use them in their test suites.

Could you clarify your intent behind this change, or consider keeping the `test_*.go` structure so that the tests remain reusable across different implementations?

Thanks again for the contribution!

mmsqe: The main issue is unable to detect those test packages, now we can 
`go test -v  -run KeeperTestSuite/TestEndBlock ./tests/integration/x/feemarket` 
instead of 
`cd evmd && go test -tags=test -v -run TestFeeMarketKeeperTestSuite/TestEndBlock ./tests/integration`

We could reuse feemarket test suite since the setup is simple, but it's better to separate SetupTest for each test suite for module like vm. Child test class could inherit `SetupTest` from its base class which is called by test runtime automatically. I can split the change by module to make review easier.

zsystm: I see the benefit of being able to run tests directly via go test without needing to go through evmd, and agree that's a useful improvement for running tests within this repo.

That said, my original intention with the `test_*.go` structure was to make the test suites reusable outside of this repo — for example, by client chains like `Mantra` or `Babylon` that use cosmos/evm as a dependency. With the current `*_test.go` structure, those client chains won’t be able to import and use our test suites in their own repositories, because Go does not expose test files as importable packages.

To clarify: is the current structure you proposed still intended to support the following use case?

A client chain (e.g., MantraApp or BabylonApp) wants to write a minimal test like this in their repo:
```
func TestFeeMarketKeeperTestSuite(t *testing.T) {
	s := feemarket.NewTestKeeperTestSuite(CreateMantraApp)
	suite.Run(t, s)
}
```
...and have that test internally run the tests from `/evmd/tests/integration/x/feemarket`.

From what I can see in the current PR, this doesn't seem possible anymore — but please let me know if I’m missing something!

mmsqe: https://github.com/cosmos/evm/pull/263/commits/2efe7c1a38ca5790489353dd8ce3308a63389341 evmd still could use testutil from evm instead of feemarket packge right?

zsystm: @mmsqe 
I tried running the example you shared, but it results in:
```
=== RUN   TestFeeMarketKeeperTestSuite
    suite.go:234: warning: no tests to run
--- PASS: TestFeeMarketKeeperTestSuite (0.00s)
PASS
```
So it seems like there are no actual test cases being picked up.

To clarify — how would you make it possible for other applications (e.g. Mantra, Babylon) to run the tests defined in `evm/tests/integration/x/feemarket/abci_test.go`? (For other `*_test.go` also)

The goal is that client chains can run all the test cases defined in evm.

If the test file is named `*_test.go`, and lives in a separate test package, it likely won’t be included when imported externally — which defeats the purpose of making these test suites reusable.

Let me know if there's a workaround you had in mind.

mmsqe: strange, I run tried sth like `cd evmd && go test -v -count 1 -run KeeperTestSuite/TestEndBlock ./tests/integration`
```
=== RUN   TestIBCKeeperTestSuite
--- PASS: TestIBCKeeperTestSuite (0.00s)
=== RUN   TestERC20KeeperTestSuite
--- PASS: TestERC20KeeperTestSuite (0.00s)
=== RUN   TestFeeMarketKeeperTestSuite
    suite.go:234: warning: no tests to run
--- PASS: TestFeeMarketKeeperTestSuite (0.00s)
=== RUN   TestKeeperTestSuite
--- PASS: TestKeeperTestSuite (0.00s)
PASS
ok  	github.com/cosmos/evm/evmd/tests/integration	0.742s
```

mmsqe: I just revert the test separation, since I didn't know other chain is using the exported test, but we do e2e with binary in python to make life easier.

zsystm (approved):
LGTM 👍  Thanks for cleaning config.
We need an one more approval from @vladjdk

vladjdk (approved):

Issue #104, opened 2025-04-25
Migrate Compatibility Tests to Cosmos/EVM Repo


vladjdk:
Move [https://github.com/b-harvest/evm-tools-compatibility](https://github.com/b-harvest/evm-tools-compatibility) and hook it up to CI.

Issue #214, opened 2025-06-13
Scope App-Side Mempool


Eric-Warehime:
Figure out if an app-side mempool could even get parity with the ethereum mempool.

Does this block load testing at all? 

* It seems like not since we can generate an infinite number of wallets.

PR #287, opened 2025-07-15
tests: evm tools compatibility

zsystm:
# Description

Closes: #104 

### Foundry 
- [x] **Uniswap v3 deployment**
  - **Note**: Batch deployment isn’t supported yet, so we’re using the `--slow` flag for now. Full batch deployment will be possible once app-side mempool is implemented — tracked in https://github.com/cosmos/evm/issues/214.
- [x] **Cast-based tests**
  - [x] Network info queries (`eth_chainId`, `net_version`, etc.)
  - [x] Read ERC-20 state (`totalSupply`, `balanceOf`, etc.)
  - [x] Send ERC-20 transfer and validate balances
  - [x] Over-balance transfer and revert check
- [x] **Forge scripts and unit tests**
  - [x] ERC-20 & counter tests (`*.t.sol`)
  - [x] Scripts for querying, deploying, reading, transferring, and reverting

### Hardhat 

- [x] Deploy ERC-20 and Uniswap via Hardhat script
- [x] Call contract functions and log state (e.g. `totalSupply`, `balanceOf`)
- [x] Send transactions and confirm state changes
- [x] Handle revert cases 

### SDK

- [x] **viem**
  - Client config, ERC-20 read/write, edge case testing
- [x] **web3.js**
  - Provider setup, read/write, revert check and logging
- **ethers.js**
  - Tested via Hardhat
  - **Note**: Revert testing currently checks only for the presence of a revert. Once #289 is merged and proper revert reason strings are returned, these tests should be updated to validate the exact revert messages.
---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


tests/evm-tools-compatibility/web3.js/test/web3js_compatability.test.js
```diff
@@ -0,0 +1,91 @@
+// test/Web3jsCompatibility.test.js
+const Web3 = require("web3");
+const { expect } = require("chai");
+const tokenArtifact = require("./contractABI/TokenExample.json");
+
+describe("Web3.js Compatibility Test", function () {
```
cloudgray: There is file name typo
web3js_compat`a`bility.test.js -> web3js_compatib`i`lity.test.js

tests/evm-tools-compatibility/foundry/.env
```diff
@@ -0,0 +1,23 @@
...
+# Foundry nightly warning disable
+FOUNDRY_DISABLE_NIGHTLY_WARNING=true
+
+# chain id
+CHAIN_ID=4221
+# endpoint of custom chain
+CUSTOM_RPC=http://127.0.0.1:8545
+# private key for tx broadcast (PUBLIC TEST KEY - NOT A SECRET)
+# Split to avoid false positive secret detection
+PRIVATE_KEY_PART1=0x8a36c69d940a92fcea94b36d0f2928c7a0ee19a90
```
cloudgray, line 14: I think committing .env is not good idea altough the secrets in it is actually already written down in other scripts.

I think there are many alternatives.
For example, we can use .toml file instead, empty secrets, let developers set their secrets, or in case of github-ci job, we can use secrets set in github repo config.

If it is thought that it will take bunch of time, how about create issue instead and revisit?

cloudgray (commented):
LGTM with some trivial comments

tests/evm-tools-compatibility/foundry/.env
zsystm: @cloudgray 
I understand your concern, but in this case the `.env` file is only used for testing, and we have made it clear in the comments what it contains. The usual reason `.env` is considered a bad practice is when it contains sensitive production values, which is not the case here. 

Since replacing it would require extra effort without clear benefit, I’m not sure I agree with making this change for this particular scenario.

tests/evm-tools-compatibility/web3.js/test/web3js_compatability.test.js
zsystm: @cloudgray 
Applied in https://github.com/cosmos/evm/pull/287/commits/def7f8b7e11e3edda35846f045d0cde216d7bb0a

almk-dev (approved):
@zsystm Could you take a look at the `TestAllowedTxSize` failure?

PR #289, opened 2025-07-15
fix: align revert reason with go ethereum that expects hex-encoded result

mmsqe:
# Description

* [this](https://github.com/cosmos/evm/pull/224/files#diff-8a87111537b2d02eee090665c70c4b737046dc534d37f37fc1b1fb678cf89852R125) change shouldn't have been included since revert reason was [aligned](https://github.com/ethereum/go-ethereum/blob/v1.15.11/internal/ethapi/errors.go#L57)
* [message](https://github.com/ethereum/go-ethereum/blob/v1.15.11/rpc/json.go#L127) field already contains error detail, `data` get duplicated
* for test [info](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/627be88824e11eb7d2ab4c92c1f3848aa5e41b5a/integration_tests/test_estimate_gas.py#L22)

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


cloudgray:
> > Hey @mmsqe could you provide more context on this - why is this being reverted?
> 
> Seems like behavior change after #224, but we could always decode for precompile tests right?

Hi @mmsqe! 
Thank you very much for correcting the part I had mistakenly modified.

To add some context, I do believe that PR#224 was necessary. The part I incorrectly modified appears to be the handling of the return data in the JSON-RPC layer, where the return data became not hex-encoded.

The issue that [PR#224](https://github.com/cosmos/evm/pull/224) aimed to address was, more precisely, that **_when a precompile is invoked via a caller contract_**, the revert reason was not included in the return data, making it impossible to decode. [ref: #223 ]

The root cause was that the precompile, upon encountering an error, **was returning it in a way that differed from how Geth’s `opCall` handles errors.** 
- In Geth, when a revert occurs, opCall returns nil for the error and the hex-encoded revert reason in the return data. 
- On the other hand, prior to PR#244, the precompile would return a custom error type containing the revert reason in the error itself, and nil as the return data—the opposite behavior.

Although this diverged from the Geth interpreter’s behavior, it didn’t cause any issues when the precompile was invoked directly.

However, when the precompile is called from within a contract, **the revert reason does not get correctly passed along to opRevert, resulting in its loss.**

The key changes I made were as follows:

1. Changed the precompile’s return values on error from (nil, Error("reason")) to (ExecRevertError, abi-encoded("reason")).
2. Set the revert reason to `EVM.Interpreter.returnData` during errors, as is done by other opcodes, so that it can be accessed by operations like opReturnDataCopy.

From what I can see in your modifications, the issue doesn't seem to stem from these parts. That said, if I’ve misunderstood anything, I’d greatly appreciate your clarification!

cc. @zsystm @vladjdk @aljo242

cloudgray:
@mmsqe 

Upon reviewing [PR#224](https://github.com/cosmos/evm/pull/224) again at this point, it seems that setting the EVM interpreter's `returndata` to the response in `ApplyMessageWithConfig()` is unnecessary. This part is already handled automatically within the EVM itself. [[ref](https://github.com/cosmos/evm/blob/main/x/vm/keeper/state_transition.go#L447-L450)]

Would you be able to remove [that section](https://github.com/cosmos/evm/blob/main/x/vm/keeper/state_transition.go#L447-L450) in this PR?

cloudgray (approved):
LGTM! Really appreciate your contribution!

testutil/util.go
```diff
@@ -108,3 +116,22 @@ func CreateTx(ctx context.Context, txCfg client.TxConfig, priv cryptotypes.PrivK
 
 	return txBuilder.GetTx(), nil
 }
+
+func DecodeRevertReason(evmRes evmtypes.MsgEthereumTxResponse) error {
```
aljo242, line 137: Can you add a godoc that properly describes the behavior please?

aljo242 (commented):
LGTM with  a small nit

testutil/util.go
mmsqe: added

aljo242 (approved):

PR #291, opened 2025-07-15
fix: use address codecs in precompiles

technicallyty:
# Description

uses address codecs in precompiles

closes: EVM-77

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content

### DIFF
--- a/precompiles/gov/tx.go
+++ b/precompiles/gov/tx.go
@@ -33,7 +33,7 @@ func (p *Precompile) SubmitProposal(
 	method *abi.Method,
 	args []interface{},
 ) ([]byte, error) {
-	msg, proposerHexAddr, err := NewMsgSubmitProposal(args, p.codec)
+	msg, proposerHexAddr, err := NewMsgSubmitProposal(args, p.codec, p.addrCdc)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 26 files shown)


precompiles/gov/tx.go
```diff
@@ -33,7 +33,7 @@ func (p *Precompile) SubmitProposal(
 	method *abi.Method,
 	args []interface{},
 ) ([]byte, error) {
-	msg, proposerHexAddr, err := NewMsgSubmitProposal(args, p.codec)
+	msg, proposerHexAddr, err := NewMsgSubmitProposal(args, p.codec, p.addrCdc)
```
aljo242, line 36: Can we just make the `NewMsg*` functions members of the `Precompile` so they can just directly use the address codec from there?

technicallyty: i feel message constructors should be functions, not methods.

aljo242: okey that is fair.  do the methods need to be public or can we make them private?  They seem like internal precompile specific logic that we would not want other packages to use

technicallyty: hmm hard to say really - i'd wager it's not entirely out of the realm of possibility for a user to want to construct these messages for a different use case. maybe a custom module that does something with cosmos SDK messages. 🤷 

making them private does have the benefit of cleaning up the public API of the package, but it doesn't feel busy enough to worry about that to me

Issue #240, opened 2025-06-23
[Bug]: incorrect nonce when send multiple eth txs in single cosmos tx


mmsqe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

nonce get [overwritten](https://github.com/cosmos/evm/blob/0e511d32206b1ac709a0eb0ddb1aa21d29e833b8/x/vm/statedb/statedb.go#L684) to previous value when no more nonceDirty check 

### Cosmos EVM Version

main

### How to reproduce?

3 eth txs success, while sender [nonce](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/b014cceb311558d3b23a4e1a847bb7b27040ec79/integration_tests/test_basic.py#L408) only bump 1 instead of 3

mmsqe:
> [@mmsqe](https://github.com/mmsqe) I believe this is working as intended and matches the EVM spec behavior. `go-ethereum` only [supports one transaction](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L75-L106) and [this is enforced](https://github.com/cosmos/evm/blob/main/ante/evm/mono_decorator.go#L78-L79) in `MonoDecorator`, so it's not valid or supported for a single Cosmos transaction to contain multiple Ethereum transactions.
> 
> Let me know if we're on the same page and I'll close out this issue.

The comment [here](https://github.com/cosmos/evm/blob/main/ante/evm/mono_decorator.go#L78-L79) is inaccurate that multiple tx msgs can be processed without any enforcement.

PR #294, opened 2025-07-16
EVM-131: Add enforcement of a single EVM tx per Cosmos tx

almk-dev:
# Description

Closes: #240 

The original issue mentions that we are resetting the nonce to the original value for Cosmos txs containing multiple Ethereum txs, and that we should ensure that we only write the account state if the nonce has actually been modified (e.g. 1 Cosmos Tx with 3 Ethereum Txs should increment the nonce by 3). 

However, this is actually [expected behavior](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L75-L106) for the EVM spec and `go-ethereum`, as it always expects one Ethereum Tx per message. So, we should enforce this restriction instead and return an error if there are multiple EVM Txs. 

Please review if this addresses the intent of the original issue and interacts with `MonoDecorator`/`AnteHandler` correctly.

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [x] added a relevant changelog entry to the `Unreleased` section in `CHANGELOG.md`
- [ ] reviewed content


aljo242:
Lgtm but lets add a test similar to how some are done [here](https://github.com/cosmos/evm/blob/main/ante/evm/fee_checker_test.go)

yihuang:
> However, this is actually [expected behavior](https://github.com/ethereum/go-ethereum/blob/master/core/types/transaction.go#L75-L106) for the EVM spec and go-ethereum, as it always expects one Ethereum Tx per message. So, we should enforce this restriction instead and return an error if there are multiple EVM Txs.

Just to clarify, batch tx is a feature that is supported in the original ethermint, and the json-rpc layer is implemented with that in mind.
But it's understandable to remove this feature to reduce attack vector, and there are some cleanups we can do if we don't need to support this.

aljo242:
@yihuang what cleanups do you have in mind?

aljo242:
@yihuang we will go ahead and merge this - please feel free to open an issue for the cleanups you're mentioning :)

aljo242 (approved):

Issue #334, opened 2025-07-23
[Bug]: debug_traceBlockByNumber returns failed transactions


mmsqe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

trace result contains fail transaction

### Cosmos EVM Version

main

### How to reproduce?

submit [3](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/a2a293cf440d9bbb0186d1d39f5f838754debae9/integration_tests/test_debug_traceblock.py#L26) txs, 2 success and 1 fail with insufficient funds

PR #335, opened 2025-07-23
fix: align filter rule for debug trace block

mmsqe:
# Description

upstream fix from https://github.com/evmos/ethermint/pull/1688, for test [info](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/a2a293cf440d9bbb0186d1d39f5f838754debae9/integration_tests/test_debug_traceblock.py#L59)

Closes: https://github.com/cosmos/evm/issues/334

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


vladjdk:
@cloudgray after an initial look, it looks like you're right - both successful and failed transactions are included in blocks.

However, this seems like it's a bit more nuanced. If a transaction fails for insufficient funds, it's never included in an EVM block, but it *would* be included in a Cosmos block, so we would need to filter that out. The `TxSucessOrExpectedFailure` function handles this by keeping transactions that run out of gas (revert on the EVM), which is the expected output, as they would be included in the EVM.

vladjdk (approved):
LGTM. `TxSucessOrExpectedFailure` covers the EVM-included errors in the "Expected Failure" case.

Issue #336, opened 2025-07-23
[Feature]: support state overrides in eth_call


mmsqe:
### Summary

for useful example, see https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#ethcall-override-example

### Proposed Feature

add the feature

### Work Breakdown

```[tasklist]
#### Must have
- [ ] add in eth_call 
```

```[tasklist]
#### Nice to have
- [ ] could add block overrides for debug api later
```

PR #337, opened 2025-07-23
feat: support state overrides in eth_call

mmsqe:
# Description

upstream feature from https://github.com/crypto-org-chain/ethermint/pull/369, for test [info](https://github.com/MANTRA-Chain/mantrachain-e2e/pull/49/commits/cc913f2fade1ea7f85b66399d3c4be81805d8bce)

Closes: https://github.com/cosmos/evm/issues/336

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


rpc/namespaces/ethereum/eth/api.go
```diff
@@ -269,15 +270,15 @@ func (e *PublicAPI) GetProof(address common.Address,
 // Call performs a raw contract call.
 func (e *PublicAPI) Call(args evmtypes.TransactionArgs,
 	blockNrOrHash rpctypes.BlockNumberOrHash,
-	_ *rpctypes.StateOverride,
+	overrides *json.RawMessage,
```
aljo242, line 268: nit: formatting

rpc/backend/call_tx.go
```diff
@@ -327,7 +327,7 @@ func (b *Backend) EstimateGas(args evmtypes.TransactionArgs, blockNrOptional *rp
 // DoCall performs a simulated call operation through the evmtypes. It returns the
 // estimated gas used on the operation or an error if fails.
 func (b *Backend) DoCall(
-	args evmtypes.TransactionArgs, blockNr rpctypes.BlockNumber,
+	args evmtypes.TransactionArgs, blockNr rpctypes.BlockNumber, overrides *json.RawMessage,
 ) (*evmtypes.MsgEthereumTxResponse, error) {
```
aljo242, line 364: nit: formatting of the arguments - can we make them vertical?

x/vm/keeper/grpc_query.go
```diff
@@ -223,6 +224,13 @@ func (k Keeper) EthCall(c context.Context, req *types.EthCallRequest) (*types.Ms
 		return nil, status.Error(codes.InvalidArgument, "empty request")
 	}
 
+	var overrides rpctypes.StateOverride
+	if len(req.Overrides) > 0 {
+		if err := json.Unmarshal(req.Overrides, &overrides); err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
```
aljo242: I think we should add some context to this error that it is specifically about the state overrides not being able to be unmarshalled

mmsqe: done

tests/integration/rpc/backend/test_call_tx.go
```diff
@@ -474,7 +474,7 @@ func (s *TestSuite) TestDoCall() {
 			s.SetupTest() // reset test and queries
 			tc.registerMock()
 
-			msgEthTx, err := s.backend.DoCall(tc.callArgs, tc.blockNum)
```
cloudgray, line 472: It would be good to add a simple test case to check whether the expected result is returned when the state overrides input is set.

tests/integration/x/vm/test_state_transition.go
```diff
@@ -788,6 +788,7 @@ func (s *KeeperTestSuite) TestApplyMessageWithConfig() {
 				true,
 				config,
 				txConfig,
+				nil,
```
cloudgray: Similar to the case in [test_call_tx.go](https://github.com/cosmos/evm/pull/337/files#diff-d89abf2cf9b2fb4377706e9cecddb8bd4582e9b3b3ef8341a0552b83ed06029d), it would be good to add a test case for when the state overrides input is set.

rpc/types/types.go
```diff
@@ -55,6 +60,42 @@ type RPCTransaction struct {
 // StateOverride is the collection of overridden accounts.
 type StateOverride map[common.Address]OverrideAccount
 
+// Apply overrides the fields of specified accounts into the given state.
+func (diff *StateOverride) Apply(db *statedb.StateDB) error {
```
cloudgray: As of go-ethereum v1.15.10, the [StateOverride.Apply](https://github.com/cosmos/go-ethereum/blob/baa6eeda43ac9c40de082103017bb25b1ea1474c/internal/ethapi/override/override.go#L64-L87) method includes logic for handling precompile addresses. 

From a compatibility standpoint, it may be best to either align the logic to match exactly, or alternatively, directly import the override package from Geth and use its `Apply` method. In the latter case, the `StateDB.Finalise` method would likely need to be implemented as a noop function.

x/vm/statedb/statedb.go
```diff
@@ -477,6 +477,22 @@ func (s *StateDB) SetState(addr common.Address, key, value common.Hash) common.H
 	return common.Hash{}
 }
 
+// SetBalance sets the balance of account associated with addr to amount.
+func (s *StateDB) SetBalance(addr common.Address, amount *uint256.Int) {
```
cloudgray: I think this interface is aligned to go-ethereum v1.11.x.

Currently, cosmos/evm is compatibiile to go-ethereum v1.15.x.

You can check added arg, `reason tracing.BalanceChangeReason`. ([ref](https://github.com/cosmos/go-ethereum/blob/baa6eeda43ac9c40de082103017bb25b1ea1474c/core/state/statedb.go#L428-L433))

x/vm/statedb/state_object.go
```diff
@@ -263,6 +273,12 @@ func (s *stateObject) SetState(key common.Hash, value common.Hash) common.Hash {
 	return prev
 }
 
+func (s *stateObject) SetStorage(storage Storage) {
```
cloudgray, line 267: I think Comment is needed because it can be confusing considering stateObject.SetStorage does not exist in go-ethereum (v1.15.10)

I suspect the reason this method was added is, it is difficult to use stateObject.StateState as it is. Because SetState sets value to dirtyStorage, and GetCommitedState method cannot get overrided state. So it seems that you alse added stateObject.overrideStorage and let `GetCommittedState` get state from overrideStorage.

I'm not sure my suspection is correct but anyway, it will helpful for devs to manange code if background explanation of this method is added.

cloudgray (changes requested):
Thanks for great work!
I left some comments for test and go-ethereum v1.15.x compatibility.

rpc/namespaces/ethereum/eth/api.go
```diff
@@ -266,13 +267,14 @@ func (e *PublicAPI) GetProof(address common.Address,
...
+func (e *PublicAPI) Call(
+	args evmtypes.TransactionArgs,
 	blockNrOrHash rpctypes.BlockNumberOrHash,
-	override *rpctypes.StateOverride,
+	overrides *json.RawMessage,
 ) (hexutil.Bytes, error) {
 	e.logger.Debug("eth_call", "args", args.String(), "block number or hash", blockNrOrHash)
 
-	if override != nil {
+	if overrides != nil {
```
dudong2:
```
> cast call 0x6E779A52fe86E5c739121C20Be090d94f20C7f5d "number()" \
  --override-state 0x6E779A52fe86E5c739121C20Be090d94f20C7f5d:0:0x0000000000000000000000000000000000000000000000000000000000000003e7 \
  --rpc-url http://127.0.0.1:8545
  
Error: server returned an error response: error code -32000: overrides are unsupported in call queries
```

To support state overrides, you must remove this nil check.

rpc/types/types.go
```diff
@@ -62,18 +67,91 @@ type RPCTransaction struct {
 // StateOverride is the collection of overridden accounts.
 type StateOverride map[common.Address]OverrideAccount
 
+func (diff *StateOverride) has(address common.Address) bool {
+	_, ok := (*diff)[address]
+	return ok
+}
+
+// Apply overrides the fields of specified accounts into the given state.
+func (diff *StateOverride) Apply(db *statedb.StateDB, precompiles vm.PrecompiledContracts) error {
```
aljo242, line 76: we should unit test this new function

rpc/backend/tx_info.go
```diff
@@ -381,8 +382,12 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
 
 // CreateAccessList returns the list of addresses and storage keys used by the transaction (except for the
 // sender account and precompiles), plus the estimated gas if the access list were added to the transaction.
-func (b *Backend) CreateAccessList(args evmtypes.TransactionArgs, blockNrOrHash rpctypes.BlockNumberOrHash) (*rpctypes.AccessListResult, error) {
-	accessList, gasUsed, vmErr, err := b.createAccessList(args, blockNrOrHash)
+func (b *Backend) CreateAccessList(
+	args evmtypes.TransactionArgs,
+	blockNrOrHash rpctypes.BlockNumberOrHash,
+	overrides *json.RawMessage,
```
aljo242, line 402: Now that we are just allowing a generic json message, how / where are we asserting that it has the proper payload structure?

tests/integration/x/vm/test_state_transition.go
```diff
@@ -852,6 +867,65 @@ func (s *KeeperTestSuite) TestApplyMessageWithConfig() {
...
+		},
+		{
+			"success - message applied with balance override",
+			func() core.Message {
+				sender := s.Keyring.GetKey(0)
+				recipient := s.Keyring.GetAddr(1)
+				msg, err := s.Factory.GenerateGethCoreMsg(sender.Priv, types.EvmTxArgs{
+					To:       &recipient,
+					Amount:   big.NewInt(100),
+					GasLimit: params.TxGas,
```
aljo242: Can we change these test cases to just Key-Values for better readability?

x/vm/keeper/state_transition.go
```diff
@@ -27,20 +28,18 @@ import (
...
-//
-// NOTE: the RANDOM opcode is currently not supported since it requires
-// RANDAO implementation. See https://github.com/evmos/ethermint/pull/1520#pullrequestreview-1200504697
-// for more information.
-func (k *Keeper) NewEVM(
+// NewEVMWithOverridePrecompiles creates a new EVM instance with opcode hooks and optionally overrides
+// the precompiles call hook. If overridePrecompiles is true, the EVM will use the keeper's static precompiles
+// for call hooks; otherwise, it will use the recipient-specific precompile hook.
+// This is useful for scenarios such as eth_call, state overrides, or testing where custom precompile logic is needed.
+// The function sets up the block context, transaction context, and VM configuration before returning the
```
aljo242: Looks like this comment is incomplete

x/vm/keeper/state_transition.go
```diff
@@ -350,14 +382,32 @@ func (k *Keeper) ApplyMessage(ctx sdk.Context, msg core.Message, tracer *tracing
...
-func (k *Keeper) ApplyMessageWithConfig(ctx sdk.Context, msg core.Message, tracer *tracing.Hooks, commit bool, cfg *statedb.EVMConfig, txConfig statedb.TxConfig, internal bool) (*types.MsgEthereumTxResponse, error) {
+func (k *Keeper) ApplyMessageWithConfig(
+	ctx sdk.Context,
+	msg core.Message,
+	tracer *tracing.Hooks,
+	commit bool,
+	cfg *statedb.EVMConfig,
+	txConfig statedb.TxConfig,
+	internal bool,
+	overrides *rpctypes.StateOverride,
```
aljo242, line 393: @vladjdk this functions is starting to have a complicated signature. do you think we could move more of these options into configs etc?

rpc/backend/tx_info.go
mmsqe: similar to [previous](https://github.com/cosmos/evm/pull/343#discussion_r2244429741) case

rpc/namespaces/ethereum/eth/api.go
cloudgray: @dudong2 (cc. @mmsqe) I think nil check was removed. I'll resolve this conversation.

Issue #342, opened 2025-07-23
[Bug]: tracer config is not aligned with go-ethereum


mmsqe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

for more info about [callTracerConfig](https://github.com/ethereum/go-ethereum/blob/v1.15.11/eth/tracers/native/call.go#L123) 

### Cosmos EVM Version

main

### How to reproduce?

trace with [onlyTopCall](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/eff3a530b7e711f0fca7c13fc515138a1135586f/integration_tests/test_tracers.py#L128) will get `invalid argument 1: json: cannot unmarshal object into Go struct field TraceConfig.tracerConfig of type string`

PR #343, opened 2025-07-23
fix: align tracer config with go-ethereum

mmsqe:
# Description

upstream fix from https://github.com/crypto-org-chain/ethermint/pull/382, for test [info](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/3de1b7a67e99b56fbe9eb2aabd56eb2daf24138b/integration_tests/test_tracers.py#L121)

Closes: https://github.com/cosmos/evm/issues/342

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content


rpc/backend/tracing.go
```diff
@@ -109,7 +109,7 @@ func (b *Backend) TraceTransaction(hash common.Hash, config *evmtypes.TraceConfi
 	}
 
 	if config != nil {
-		traceTxRequest.TraceConfig = config
+		traceTxRequest.TraceConfig = b.convertConfig(config)
 	}
```
aljo242, line 114: looks like we don't need this nil check since `convertConfig` handles it internally

mmsqe: since `TraceConfig` default is nil, set to `evmtypes.TraceConfig{}` to change origin behaviour

aljo242: I'm not sure I follow this.  I just mean that `convertConfig` already has the `nil` check so we're checking if `config != nil`  multiple times right?

mmsqe: If we remove the nil check, `convertConfig` will set the TraceConfig to an empty `evmtypes.TraceConfig`, which is not nil. We need to keep `traceTxRequest.TraceConfig` as nil when config is nil.

rpc/types/types.go
```diff
@@ -87,3 +90,8 @@ type OneFeeHistory struct {
 	Reward               []*big.Int // each element of the array will have the tip provided to miners for the percentile given
 	GasUsedRatio         float64    // the ratio of gas used to the gas limit for each block
 }
+
```
cloudgray, line 93: It would be great if comment is added so that explains why `TraceConfig` embedding  `evmtypes.TraceConfig` exists.

rpc/backend/tracing.go
```diff
@@ -134,11 +134,20 @@ func (b *Backend) TraceTransaction(hash common.Hash, config *evmtypes.TraceConfi
 	return decodedResult, nil
 }
 
+func (b *Backend) convertConfig(config *rpctypes.TraceConfig) *evmtypes.TraceConfig {
```
cloudgray, line 138: Isn't JSON validation needed?

e.g.
```go
func (b *Backend) convertConfig(config *rpctypes.TraceConfig) (*evmtypes.TraceConfig, error) {
	if config == nil {
		return &evmtypes.TraceConfig{}, nil
	}

	var tracerConfig string
	if len(config.TracerConfig) > 0 {
		// Validate JSON format
		if !json.Valid(config.TracerConfig) {
			return nil, fmt.Errorf("invalid JSON in tracerConfig: %s",
				string(config.TracerConfig))
		}
		tracerConfig = string(config.TracerConfig)
	}

	cfg := config.TraceConfig
	cfg.TracerJsonConfig = tracerConfig
	return &cfg, nil
}
```

mmsqe: I think `config.TracerConfig` being a `json.RawMessage` should be in valid JSON format in the first place?

cloudgray: If you try test the following test case for `convertConfig`, both test case passes without panic or error.

```go
{
	"valid JSON passes through",
	&rpctypes.TraceConfig{
		TracerConfig: []byte(`{"timeout": "5s"}`),
	},
	false,
	"Valid JSON should be converted successfully",
},
{
	"invalid JSON passes through without validation",
	&rpctypes.TraceConfig{
		TracerConfig: []byte(`{invalid json here`),
	},
	false,
	"Invalid JSON passes through convertConfig without validation - this demonstrates the vulnerability",
},
```

In fact, if invalid JSON bytes are provided as input, a parsing error will eventually be returned somewhere down the line.
But I thought it would be better if we could block it earlier in the process.

yihuang: in unit test it's possible to be invalid json, but in practice, they are decoded from json from json-rpc requests in the first place, so invalid json are checked there already.

cloudgray: @yihuang @mmsqe 
Ok, I understood. I'll resolve this conversation!

PR #346, opened 2025-07-23
feature: Add eth_createAccessList method and implementation

almk-dev:
# Description
Closes: https://github.com/sherlock-audit/2025-05-interchain-labs-evm-update-may-19th/issues/583
Addresses: [STACK-1195](https://linear.app/cosmoslabs/issue/STACK-1195/audit-583-missing-eip-2930-access-list-creation-rpc-method)

Review for trace response and validate correct gas estimation.

---

## Author Checklist
I have...

## Reviewers Checklist
I have...

- [ ] reviewed content


aljo242:
@aljo242 to review

local_node.sh
```diff
@@ -191,11 +191,6 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
 	sed -i.bak 's/"voting_period": "172800s"/"voting_period": "30s"/g' "$GENESIS"
 	sed -i.bak 's/"expedited_voting_period": "86400s"/"expedited_voting_period": "15s"/g' "$GENESIS"
 
-	# set custom pruning settings
-	sed -i.bak 's/pruning = "default"/pruning = "custom"/g' "$APP_TOML"
```
aljo242: what are the implications of this change?

rpc/backend/tx_info.go
```diff
@@ -465,3 +467,155 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *tmrpctypes.ResultBlock, i
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Determine if this is a prospective transaction or an executed transaction
+	// by checking if the transaction hash exists in the block and finding its exact position
+	var isExecuted bool
+	var targetTxIndex int
+	var targetMsgIndex int
+	var predecessors []*evmtypes.MsgEthereumTx
```
aljo242: let's group these

rpc/backend/tx_info.go
```diff
@@ -465,3 +467,155 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *tmrpctypes.ResultBlock, i
 		b.EvmChainID,
 	)
 }
+
+// CreateAccessList returns the list of addresses and storage keys used by the transaction (except for the
+// sender account and precompiles), plus the estimated gas if the access list were added to the transaction.
+func (b *Backend) CreateAccessList(args evmtypes.TransactionArgs, blockNrOrHash rpctypes.BlockNumberOrHash) (*rpctypes.AccessListResult, error) {
```
aljo242, line 384: should we log info / errors as things come along

local_node.sh
almkx: We are no longer using custom pruning settings that keep pruning after 2 blocks. Now it keeps the default configuration which is much larger (I think 64k) and more usable. Not entirely sure why it was set up this way to begin with tbh.

almkx: Using the default config mainly avoids this error that happens when testing requests against a local node:
```
failed to load state at height 1; version mismatch on immutable IAVL tree; version does not exist.
Version has either been pruned, or is for a future block height
```

rpc/backend/tx_info.go
```diff
@@ -465,3 +467,155 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *tmrpctypes.ResultBlock, i
...
+				isExecuted = true
+				targetTxIndex = txIdx
+				targetMsgIndex = msgIdx
+				break
+			}
+		}
+		if isExecuted {
+			break
+		}
+	}
```
aljo242: could we make this an internal function where we `return` instead of `break`

rpc/backend/tx_info.go
aljo242, line 384: feels like this function should be split into logical groupings of the different steps we take

x/vm/keeper/grpc_query.go
```diff
@@ -641,6 +647,44 @@ func (k *Keeper) traceTx(
 		overrides = traceConfig.Overrides.EthereumConfig(types.GetEthChainConfig().ChainID)
 	}
 
+	// Handle access list tracer specifically
+	if traceConfig.Tracer == types.TracerAccessList {
```
aljo242: this should be its own function since it is some separate specific logic with no side effects

almkx: Removed (https://github.com/cosmos/evm/pull/346/commits/3d9e60abf9cd81ffc5b41cee61cec5f0d81a8299) in favor of iterative crawling through calls.

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/d9f1568ce406e4302903d8f8fcd2b23dd09c8c71, https://github.com/cosmos/evm/pull/346/commits/ac62325a64f7e1a874ea6d59feeb10dcda71f6da

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/d00079dbe994b8154582454ff9f40eb63f82787e

rpc/backend/tx_info.go
almkx: removed in favor of iterative DoCall steps with tracer

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
 		b.EvmChainID,
 	)
 }
+
+// CreateAccessList returns the list of addresses and storage keys used by the transaction (except for the
+// sender account and precompiles), plus the estimated gas if the access list were added to the transaction.
+func (b *Backend) CreateAccessList(args evmtypes.TransactionArgs, blockNrOrHash rpctypes.BlockNumberOrHash) (*rpctypes.AccessListResult, error) {
+	acl, gasUsed, vmerr, err := b.createAccessList(args, blockNrOrHash)
```
aljo242:
```suggestion
	accessList, gasUsed, vmErr, err := b.createAccessList(args, blockNrOrHash)
```

aljo242: we prefer more specific variable names and camel case

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+	if err != nil {
+		return nil, err
+	}
+
+	hexGasUsed := hexutil.Uint64(gasUsed)
+	result := rpctypes.AccessListResult{
+		AccessList: &acl,
```
aljo242: are we taking a pointer to a pointer here?

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+	}
+
+	prevTracer, traceArgs, err := b.initAccessListTracer(args, blockNum, addressesToExclude)
+	if err != nil {
+		b.Logger.Error("failed to init access list tracer", "error", err)
+		return nil, 0, nil, err
+	}
+
+	// iteratively expand the access list
+	for {
```
aljo242, line 432: should we have some max number of iterations or constraints on this?

Do you see risks of this being a DOS vector?

aljo242, line 432:
```suggestion
	for !newTracer.Equal(prevTracer) {
```

could we do something like this to make this more readable?

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+	if vmerr != nil {
+		result.Error = vmerr.Error()
+	}
+	return &result, nil
+}
+
+// If the accesslist creation fails an error is returned.
+// If the transaction itself fails, an vmErr is returned.
+// If the transaction succeeds, the access list is returned.
+func (b *Backend) createAccessList(args evmtypes.TransactionArgs, blockNrOrHash rpctypes.BlockNumberOrHash) (ethtypes.AccessList, uint64, error, error) {
```
aljo242, line 406: can we make this in the format of a godoc with the comment starting with `// createAccessList`

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+				b.Logger.Error("vm error after access list converged", "vmError", res.VmError)
+				vmErr = errors.New(res.VmError)
+			}
+			return accessList, res.GasUsed, vmErr, nil
+		}
+		prevTracer = newTracer
+	}
+}
+
+func (b *Backend) getAccessListExcludes(args evmtypes.TransactionArgs, blockNum rpctypes.BlockNumber) (map[common.Address]struct{}, error) {
```
aljo242, line 459: can we godoc this?

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+	}
+
+	// exclude sender and precompiles
+	addressesToExclude := map[common.Address]struct{}{args.GetFrom(): {}}
+	precompiles := vm.ActivePrecompiles(b.ChainConfig().Rules(header.Number, true, header.Time)) // isMerge is true
+	for _, addr := range precompiles {
+		addressesToExclude[addr] = struct{}{}
+	}
+
+	// check authorization list
```
aljo242:
```suggestion
	// check if enough gas was provided to cover all authorization lists
```

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+	maxAuthorizations := uint64(*args.Gas) / params.CallNewAccountGas
+	if uint64(len(args.AuthorizationList)) > maxAuthorizations {
+		b.Logger.Error("insufficient gas to process all authorizations", "maxAuthorizations", maxAuthorizations)
+		return nil, errors.New("insufficient gas to process all authorizations")
+	}
+
+	for _, auth := range args.AuthorizationList {
+		// validate authorization (duplicating stateTransition.validateAuthorization() logic from geth)
```
aljo242: can you permalink to the logic youre referring to?

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+		if (!auth.ChainID.IsZero() && auth.ChainID.CmpBig(b.ChainConfig().ChainID) != 0) || auth.Nonce+1 < auth.Nonce {
```
aljo242:
```
auth.Nonce+1 < auth.Nonce
```

what is this checking? overflow?

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+		if authority, err := auth.Authority(); err == nil {
+			addressesToExclude[authority] = struct{}{}
+		}
+	}
+
+	b.Logger.Debug("access list excludes created", "addressesToExclude", addressesToExclude)
+	return addressesToExclude, nil
+}
+
+func (b *Backend) initAccessListTracer(args evmtypes.TransactionArgs, blockNum rpctypes.BlockNumber, addressesToExclude map[common.Address]struct{}) (*logger.AccessListTracer, *evmtypes.TransactionArgs, error) {
```
aljo242, line 506: godoc

rpc/backend/tx_info.go
```diff
@@ -375,3 +378,142 @@ func (b *Backend) GetTransactionByBlockAndIndex(block *cmtrpctypes.ResultBlock,
...
+	header, err := b.HeaderByNumber(blockNum)
```
aljo242, line 507: we have to make this call all of the time throughout our RPC layer - wondering if we should start caching these.  (would be a future issue)

tests/jsonrpc/scripts/run-compat-test.sh
```diff
@@ -43,7 +43,7 @@ echo "🔧 evmd will auto-initialize when container starts..."
 
 # Run the compatibility tests
 echo "🚀 Running JSON-RPC compatibility tests..."
-cd "$JSONRPC_DIR" && docker compose up --build --abort-on-container-exit
+cd "$JSONRPC_DIR" && docker compose up --build --force-recreate --abort-on-container-exit
```
aljo242: what does this change do?

x/vm/types/tracer.go
```diff
@@ -28,9 +28,8 @@ func NewTracer(tracer string, msg core.Message, cfg *params.ChainConfig, height
 
 	switch tracer {
 	case TracerAccessList:
-		blockAddrs := map[common.Address]struct{}{
-			*msg.To: {}, msg.From: {},
-		}
+		blockAddrs := make(map[common.Address]struct{})
+		blockAddrs[msg.From] = struct{}{}
```
aljo242, line 32: why did we make this change.  seems like we are no longer setting the `To` address as a blocked Address?

aljo242 (commented):
overall LGTM! mainly nits and questions

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/3b867caf93e82d245a84eeb4a0b573e8020982c3

rpc/backend/tx_info.go
almkx: Yeah, that way the return type keeps parity with `geth`. It also changes how it in serialized for the JSON rpc endpoint I think, so that you can get a real `null` value for the field.

tests/jsonrpc/scripts/run-compat-test.sh
almkx: Nothing, I was trying to fix the `run-compat-test` image not rebuilding; it was committed by accident.

almkx: https://github.com/cosmos/evm/pull/346/commits/e6daabc34e03b3d46f2c95a8dbf6a71670b62fb1

x/vm/types/tracer.go
almkx: Hm, there is conflicting documentation on whether both `from` and `to` should be in the exclusion list, or just `from`. For example, [AWS managed blockchain](https://docs.aws.amazon.com/managed-blockchain/latest/ethereum-dev/supported-json-rpc.html#:~:text=except%20for%20the%20sender%20account%20and%20the%20precompiles), [Kaia docs](https://docs.kaia.io/references/json-rpc/klay/create-access-list/#:~:text=except%20for%20the%20sender%20account%20and%20the%20precompiles), and even Ethereum's official `web3py` [docs](https://web3py.readthedocs.io/en/stable/web3.eth.html#:~:text=except%20for%20the%20sender%20account%20and%20the%20precompiles): 
> This method creates an [EIP2930](https://eips.ethereum.org/EIPS/eip-2930) type accessList based on a given Transaction. The accessList contains all the storage slots and addresses read and written by the transaction, except for the sender account and the precompiles. This method uses the same transaction call object and blockNumberOrTag object as eth_call.

However, the latest `geth` [seems to exclude both](https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L1191), so I'll mirror that.

almkx: https://github.com/cosmos/evm/pull/346/commits/5f1f195e5905928009aa40136bde7341ac633551

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/3c5b6310d43d9086f47dd1c70443a6494f756900

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/3c5b6310d43d9086f47dd1c70443a6494f756900

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/3c5b6310d43d9086f47dd1c70443a6494f756900

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/5f1f195e5905928009aa40136bde7341ac633551

rpc/backend/tx_info.go
almkx: https://github.com/cosmos/evm/pull/346/commits/3cd3ecbca6188b9d82f4adaa741b0f7ebf64121e

rpc/backend/tx_info.go
almkx: Yep. This should make it less opaque: https://github.com/cosmos/evm/pull/346/commits/7d08f5ad43b1a0d3e61832a9bc6d52389182e847

aljo242 (approved):

rpc/backend/tx_info.go
almkx: Hm, not to have it run infinitely, but I suppose they could send arbitrarily complex contracts to make the execution time longer and longer. However that is a characteristic of many of our endpoints.

almkx: I think it would make it less readable since we would have to move the `accessList` outside of the block and need to have special handling of the first loop. I think it may be easiest to parse the way it is.

PR #347, opened 2025-07-23
fix: add break label in ws loop to properly escape outer loop

almk-dev:
# Description
Closes: #348

The WebSocket logic conditionally uses `return`, `continue`, and `break` as part of its control flow. However, the `break` statements inside the `switch` block is a no-op as it just exits the switch (which will happen anyways as it's the last statement in the case).

If the original intent is to exit the read loop and stop the socket flow entirely, we need a label for the outer loop and `break` that instead.

---

## Author Checklist
I have...

## Reviewers Checklist
I have...

- [ ] reviewed content


aljo242:
Is there a way to test this functionality?

almk-dev:
@aljo242 We'd have to devise a test suite for the private methods. It might be more trouble than it's worth as this is just a fix meant to reflex the intent of the original logic and not a behavioral change.

PR #368, opened 2025-07-29
feat: support query gas limit flag

thomas-nguy:
# Description

This PR enforces a max gas cap when calling the endpoint `eth_call` or `eth_estimateGas`

Currently, an attacker can bypass the gas cap by querying directly at the cosmos level and specifying a very large gas cap

If done in a malicious way, an attacker could use a high GasCap to execute an infinite loop function, potentially impacting node performance & crashing the node

Step to reproduce

Deploy

```
// SPDX-License-Identifier: MIT
pragma solidity =0.8.6;

contract InfiniteLoop {
    uint256 public count;

    constructor() {
        count = 0;
    }

    function infiniteLoop() public {
        while (true) {
            count++;
            // This will cause an out-of-gas error if called
            // in a transaction, but will not revert the contract.
            // It will just keep incrementing `count`.
        }
    } 

    function getCount() public view returns (uint256) {
        return count;
    }
}
```

Call

```
curl http://localhost:1317/cosmos/evm/vm/v1/eth_call\?args\=eyJmcm9tIjogIjB4QzZGZTVEMzM2MTVhMUM1MmMwODAxOGM0N0U4QmM1MzY0NkEwRTEwMSIsICJ0byI6ICIweDNENjQxYTI3OTE1MzNCNEEwMDAwMzQ1ZUE4ZDUwOWQwMUUxZWMzMDEiLCJkYXRhIjogIjB4MWRiZjM1M2QifQ==\&gas_cap\=18446744073709551615
The base64 decoded value of args is {"from": "0xC6Fe5D33615a1C52c08018c47E8Bc53646A0E101", "to": "0x3D641a2791533B4A0000345eA8d509d01E1ec301","data": "0x1dbf353d"} where from is the address of user1 , to is the address of deployed contract and data is the signature of infiniteLoop() function.
Change the to address to the address of the deployed smart contract during testing if it changes in your setup.
```

This PR remediates it by setting a global query gas limit

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

**All** items are required.

I have...

- [ ] reviewed content

### DIFF
--- a/evmd/app.go
+++ b/evmd/app.go
@@ -481,6 +481,7 @@ func NewExampleApp(
 		app.FeeMarketKeeper,
 		&app.Erc20Keeper,
 		tracer,
+		cast.ToUint64(appOpts.Get(server.FlagQueryGasLimit)),
 	)
 
 	app.Erc20Keeper = erc20keeper.NewKeeper(
--- a/x/vm/keeper/grpc_query.go
+++ b/x/vm/keeper/grpc_query.go
@@ -240,7 +240,17 @@ func (k Keeper) EthCall(c context.Context, req *types.EthCallRequest) (*types.Ms
 	nonce := k.GetNonce(ctx, args.GetFrom())
 	args.Nonce = (*hexutil.Uint64)(&nonce)
 
-	msg, err := args.ToMessage(req.GasCap, cfg.BaseFee, false, false)
+	// Enforce the gas limit cap
+	gasCap := req.GasCap
+	if k.queryGasLimit != GasNoLimit {
+		if gasCap == 0 {
+			gasCap = k.queryGasLimit
+		} else if k.queryGasLimit < gasCap {
+			gasCap = k.queryGasLimit
+		}
+	}
+
+	msg, err := args.ToMessage(gasCap, cfg.BaseFee, false, false)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
@@ -283,11 +293,20 @@ func (k Keeper) EstimateGasInternal(c context.Context, req *types.EthCallRequest
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	// Enforce the gas limit cap
+	gasCap := req.GasCap
+	if k.queryGasLimit != GasNoLimit {
+		if gasCap == 0 {
+			gasCap = k.queryGasLimit
+		} else if k.queryGasLimit < gasCap {
+			gasCap = k.queryGasLimit
+		}
+	}
+
 	// Binary search the gas requirement, as it may be higher than the amount used
 	var (
-		lo     = ethparams.TxGas - 1
-		hi     uint64
-		gasCap uint64
+		lo = ethparams.TxGas - 1
+		hi uint64
 	)
 
 	// Determine the highest gas limit can be used during the estimation.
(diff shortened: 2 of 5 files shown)


x/vm/keeper/grpc_query.go
```diff
@@ -240,7 +240,17 @@ func (k Keeper) EthCall(c context.Context, req *types.EthCallRequest) (*types.Ms
...
-	msg, err := args.ToMessage(req.GasCap, cfg.BaseFee, false, false)
+	// Enforce the gas limit cap
+	gasCap := req.GasCap
+	if k.queryGasLimit != GasNoLimit {
+		if gasCap == 0 {
+			gasCap = k.queryGasLimit
+		} else if k.queryGasLimit < gasCap {
+			gasCap = k.queryGasLimit
+		}
+	}
```
aljo242: can we refactor this to a common function?

thomas-nguy: done

x/vm/types/tx_args.go
```diff
@@ -182,7 +181,8 @@ func (args *TransactionArgs) ToMessage(globalGasCap uint64, baseFee *big.Int, sk
 	// Set default gas & gas price if none were set
 	gas := globalGasCap
 	if gas == 0 {
-		gas = uint64(math.MaxUint64 / 2)
+		// Ethereum block size is ~36000000, we use this value as default in case neither gas or global cap is specified, to protect against DOS
+		gas = uint64(100000000)
```
aljo242: could we exclude this change from the PR and add in a separate one?  The rest of the PR looks good, but having this would be 2 changes in one PR.  We can review this separately and discuss

thomas-nguy: no problem, done

evmd/app.go
```diff
@@ -482,6 +482,7 @@ func NewExampleApp(
 		&app.ConsensusParamsKeeper,
 		&app.Erc20Keeper,
 		tracer,
+		cast.ToUint64(appOpts.Get(server.FlagQueryGasLimit)),
```
technicallyty, line 495: this flag is meant to control the cosmos SDK query gas limit. im not sure if it makes sense to use it in this way.

imo it makes more sense to set the value like this:

[root.go](https://github.com/cosmos/evm/blob/369defb94ac5c1cc63751c39eb2b24ba6f1a9be4/evmd/cmd/evmd/cmd/root.go#L301-L315)
```
baseapp.SetQueryGasLimit(cast.ToUint64(appOpts.Get(sdkserver.FlagQueryGasLimit))),
```

this will make the SDK set a gas limit in context during query executions. then we can update the keeper calls to respect the value set here: https://github.com/cosmos/cosmos-sdk/blob/553f8955c3214c21c2755811aae5a8f3e021f0f8/baseapp/abci.go#L1282-L1286

-----

keep in mind, this would enforce a query gas limit for **ALL** gRPC queries. if we want a separate gas limit (i.e. no gas limit for SDK queries, but one for EVM), i suggest we create a separate flag.

thomas-nguy: > if we want a separate gas limit (i.e. no gas limit for SDK queries, but one for EVM), i suggest we create a separate flag.

Do we want to support such use case? The `FlagQueryGasLimit` is meant to set a max gas limit for all Rest/Grpc queries, intended to prevent DOS attack
My original thought is that the x/evm module rpcs is a subset of those queries (as they all endup calling cosmos level rest/grpc) and need to follow the same limitation. Adding an extra flag may bring an additional layer of complexity, and confuse users.

I can do the change if it is really needed though

technicallyty: main reason i would think to separate them is because im unsure of the relation between eth gas units and SDK gas units. i know eth gas costs are quite calculated and well thought out, whereas the SDK's feels more arbitrarily chosen.

any thoughts here? @aljo242? i know youve mentioned we dont want to have too many ways to configure things these days, so maybe lets just go with a catch-all flag

aljo242: I think it makes sense to put it under the one flag

PR #387, opened 2025-07-31
(feat):Appside Mempool

vladjdk:
# Appside Mempool Implementation

**Closes: #417**

*Disclaimer: This PR is meant to introduce an *experimental* version of the appside mempool to the main branch so that teams can begin testing their workflows and highlight improvements that need to be made.*

Please read the issue document for the description of the architecture before reviewing this PR.

We Introduce an appside mempool, fully compatible with EVM transaction workflows. The main usecase that it solves for is out-of-order transactions, and transactions with nonce gaps.

### Mempool
The `/mempool` folder includes all of the necessary components that build out the mempool. It implements the `ExtMempool` interface from the Cosmos SDK and uses the `Iterator`. Note that it tracks two separate mempools: one for EVM transactions, and it uses a `PriorityNonceMempool` to track other Cosmos transactions.

#### ExtMempool Implementation
The mempool implements the `ExtMempool` interface, which is used by the Comet mempool for insertion and deletion, and the `PrepareProposalHandler` for selection and inclusion into blocks.

#### Iterator Implementation
The iterator combines the EVM `TxPool` with the Cosmos `PriorityNonceMempool` to create a single, combined iterator based on priority. The priority algorithm can be set to a custom one in the constructor.

#### CheckTxHandler
The `CheckTxHandler` adds a special failure case to out-of-order nonce errors, which will add the transactions into the queued state in the `TxPool` instead of outright rejecting them. **They will not be broadcast to other nodes in the state.**

### TxPool
This is a forked version of the go-ethereum txpool from v1.15.11. The main changes are the following:
- Uses the `vm.StateDB` interface instead of the StateDB object from go-ethereum.
- Implements a `BroadcastTxFn` that broadcasts back to the Comet mempool once a transaction is promoted from queued to `pending`.

**⚠️ Known Inefficiencies**: The broadcasting mechanism has identified inefficiencies where transactions can be broadcast multiple times due to the promotion/demotion cycle. See comments in `mempool/txpool/legacypool/legacypool.go:1327-1332` for details. However, more testing needs to be done as Comet may cache already sent transactions to prevent them from being sent again.

### Miner
This is a forked ordering mechanism from go-ethereum v1.15.11 that automatically creates a list, similar to an iterator, based on priority and account. Nothing was changed here so far.

### Registry
This file registers a global variable for the mempool. It is *only* used and accessed by the RPC, which reads from the pool after initialization. There is a testing version, and a production version. The latter disallows the mempool pointer to be overwritten, which is required in tests.

### RPC Additions
The `/txpool` namespace was added, and all of the function stubs now contain logic that look into the mempool. This *only* includes the basic transaction pool querying functionality.

### Instant Finality Design
The mempool is specifically designed for Cosmos chains with instant finality where reorgs never occur. This affects several design decisions:
- `GetBlock()` method panics if called for blocks beyond block 1, as this would indicate an attempted reorg
- Reorg logic is skipped in testing mode for Cosmos chains
- The `Blockchain` implementation assumes no reorgs will happen

### Asynchronous Promotion
The mempool uses asynchronous promotion via goroutines (`scheduleReorgLoop()`), which can create race conditions between promotion and block proposal timing. This is a known issue that may affect transaction inclusion in fast block-time chains by including them in later blocks than expected.

### Transaction States
- **Pending**: Immediately executable transactions (nonce matches account state)
- **Queued**: Future transactions waiting for nonce gaps to be filled
- **Promotion**: Asynchronous movement from queued to pending via `promoteExecutables()`

## Tests
We have basic single integration tests and multi-node system tests which send out-of-order transactions. We still have yet to test the broadcasting and block inclusion mechanics. These will be included in a subsequent PR.

**Test Coverage Areas**:
- Nonce gap scenarios and queued → pending transitions
- Fee prioritization and transaction ordering verification
- Invalid transaction handling and error scenarios

## Known Issues & Limitations

1. **Broadcasting Inefficiencies**: Transactions may be broadcast multiple times due to promotion/demotion cycles. More testing is required to verify proper Comet mempool broadcasting caches.
2. **Asynchronous Promotion Race Conditions**: Fast block times may outpace promotion goroutines, causing  transactions to appear in later blocks than expected
3. **Experimental Status**: This is an experimental implementation requiring further testing for the p2p layer

## Block Height Requirement

The mempool requires **block 1 to be committed** before accepting any transactions. This is a critical synchronization requirement that affects the entire transaction flow.

#### Context Query Lag
The mempool queries the blockchain context to determine the current state, but this context **lags by one block**. This means:

- At block 0: Context is not yet available
- At block 1: Context reflects block 0 state
- At block 2: Context reflects block 1 state

#### Block Announcement Timing
The mempool receives block announcements during the `EndBlocker` phase, which means:

1. **Block N is proposed and executed**
2. **EndBlocker runs** → Mempool receives block N announcement
3. **Block N+1 begins** → Context now reflects block N state
4. **Transactions can be processed** using the updated context

## Future Improvements

- Add comprehensive testing for broadcasting and block inclusion mechanics
- Performance optimization for duplicate broadcasting, antehandler
- Testing various fee scenarios with the skewed context
- Adding more test cases for scripts and deployments (OP Superchain, arbitrage)


ante/evm/09_increment_sequence.go
```diff
@@ -20,6 +21,12 @@ func IncrementNonce(
 	// we merged the nonce verification to nonce increment, so when tx includes multiple messages
 	// with same sender, they'll be accepted.
 	if txNonce != nonce {
+		if txNonce > nonce {
+			return errorsmod.Wrapf(
+				mempool.ErrNonceGap,
+				"tx nonce: %d, account nonce: %d", txNonce, nonce,
+			)
+		}
```
Eric-Warehime: Why do we need a custom error for this case?

tests/systemtests/Counter/README.md
Eric-Warehime, line 1: Is this useful to copy into here?

evmd/app.go
```diff
@@ -838,7 +869,16 @@ func (app *EVMD) BeginBlocker(ctx sdk.Context) (sdk.BeginBlock, error) {
...
 // EndBlocker application updates every end block
 func (app *EVMD) EndBlocker(ctx sdk.Context) (sdk.EndBlock, error) {
-	return app.ModuleManager.EndBlock(ctx)
+	endBlock, err := app.ModuleManager.EndBlock(ctx)
+	if err != nil {
+		return endBlock, err
+	}
+
+	if app.EVMMempool != nil {
+		app.EVMMempool.GetBlockchain().NotifyNewBlock()
```
Eric-Warehime: Why should we customize the endblocker like this instead of passing the mempool in the /xvm module and writing an endblocker implementation there which handles block transitions? This seems kind of hacky as is and will further increase the complexity of wiring up the EVM to chains.

aljo242:
@vladjdk follow up item should be creating release notes for `v0.4.0` and documenting experimental status along with the timeline for full release in `v0.5.0`

.markdownlintignore
```diff
@@ -0,0 +1 @@
+tests/systemtests/Counter
```
aljo242:
```suggestion
tests/systemtests/Counter

```

ante/evm/09_increment_sequence.go
```diff
@@ -22,6 +23,12 @@ func IncrementNonce(
 	// we merged the nonce verification to nonce increment, so when tx includes multiple messages
 	// with same sender, they'll be accepted.
 	if txNonce != nonce {
+		if txNonce > nonce {
```
aljo242: nit: can we update the variable name for `nonce` in this function to be `currentNonce` or something similar for readability

evmd/app.go
```diff
@@ -3,6 +3,9 @@ package evmd
 import (
 	"encoding/json"
 	"fmt"
+	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
+	"github.com/cosmos/evm/mempool"
+	evmserver "github.com/cosmos/evm/server"
 	"io"
 	"os"
```
aljo242: formatting

evmd/cmd/evmd/cmd/testnet.go
```diff
@@ -297,19 +306,25 @@ func initTestnetFiles(
...
+			nodeConfig.Instrumentation.Prometheus = false
+			nodeConfig.RPC.PprofListenAddress = ""
 			evmCfg.Config.API.Address = fmt.Sprintf("tcp://0.0.0.0:%d", sdkAPIPort+portOffset)
 			evmCfg.Config.GRPC.Address = fmt.Sprintf("0.0.0.0:%d", sdkGRPCPort+portOffset)
 			evmCfg.JSONRPC.Address = fmt.Sprintf("127.0.0.1:%d", evmJSONRPC+evmPortOffset)
 			evmCfg.JSONRPC.MetricsAddress = fmt.Sprintf("127.0.0.1:%d", evmJSONRPCMetrics+evmPortOffset)
 			evmCfg.JSONRPC.WsAddress = fmt.Sprintf("127.0.0.1:%d", evmJSONRPCWS+evmPortOffset)
+			evmCfg.JSONRPC.Enable = true
+			evmCfg.JSONRPC.EnableIndexer = true
+			evmCfg.JSONRPC.API = []string{"eth", "txpool", "personal", "net", "debug", "web3"}
```
aljo242: can we export these as constants from our RPC library?

`rpc.NamespaceEth` etc

can create a follow up task

evmd/cmd/evmd/cmd/testnet.go
```diff
@@ -371,13 +386,18 @@ func initTestnetFiles(
 		accTokens := sdk.TokensFromConsensusPower(1000, sdk.DefaultPowerReduction)
 		accStakingTokens := sdk.TokensFromConsensusPower(500, sdk.DefaultPowerReduction)
 		coins := sdk.Coins{
-			sdk.NewCoin("testtoken", accTokens),
+			sdk.NewCoin("atest", accTokens),
```
aljo242: let's use a constant

evmd/go.mod
```diff
@@ -17,7 +17,7 @@ require (
 	cosmossdk.io/x/upgrade v0.2.0
 	github.com/cometbft/cometbft v0.38.17
 	github.com/cosmos/cosmos-db v1.1.3
-	github.com/cosmos/cosmos-sdk v0.53.3
+	github.com/cosmos/cosmos-sdk v0.53.4-0.20250725182846-908df9d4e2e0
```
aljo242: Do we need to use this hash or can we use an actual release tag?

go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/cometbft/cometbft v0.38.17
 	github.com/cosmos/cosmos-db v1.1.3
 	github.com/cosmos/cosmos-proto v1.0.0-beta.5
-	github.com/cosmos/cosmos-sdk v0.53.3
+	github.com/cosmos/cosmos-sdk v0.53.4-0.20250725182846-908df9d4e2e0
```
aljo242: same here - do we need a new release? or can we use v0.53.4

mempool/blockchain.go
```diff
@@ -0,0 +1,158 @@
...
+	_ txpool.BlockChain     = Blockchain{}
+	_ legacypool.BlockChain = Blockchain{}
+)
+
+// Blockchain implements the BlockChain interface required by Ethereum transaction pools.
+// It bridges Cosmos SDK blockchain state with Ethereum's transaction pool system by providing
+// access to block headers, chain configuration, and state databases. This implementation is
+// specifically designed for instant finality chains where reorgs never occur.
+type Blockchain struct {
+	ctx                func(height int64, prove bool) (sdk.Context, error)
```
aljo242: func(height int64, prove bool) (sdk.Context, error)

should we define this function as a named type?

`type getCtxCallback = func(height int64, prove bool) (sdk.Context, error)`

mempool/blockchain.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return statedb.New(ctx, b.vmKeeper, statedb.NewEmptyTxConfig(common.Hash(ctx.BlockHeader().AppHash))), nil
+}
+
+// GetLatestCtx retrieves the most recent query context from the application.
+// This provides access to the current blockchain state for transaction validation and execution.
+func (b Blockchain) GetLatestCtx() (sdk.Context, error) {
+	ctx, err := b.ctx(0, false)
+	if err != nil {
+		return sdk.Context{}, errors2.Wrapf(err, "getting latest context")
```
aljo242: can we rename the `errors2` import to something else?

mempool/blockchain.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+	ctx = ctx.WithBlockGasMeter(types2.NewGasMeter(b.blockGasLimit))
```
aljo242: same here `types2`

mempool/mempool.go
```diff
@@ -0,0 +1,395 @@
...
+)
+
+var _ sdkmempool.ExtMempool = &EVMMempool{}
+
+type (
+	// EVMMempool is a unified mempool that manages both EVM and Cosmos SDK transactions.
+	// It provides a single interface for transaction insertion, selection, and removal while
+	// maintaining separate pools for EVM and Cosmos transactions. The mempool handles
+	// fee-based transaction prioritization and manages nonce sequencing for EVM transactions.
+	EVMMempool struct {
```
aljo242: I think it most likely makes sense to add a logger to this.  Right now we really don't have any logging from this new mempool logic and it could be useful to add debug logs to make our lives easier in the future. Happy to just have this as a follow-up task

mempool/mempool.go
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// NewEVMMempool creates a new unified mempool for EVM and Cosmos transactions.
+// It initializes both EVM and Cosmos transaction pools, sets up blockchain interfaces,
+// and configures fee-based prioritization. The config parameter allows customization
+// of pools and verification functions, with sensible defaults created if not provided.
+func NewEVMMempool(ctx func(height int64, prove bool) (sdk.Context, error), vmKeeper VMKeeperI, feeMarketKeeper FeeMarketKeeperI, txConfig client.TxConfig, clientCtx client.Context, config *EVMMempoolConfig) *EVMMempool {
+	var txPool *txpool.TxPool
+	var cosmosPool sdkmempool.ExtMempool
+	var anteHandler sdk.AnteHandler
```
aljo242: plz group

mempool/mempool.go
```diff
@@ -0,0 +1,395 @@
...
+
+	if config.BlockGasLimit == 0 {
+		config.BlockGasLimit = 100_000_000
+	}
+
+	var blockchain *Blockchain
+
+	// Default txPool
+	if txPool == nil {
+		blockchain = NewBlockchain(ctx, vmKeeper, feeMarketKeeper, config.BlockGasLimit)
```
aljo242: does `NewBlockchain` need to be exported?

mempool/mempool.go
```diff
@@ -0,0 +1,395 @@
...
+	anteHandler = config.AnteHandler
+
+	if config.BlockGasLimit == 0 {
+		config.BlockGasLimit = 100_000_000
+	}
+
+	var blockchain *Blockchain
+
+	// Default txPool
+	if txPool == nil {
```
aljo242: I think this code is more structurally readable if we do something like this:

```suggestion
	// Default txPool
	txPool := config.TxPool
	if txPool == nil {
```

We assign from the config and then immediately check the value

mempool/mempool.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+		cosmosPool = sdkmempool.NewPriorityMempool(priorityConfig)
+	}
+
+	if len(txPool.Subpools) != 1 {
+		panic("tx pool should contain one subpool")
+	}
+	if _, ok := txPool.Subpools[0].(*legacypool.LegacyPool); !ok {
+		panic("tx pool should contain only legacypool")
+	}
```
aljo242: These checks should be right after we set up `txPool`, not after we set up the `cosmosPool`

rpc/backend/call_tx.go
```diff
@@ -151,6 +152,12 @@ func (b *Backend) SendRawTransaction(data hexutil.Bytes) (common.Hash, error) {
 		err = errorsmod.ABCIError(rsp.Codespace, rsp.Code, rsp.RawLog)
 	}
 	if err != nil {
+		// Check if this is a nonce gap error that was successfully queued
+		if strings.Contains(err.Error(), "tx nonce is higher than account nonce") {
```
aljo242: I think it would be safer to check the typed error itself using `errors.Is` since it should be the same type every time

evmd/app.go
```diff
@@ -752,6 +758,31 @@ func NewExampleApp(
 
 	app.setAnteHandler(app.txConfig, maxGasWanted)
 
+	// set the EVM priority nonce mempool
+	// If you wish to use the noop mempool, remove this codeblock
+	if evmtypes.GetChainConfig() != nil {
+		// Get the actual block gas limit from consensus parameters
```
zsystm:
```suggestion
		// TODO: Get the actual block gas limit from consensus parameters
```

evmd/tests/integration/create_app.go
```diff
@@ -21,18 +21,21 @@ import (
...
 	}
-	// create evmos app
+
 	db := dbm.NewMemDB()
 	logger := log.NewNopLogger()
 	loadLatest := true
 	appOptions := simutils.NewAppOptionsWithFlagHome(defaultNodeHome)
-	baseAppOptions := append(customBaseAppOptions, baseapp.SetChainID(chainID)) //nolint:gocritic
+
+	// Use noop mempool for regular integration tests to avoid EVM mempool state issues
```
zsystm: Would many test cases fail without using the noop mempool?
If so, could you briefly note which kinds of tests fail and why?

No need to list every failing case—just a representative example would help future developers understand the context.

evmd/tests/integration/create_app.go
```diff
@@ -21,18 +21,21 @@ import (
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 )
 
-// CreateEvmd creates an evmos app
+// CreateEvmd creates an evmos app for regular integration tests (non-mempool)
```
Eric-Warehime:
```suggestion
// CreateEvmd creates an evm app for regular integration tests (non-mempool)
```

evmd/cmd/evmd/cmd/root.go
```diff
@@ -315,25 +313,6 @@ func newApp(
...
-				MaxTx:           maxTxs,
-			})
-		} else {
-			mpool = mempool.NoOpMempool{}
-		}
-		app.SetMempool(mpool)
-		handler := baseapp.NewDefaultProposalHandler(mpool, app)
-		app.SetPrepareProposal(handler.PrepareProposalHandler())
-		app.SetProcessProposal(handler.ProcessProposalHandler())
-	})
```
aljo242, line 338: so now we are just using default for everything in the default evmd app?

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+		chainID:        chainID,
+	}
+}
+
+// shouldUseEVM determines which transaction type to prioritize based on fee comparison.
+// Returns true if the EVM transaction should be selected, false if Cosmos transaction should be used.
+// EVM transactions are preferred when Cosmos transactions lack fees or use non-bondDenom denominations.
+func (i *EVMMempoolIterator) shouldUseEVM() bool {
+	var nextEVMTx *txpool.LazyTransaction
+	var evmFee *uint256.Int
```
aljo242: group

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+	if i.evmIterator != nil {
+		nextEVMTx, evmFee = i.evmIterator.Peek()
+	}
+
+	var nextCosmosTx sdk.Tx
+	if i.cosmosIterator != nil {
```
aljo242: can `i.cosmosIterator` and `i.evmIterator` both not be nil?

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+		txConfig:       txConfig,
+		bondDenom:      bondDenom,
+		chainID:        chainID,
+	}
+}
+
+// shouldUseEVM determines which transaction type to prioritize based on fee comparison.
+// Returns true if the EVM transaction should be selected, false if Cosmos transaction should be used.
+// EVM transactions are preferred when Cosmos transactions lack fees or use non-bondDenom denominations.
+func (i *EVMMempoolIterator) shouldUseEVM() bool {
```
aljo242: This function is very confusing to read.  The godoc should explicitly outline all of the rules and assumptions in a simple bullet list probably.  We should also make sure this function is unit tested

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+	// Determine what transaction we just returned and advance the correct iterator
+	var nextEVMTx *txpool.LazyTransaction
+	if i.evmIterator != nil {
+		nextEVMTx, _ = i.evmIterator.Peek()
+	}
+
+	var nextCosmosTx sdk.Tx
+	if i.cosmosIterator != nil {
+		nextCosmosTx = i.cosmosIterator.Tx()
+	}
```
aljo242: feels like this logic is repeated a lot.  maybe we should have small helper functions like

i.getNextEVMTx ->  *tx, bool

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+
+	// Check if we still have transactions after advancing
+	var hasMoreEVM, hasMoreCosmos bool
+	if i.evmIterator != nil {
+		nextEVMTx, _ = i.evmIterator.Peek()
+		hasMoreEVM = nextEVMTx != nil
+	}
+	if i.cosmosIterator != nil {
+		hasMoreCosmos = i.cosmosIterator.Tx() != nil
+	}
```
aljo242: why do we call `Peek` on the evmIterator but `Tx` on the cosmos iterator?  Feels like since both should have the same unified API we should be able to just do the same logic on both for this check

rpc/backend/tx_pool.go
```diff
@@ -13,15 +19,90 @@ func (b *Backend) Content() (map[string]map[string]map[string]*types.RPCTransact
...
+		return content, nil
+	}
+
+	// Get pending (runnable) and queued (blocked) transactions from the mempool
+	runnable, blocked := evmMempool.GetTxPool().Content()
+
+	// Convert pending (runnable) transactions
+	for addr, txList := range runnable {
+		addrStr := addr.Hex()
+		if content["pending"][addrStr] == nil {
```
aljo242: looks like we should make a string constant for `"pending"` and `"queued`"

rpc/backend/tx_pool.go
```diff
@@ -13,15 +19,90 @@ func (b *Backend) Content() (map[string]map[string]map[string]*types.RPCTransact
...
 	}
+
+	// Get the global mempool instance
+	evmMempool := mempool.GetGlobalEVMMempool()
+	if evmMempool == nil {
+		return content, nil
+	}
+
+	// Get pending (runnable) and queued (blocked) transactions from the mempool
+	runnable, blocked := evmMempool.GetTxPool().Content()
```
aljo242: Does it make sense to call these return variabled `pending, queued` instead to align the logic throughout this code.  We can add a godoc callout that `pending -> tx is runnable at any point` and `queued -> tx is in mempool but currently blocked on execution (nonce issue etc)`

rpc/backend/tx_pool.go
```diff
@@ -13,15 +19,90 @@ func (b *Backend) Content() (map[string]map[string]map[string]*types.RPCTransact
...
+	evmMempool := mempool.GetGlobalEVMMempool()
+	if evmMempool == nil {
+		return content, nil
 	}
+
+	// Get transactions for the specific address
+	pending, queue := evmMempool.GetTxPool().ContentFrom(addr)
+
+	// Build the pending transactions
+	dump := make(map[string]*types.RPCTransaction, len(pending))
```
aljo242: better var name needed than dump

rpc/backend/tx_pool.go
```diff
@@ -31,13 +112,121 @@ func (b *Backend) Inspect() (map[string]map[string]map[string]string, error) {
 		"pending": make(map[string]map[string]string),
 		"queued":  make(map[string]map[string]string),
 	}
+
+	// Get the global mempool instance
+	evmMempool := mempool.GetGlobalEVMMempool()
+	if evmMempool == nil {
+		return inspect, nil
+	}
```
aljo242, line 127: should we add this to the top of the function and just return `nil, nil` to avoid allocation?

rpc/backend/tx_pool.go
```diff
@@ -31,13 +112,121 @@ func (b *Backend) Inspect() (map[string]map[string]map[string]string, error) {
...
+
+	// Get pending (runnable) and queued (blocked) transactions from the mempool
+	runnable, blocked := evmMempool.GetTxPool().Content()
```
aljo242: sometimes we use `pending, queued`, sometimes we use this

tests/systemtests/.gitignore
```diff
@@ -0,0 +1,3 @@
+testnet/
+binaries/
+Counter/broadcast/*
```
aljo242:
```suggestion
Counter/broadcast/*

```

tests/systemtests/go.mod
```diff
@@ -2,78 +2,88 @@ module github.com/evmos/tests/systemtests
...
+)
 
 require (
-	cosmossdk.io/api v0.9.2 // indirect
-	cosmossdk.io/collections v1.2.0 // indirect
-	cosmossdk.io/core v0.11.3 // indirect
-	cosmossdk.io/depinject v1.2.0 // indirect
+	cosmossdk.io/api v1.0.0-rc.1 // indirect
+	cosmossdk.io/collections v1.3.1 // indirect
+	cosmossdk.io/core v1.1.0-rc.1 // indirect
```
aljo242: bumping to these versions means we are moving to `v0.54.x` family in this packages.  Might be fine - just flagging

ante/evm/09_increment_sequence.go
Eric-Warehime: Maybe we can just change the existing error type if you're unable to use `ErrInvalidSequence` here?

mempool/iterator.go
vladjdk: EVM iterator is never nil, but it can have 0 objects.

evmd/cmd/evmd/cmd/root.go
vladjdk: We set up the appside mempool in app.go now. It overwrites this one, so having both is kind of confusing.

mempool/iterator.go
vladjdk: The EVM iterator is taken directly from geth, and the Cosmos iterator follows the Iterator interface. I could make a wrapper for the EVM iterator to make it follow the same iterator if you think that's useful.

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil
+	}
+	msgEthereumTx := &msgtypes.MsgEthereumTx{}
+	if err := msgEthereumTx.FromSignedEthereumTx(nextEVMTx.Tx, ethtypes.LatestSignerForChainID(i.chainID)); err != nil {
+		return nil // Return nil for invalid tx instead of panicking
+	}
+	cosmosTx, err := msgEthereumTx.BuildTx(i.txConfig.NewTxBuilder(), i.bondDenom)
+	if err != nil {
+		return nil
+	}
```
Eric-Warehime: Does silent failure here potentially represent a class of txns which can get into the mempool, but never be removed since they fail conversion for some reason? 

Should probably at least log errors here no?

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+			// We used EVM transaction, advance EVM iterator
+			if i.evmIterator != nil {
+				i.evmIterator.Shift()
+			}
+		} else {
+			// No EVM tx, we used cosmos, advance cosmos
+			if i.cosmosIterator != nil {
+				i.cosmosIterator = i.cosmosIterator.Next()
+			}
+		}
```
Eric-Warehime: What are the cases in which our priority function tells us we should have selected an EVM txn, but the nextEVMTx is actually nil? 

Does this function get called concurrently with the maintenance loop which is reaping txns? i.e. can the mempool be mutated underneath the iterator?

mempool/iterator.go
```diff
@@ -0,0 +1,218 @@
...
+	if i.evmIterator != nil {
+		nextEVMTx, _ = i.evmIterator.Peek()
+	}
+
+	var nextCosmosTx sdk.Tx
+	if i.cosmosIterator != nil {
+		nextCosmosTx = i.cosmosIterator.Tx()
+	}
+
+	// If no transactions available, this shouldn't happen if iterator was created properly
```
Eric-Warehime: Should probably log either a warn or error on exceptional cases.

ante/evm/09_increment_sequence.go
vladjdk: A nonce gap is a tx nonce that is greater than the current account sequence, and a sequence error would be for when the tx nonce is lesser. I could parse the string for number comparisons, but that seemed more convoluted than just using a different error.

tests/systemtests/Counter/README.md
vladjdk: Took this from the merging in the system tests PR, no longer diffed here. Could open a different issue to clean this up.

ante/evm/09_increment_sequence.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/9bb166e1bc36ceaeae45068feb38535119b1391b

evmd/app.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/ac25c26467d3f18af34a32b08aa3527a94def983

evmd/cmd/evmd/cmd/testnet.go
vladjdk: good first issue: https://github.com/cosmos/evm/issues/421

evmd/cmd/evmd/cmd/testnet.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/e4c19130b20137c1dbd694e53adcb6199d84377b

evmd/go.mod
vladjdk: can use the release tag: https://github.com/cosmos/evm/pull/387/commits/99b8e772a9353715ae4b4cf085501a78866dc1a5

go.mod
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/ce52d42c83320c83bd9bbf4db6285ad4d00a9dcb

mempool/blockchain.go
vladjdk: fixeroo: https://github.com/cosmos/evm/pull/387/commits/e01997724de0f20173603fb9725bdb80d808bc27

mempool/blockchain.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/83b119961b48e68af8bd114e94afab45ffefd02e

mempool/blockchain.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/83b119961b48e68af8bd114e94afab45ffefd02e

mempool/mempool.go
vladjdk: grouped: https://github.com/cosmos/evm/pull/387/commits/c87f01b55c9a1b33e7a3956491ea615bb3b4d5ca

mempool/mempool.go
vladjdk: privatized: https://github.com/cosmos/evm/pull/387/commits/072d23f619ea3e0a89c08a2dfc3eec48a431a2ae

mempool/mempool.go
vladjdk: fixed for both pools: https://github.com/cosmos/evm/pull/387/commits/aa207d7088689793ed7cef7a26a94bd966bcca8a

mempool/mempool.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/58764fad9ec506f10950af095478c2601b1d0267

rpc/backend/call_tx.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/49a7badf2391589980b4d8a69b5e48f710e9cd39

mempool/iterator.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/76d9ae3773f1ec6d945ed0e954761a9c7e81817f

mempool/iterator.go
vladjdk: Refactored to (hopefully) make this whole file more readable: https://github.com/cosmos/evm/pull/387/commits/3e8f87a65777dce8d5a8bc86d8619f5b903716f6

mempool/iterator.go
vladjdk: fixed in the refactor above^ https://github.com/cosmos/evm/pull/387/commits/3e8f87a65777dce8d5a8bc86d8619f5b903716f6

rpc/backend/tx_pool.go
vladjdk: Decided to change this and rename the comments as well: https://github.com/cosmos/evm/pull/387/commits/5488611445a0a70b5a70f209abe98102a9b9b2f2

rpc/backend/tx_pool.go
vladjdk: added some explanations https://github.com/cosmos/evm/pull/387/commits/c9fbb4945642139c089b0c4f6f93023927bdffff

rpc/backend/tx_pool.go
vladjdk: fixed: https://github.com/cosmos/evm/pull/387/commits/33fd5d31483dfb048d75f1537e8cd4b7170ec77f

rpc/backend/tx_pool.go
vladjdk: I think it's best to return the structured maps here to conform to the JSON-RPC spec at all times and just show it as empty. This is what currently happens if we use the noop mempool for instance.

rpc/backend/tx_pool.go
vladjdk: Renamed: https://github.com/cosmos/evm/commit/5488611445a0a70b5a70f209abe98102a9b9b2f2

mempool/iterator.go
vladjdk: Yeah, the tx could theoretically be cleaned up while this is executing, though highly unlikely. This could happen if the baseFee increases, triggers the routine to clean up, and a transaction gets removed while this is executed. We should just keep this in for sanity in case something ever does happen like that.

mempool/iterator.go
vladjdk: Refactored [here](https://github.com/cosmos/evm/pull/387/commits/3e8f87a65777dce8d5a8bc86d8619f5b903716f6). Let me know if it's more readable now

mempool/iterator.go
vladjdk: Added in the refactor as well: https://github.com/cosmos/evm/pull/387/commits/3e8f87a65777dce8d5a8bc86d8619f5b903716f6

rpc/backend/tx_pool.go
vladjdk: In the case of this specific error, it means we're probably using something like the noop mempool, in which case we'd just want to show an empty TxPool in the response since we aren't using it.

rpc/backend/tx_pool.go
vladjdk: I changed these all to pending, queued. This was taken from geth, but it doesn't really make sense to me either.

mempool/iterator.go
vladjdk: These should never fail since we know that all the txes in the mempool are valid, but I'm handling the error anyway since it gets returned from the functions.

mempool/mempool.go
vladjdk: Added: https://github.com/cosmos/evm/pull/387/commits/69dd478e398b308098e53bb1078418a1d016ad87

mempool/iterator.go
vladjdk: Added: https://github.com/cosmos/evm/pull/387/commits/69dd478e398b308098e53bb1078418a1d016ad87

evmd/app.go
vladjdk: Fixed: https://github.com/cosmos/evm/pull/387/commits/6fd0eaadfce62a65dfad2b4151c6a24edb08aa55

mempool/iterator.go
```diff
@@ -0,0 +1,300 @@
...
+	cosmosHigher := cosmosFee.Gt(evmFee)
+	i.logger.Debug("comparing transaction fees",
+		"evm_fee", evmFee.String(),
+		"cosmos_fee", cosmosFee.String())
+
+	return !cosmosHigher
+}
+
+// getNextEVMTx retrieves the next EVM transaction and its fee
+func (i *EVMMempoolIterator) getNextEVMTx() (*txpool.LazyTransaction, *uint256.Int) {
```
aljo242, line 167: we should test all of the helper functions in this package (down to add in a follow up)

mempool/registry_production.go
```diff
@@ -0,0 +1,33 @@
+//go:build !test
+// +build !test
+
+package mempool
+
+import "errors"
+
+// globalEVMMempool holds the global reference to the ExperimentalEVMMempool instance.
+// It can only be set during application initialization.
+var globalEVMMempool *ExperimentalEVMMempool
```
aljo242, line 10: could the global mempool be an interface so that we can just have different implemenations (prod and test) instead of using build flags?

We don't have to do this now, I just want to get us away from these build flag shenanigans eventually

vladjdk: Yeah a good idea for cleanup

vladjdk: Maybe we could do this in general for all build flags in a future PR

rpc/backend/tx_pool.go
```diff
@@ -1,43 +1,239 @@
...
+			StatusPending: hexutil.Uint(0),
+			StatusQueued:  hexutil.Uint(0),
+		}, nil
+	}
+
+	pending, queued := evmMempool.GetTxPool().Stats()
 	return map[string]hexutil.Uint{
-		"pending": hexutil.Uint(0),
-		"queued":  hexutil.Uint(0),
+		StatusPending: hexutil.Uint(pending), // #nosec G115 -- overflow not a concern for tx counts
```
aljo242: why is this not a concern? can we add that to the comment?

rpc/backend/tx_pool.go
```diff
@@ -1,43 +1,239 @@
...
+		Value:    (*hexutil.Big)(tx.Value()),
+		V:        (*hexutil.Big)(v),
+		R:        (*hexutil.Big)(r),
+		S:        (*hexutil.Big)(s),
+		ChainID:  (*hexutil.Big)(chainConfig.ChainID),
+	}
+
+	// Handle transaction type specific fields
+	switch tx.Type() {
+	case ethtypes.AccessListTxType:
```
aljo242, line 215: do we want a default case for invalid types?

mempool/iterator.go
vladjdk: Yeah let's do this in a followup, but will start work on that immediately

rpc/backend/tx_pool.go
vladjdk: Sure - there is a maximum limit to the mempool size that is significantly smaller than max uint (by default)

mempool/registry_production.go
aljo242: yeah i will create public task for this

rpc/backend/tx_pool.go
vladjdk: added: https://github.com/cosmos/evm/pull/387/commits/4998a3306ac54e33460bfafbe1112792e9c9b4ad

aljo242:
@vladjdk needs a changelog entry

mempool/iterator.go
vladjdk: https://github.com/cosmos/evm/issues/437

rpc/backend/tx_pool.go
vladjdk: Not here, there are other valid types (like Legacy) that are valid and don't need to be changed.

vladjdk: There should also never be invalid tx in the txPool as they would be rejected on attempted addition

aljo242 (approved):
lgtm

mempool/mempool.go
```diff
@@ -0,0 +1,444 @@
...
+			GetTxPriority: func(goCtx context.Context, tx sdk.Tx) math.Int {
+				cosmosTxFee, ok := tx.(sdk.FeeTx)
+				if !ok {
+					return math.ZeroInt()
+				}
+				found, coin := cosmosTxFee.GetFee().Find(bondDenom)
+				if !found {
+					return math.ZeroInt()
+				}
+				return coin.Amount
```
canu0205: One question about the priority of txs - is it intended to use fee amount instead of gas price? As long as I know current default prioritization of `PriorityNonceMempool` is based on gas price rather than entire fee amount. [(code)](https://github.com/cosmos/cosmos-sdk/blob/v0.53.4/x/auth/ante/validator_tx_fee.go#L58-L63)

mempool/mempool.go
```diff
@@ -0,0 +1,444 @@
...
+	baseFee := m.vmKeeper.GetBaseFee(ctx)
+	var baseFeeUint *uint256.Int
+	if baseFee != nil {
+		baseFeeUint = uint256.MustFromBig(baseFee)
+	}
+
+	m.logger.Debug("getting iterators")
+
+	pendingFilter := txpool.PendingFilter{
+		MinTip:       nil,
```
zsystm, line 406: I think the value of MinTip should match the MinimumGasPrices value, which is configured via the `--minimum-gas-prices` flag when starting evmd, to ensure that the legacy mempool is fully compatible with the Geth mempool.

In Geth, `MinTip` is directly derived from the miner’s GasPrice configuration (see the `fillTransactions` function below). Aligning `MinTip` with `MinimumGasPrices` would make our transaction filtering behavior consistent with Geth’s, especially when selecting pending transactions from the mempool.

```go
// ...
tip := miner.config.GasPrice
filter := txpool.PendingFilter{
    MinTip: uint256.MustFromBig(tip),
}
// ...
```

mempool/iterator.go
```diff
@@ -0,0 +1,300 @@
...
+// 1. Cosmos mempool has no transactions
+// 2. EVM mempool has no transactions (fallback to Cosmos)
+// 3. Cosmos transaction has no fee information
+// 4. Cosmos transaction fee denomination doesn't match bond denom
+// 5. Cosmos transaction fee is lower than the EVM transaction fee
+// 6. Cosmos transaction fee overflows when converted to uint256
+func (i *EVMMempoolIterator) shouldUseEVM() bool {
+	// Get next transactions from both iterators
+	nextEVMTx, evmFee := i.getNextEVMTx()
+	nextCosmosTx, cosmosFee := i.getNextCosmosTx()
```
zsystm, line 138: I noticed that `cosmosFee` is currently computed in a way that differs significantly from `evmFee`.

For example, if we send a simple bank transfer like:
```bash
evmd tx bank send dev2 cosmos1jcltmuhplrdcwp7stlr4hlhlhgd4htqhnu0t2g 100atest --gas-prices 15atest --from dev2 --keyring-backend test
```

the calculation uses the default gas limit (`200,000`) multiplied by the gas price (`15 atest`), resulting in a total fee of `3,000,000 atest`.

By contrast, `evmFee` is computed solely based on the gas price of tx payload, without considering the gas limit.

This means that, for the same gas price, Cosmos transactions will inherently rank lower in priority compared to EVM transactions.

Could you clarify if this behavior is intentional? If not, it may be worth adjusting the calculation so that fee comparison between Cosmos and EVM transactions is done on a fair and consistent basis.

mempool/mempool.go
```diff
@@ -0,0 +1,444 @@
...
+
+	if config.BlockGasLimit == 0 {
+		logger.Debug("block gas limit is 0, setting default", "default_limit", 100_000_000)
+		config.BlockGasLimit = 100_000_000
+	}
+
+	// Default txPool
+	txPool = config.TxPool
+	if txPool == nil {
+		blockchain = newBlockchain(getCtxCallback, logger, vmKeeper, feeMarketKeeper, config.BlockGasLimit)
```
zsystm: When a custom `TxPool` is injected (`config.TxPool` is non-nil), the `blockchain` reference never gets initialized. Later, `Select` and `shouldRemoveFromEVMPool` dereference blockchain, which could trigger a `nil` panic. Was this scenario considered?

evmd/tests/integration/create_app.go
vladjdk: This comment is no longer relevant as I managed to get all the tests to work with the appside mempool. At the time, the issue was that I was using the PrepareProposalVerifyTx for removal verification in the remove function, which wasn't supported in testing contexts. This has been solved with the use of AnteHandler.

Removed the comment here: https://github.com/cosmos/evm/pull/387/commits/1b408071bf8cd42fdbad33cdfec6b0be3b7e368e

mempool/mempool.go
vladjdk: Good catch! https://github.com/cosmos/evm/pull/387/commits/5a22e7b63ce934f45af3def295417ec601e2e9eb

mempool/iterator.go
vladjdk: Nice catch, turns out fee doesn't actually mean fee :)

Fixed here: https://github.com/cosmos/evm/pull/387/commits/64773d72a77b7317fef6a451d015c367f8972f53

vladjdk: cc @aljo242, can you verify the above fix?

mempool/mempool.go
vladjdk: Fixed!

mempool/mempool.go
vladjdk: I'll add this in a subsequent issue, alongside getting the consensus gas price. It's a bit difficult to get these things into app.go, so we might need a different approach.

Issue: https://github.com/cosmos/evm/issues/449

Eric-Warehime (approved):
:shipit:

PR #416, opened 2025-08-08
fix: CometBlockResultByNumber when height is 0

### LINKED ISSUE #488: CometBlockResultByNumber returns an error when height=0
At the moment, when calling CometBlockResultByNumber with height=0, an error is returned:

```
failed to fetch block result from Tendermint: failed to fetch block result from Tendermint 0: height must be greater than 0, but got 0
```

This issue happens when trying to filter logs using Ethereum RPC. The correct behaviour would be to use the latest available block.

Pitasi:
At the moment, when calling TendermintBlockResultByNumber with height=0, an error is returned:

```
failed to fetch block result from Tendermint: failed to fetch block result from Tendermint 0: height must be greater than 0, but got 0
```

This PR replaces 0 with `nil`, making the Tendermint client happy.

This issue happens when trying to filter logs.

Fixes #488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [\#492](https://github.com/cosmos/evm/pull/492) Duplicate case switch to avoid empty execution block
 - [\#509](https://github.com/cosmos/evm/pull/509) Allow value with slashes when query token_pairs
 - [\#495](https://github.com/cosmos/evm/pull/495) Allow immediate SIGINT interrupt when mempool is not empty
+- [\#416](https://github.com/cosmos/evm/pull/416) Fix regression in CometBlockResultByNumber when height is 0 to use the latest block. This fixes eth_getFilterLogs RPC.
 
 ### IMPROVEMENTS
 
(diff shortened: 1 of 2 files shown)


rpc/backend/blocks.go
```diff
@@ -203,6 +203,9 @@ func (b *Backend) TendermintHeaderByNumber(blockNum rpctypes.BlockNumber) (*tmrp
 // TendermintBlockResultByNumber returns a Tendermint-formatted block result
 // by block number
 func (b *Backend) TendermintBlockResultByNumber(height *int64) (*tmrpctypes.ResultBlockResults, error) {
+	if height != nil && *height == 0 {
```
aljo242, line 206: Is there already a test that covers this behavior?

pitasi: hmm I don't think so, but I can pinpoint the cometbft source involved:
https://github.com/cometbft/cometbft/blob/bd1bc3bb6392a973d7b60fd0b19da6e71ca09b47/rpc/core/env.go#L240

a unit test wouldn't be sufficient in this case since we need cometbft for it, so what we could do is an integration test

aljo242 (approved):

aljo242:
@Pitasi could you add a changlog entry - then we can merge

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - [\#471](https://github.com/cosmos/evm/pull/471) Notify new block for mempool in time.
 - [\#492](https://github.com/cosmos/evm/pull/492) Duplicate case switch to avoid empty execution block
+- [\#416](https://github.com/cosmos/evm/pull/416) Fix CometBlockResultByNumber when height is 0
```
aljo242: Could you document here what the new behavior is?  Then we are good to merge

pitasi: done, should be clearer now

Pitasi:
@aljo242 had to sign the commits and push again, can you re-run the ci and re-enable automerge?

Issue #101, opened 2025-04-25
RPC Compatibility Testing


vladjdk:
Write JSON-RPC compatibility testing to ensure compatibility and expected behaviours

PR #419, opened 2025-08-11
test(jsonrpc): add jsonrpc compatibility test

cloudgray:
# Description

Closes: #101 

---

## Author Checklist

I have...


tests/jsonrpc/simulator/namespaces/engine.go
```diff
@@ -0,0 +1,16 @@
+package namespaces
+
+import "github.com/cosmos/evm/tests/jsonrpc/simulator/types"
+
+const (
+	// Engine API namespace (not applicable for Cosmos chains)
+	MethodNameEngineNewPayloadV1        types.RpcName = "engine_newPayloadV1"
+	MethodNameEngineNewPayloadV2        types.RpcName = "engine_newPayloadV2"
+	MethodNameEngineNewPayloadV3        types.RpcName = "engine_newPayloadV3"
```
zsystm, line 9: I couldn’t find any existing tests for this API.

tests/jsonrpc/simulator/namespaces/engine.go
```diff
@@ -0,0 +1,16 @@
...
+	MethodNameEngineForkchoiceUpdatedV1 types.RpcName = "engine_forkchoiceUpdatedV1"
+	MethodNameEngineForkchoiceUpdatedV2 types.RpcName = "engine_forkchoiceUpdatedV2"
+	MethodNameEngineForkchoiceUpdatedV3 types.RpcName = "engine_forkchoiceUpdatedV3"
+	MethodNameEngineGetPayloadV1        types.RpcName = "engine_getPayloadV1"
+	MethodNameEngineGetPayloadV2        types.RpcName = "engine_getPayloadV2"
+	MethodNameEngineGetPayloadV3        types.RpcName = "engine_getPayloadV3"
```
zsystm, line 15: I couldn’t find any existing tests for this API.

tests/jsonrpc/simulator/namespaces/engine.go
```diff
@@ -0,0 +1,16 @@
...
+import "github.com/cosmos/evm/tests/jsonrpc/simulator/types"
+
+const (
+	// Engine API namespace (not applicable for Cosmos chains)
+	MethodNameEngineNewPayloadV1        types.RpcName = "engine_newPayloadV1"
+	MethodNameEngineNewPayloadV2        types.RpcName = "engine_newPayloadV2"
+	MethodNameEngineNewPayloadV3        types.RpcName = "engine_newPayloadV3"
+	MethodNameEngineForkchoiceUpdatedV1 types.RpcName = "engine_forkchoiceUpdatedV1"
+	MethodNameEngineForkchoiceUpdatedV2 types.RpcName = "engine_forkchoiceUpdatedV2"
+	MethodNameEngineForkchoiceUpdatedV3 types.RpcName = "engine_forkchoiceUpdatedV3"
```
zsystm, line 12: I couldn’t find any existing tests for this API.

tests/jsonrpc/simulator/namespaces/debug.go
```diff
@@ -0,0 +1,1077 @@
...
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/cosmos/evm/tests/jsonrpc/simulator/types"
+)
+
+const (
+	NamespaceDebug = "debug"
```
zsystm, line 12: It seems that for the debug namespace, this PR doesn’t compare results from Geth and from evmd.
To ensure JSON-RPC compatibility, every test should treat Geth’s result as the source of truth and verify whether evmd matches it or if anything is missing.

If that’s the case, could you create a follow-up issue for this? We should clarify the test methodology used in this PR and the exact coverage scope.

tests/jsonrpc/scripts/evmd/start-evmd.sh
zsystm, line 1: It looks like we have two start scripts serving similar purposes but in different contexts:
- `container-start-evmd.sh`: runs inside the container (no Docker CLI), handles init/genesis tweaks, then evmd start.
- `start-evmd.sh`: runs on the host, manages Docker lifecycle, port exposure, and applies additional genesis/app.toml tweaks (prometheus, pruning, static precompiles, STRv2 pair, etc.).

To avoid drift and reduce maintenance, could we consolidate shared logic into a single container-side bootstrap (e.g., `scripts/evmd/bootstrap.sh`) controlled via env vars (`ENABLE_PROM`, `ENABLE_STRV2`, `RPC_BIND_ALL`, `BLOCK_MAX_GAS`, …)?
The host script would then be a thin wrapper that only runs Docker and passes `env/ports/volumes`.

If consolidation is out of scope here, can we at least document the intentional differences and extract the common init (keys/genesis/params) into a helper that both call?

tests/jsonrpc/simulator/namespaces/eth.go
```diff
@@ -0,0 +1,2503 @@
...
+	MethodNameEthSubmitWork       types.RpcName = "eth_submitWork"
+	MethodNameEthSubmitHashrate   types.RpcName = "eth_submitHashrate"
+	MethodNameEthCreateAccessList types.RpcName = "eth_createAccessList"
+
+	// Eth namespace - WebSocket-only subscription methods
+	MethodNameEthSubscribe   types.RpcName = "eth_subscribe"
+	MethodNameEthUnsubscribe types.RpcName = "eth_unsubscribe"
+)
+
+func EthCoinbase(rCtx *types.RPCContext) (*types.RpcResult, error) {
```
zsystm, line 111: This should be removed after #456  is merged, since even Geth no longer supports this API. Other legacy APIs should also be removed — something we should keep in mind going forward.

aljo242: @zsystm good to remove now then?

tests/jsonrpc/simulator/go.mod
```diff
@@ -0,0 +1,48 @@
+module github.com/cosmos/evm/tests/jsonrpc/simulator
+
+go 1.23.8
+
+require (
+	github.com/ethereum/go-ethereum v1.15.10
```
aljo242, line 6: do we want to update the go-ethereum dep to the [latest](https://github.com/ethereum/go-ethereum/releases/tag/v1.16.2)

PR #436, opened 2025-08-12
refactor!: consolidate and clean up chain config

almk-dev:
# Description
> Remove static chain configs from EVM

> Currently we have the configuration for chains held in .go files that are compiled into the binary.  This logic and struct is duplicated throughout the codebase, making it very complicated to think about testing and spinning up nodes.

> Cosmos apps already come with an app.toml which is already used for some of our EVM configuration.  We should just move all of this logic to a config file so that it is simple  to work with in testing and for users

[STACK-1409](https://linear.app/cosmoslabs/issue/STACK-1409/remove-static-chain-configs-from-evm)
[EVM-176](https://linear.app/cosmoslabs/issue/EVM-176/evmd-should-support-flexible-chain-ids-for-external-test-suites)

Please review that all chain config redundancy has been removed and that all tests and functionality is working exactly the same as before. 

## Author Checklist
I have...


x/vm/types/denom_config.go
```diff
@@ -58,21 +58,33 @@ func setDisplayDenom(displayDenom string) error {
 // GetEVMCoinDecimals returns the decimals used in the representation of the EVM
 // coin.
 func GetEVMCoinDecimals() Decimals {
+	if evmCoinInfo == nil {
```
aljo242: why are we doing this with a global config?

.github/workflows/codeql-analysis.yml
```diff
@@ -18,7 +18,7 @@ permissions: read-all
 jobs:
   analyze:
     name: Analyze
-    runs-on: depot-ubuntu-24.04-8
+    runs-on: depot-ubuntu-24.04-4
```
aljo242, line 21: why did we need to downgrade?

evmd/ante/validate_handler_options_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build test
+// +build test
```
aljo242: why do we need these?

evmd/cmd/evmd/cmd/testnet.go
```diff
@@ -108,7 +108,7 @@ func addTestnetFlagsToCmd(cmd *cobra.Command) {
 	cmd.Flags().Int(flagNumValidators, 4, "Number of validators to initialize the testnet with")
 	cmd.Flags().StringP(flagOutputDir, "o", "./.testnets", "Directory to store initialization data for the testnet")
 	cmd.Flags().String(flags.FlagChainID, "", "genesis file chain-id, if left blank will be randomly created")
-	cmd.Flags().String(server.FlagMinGasPrices, fmt.Sprintf("0.000006%s", sdk.DefaultBondDenom), "Minimum gas prices to accept for transactions; All fees in a tx must meet this minimum (e.g. 0.01photino,0.001stake)")
+	cmd.Flags().String(server.FlagMinGasPrices, fmt.Sprintf("0.000006%s", TEST_DENOM), "Minimum gas prices to accept for transactions; All fees in a tx must meet this minimum (e.g. 0.01photino,0.001stake)")
```
aljo242, line 115: did we need to change this?

evmd/tests/ibc/ibc_middleware_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build test
+// +build test
```
aljo242: if we just include with a `_test` package name dont we not need this?

server/config/config.go
```diff
@@ -352,6 +377,39 @@ func (c TLSConfig) Validate() error {
 	return nil
 }
 
+// Validate returns an error if the chain configuration fields are invalid.
+func (c ChainConfig) Validate() error {
```
aljo242:
```suggestion
func (c *ChainConfig) Validate() error {
```

server/config/config.go
```diff
@@ -352,6 +377,39 @@ func (c TLSConfig) Validate() error {
...
+	}
+	if c.DisplayDenom == "" {
+		return errors.New("chain display-denom cannot be empty")
+	}
+	if c.Decimals == 0 || c.Decimals > 18 {
+		return errors.New("chain decimals must be between 1 and 18")
+	}
+
+	// For 18 decimals, denom and extended denom should be the same
+	if c.Decimals == 18 && c.Denom != c.ExtendedDenom {
```
aljo242: Is there a constant defined for 18 decimals that we use throughout our codebase? not sure tbh

server/config/config.go
```diff
@@ -352,6 +377,39 @@ func (c TLSConfig) Validate() error {
...
+		return errors.New("denom and extended-denom must be the same for 18 decimals")
+	}
+
+	return nil
+}
+
+// ToEvmCoinInfo converts ChainConfig to evmtypes.EvmCoinInfo
+func (c ChainConfig) ToEvmCoinInfo() evmtypes.EvmCoinInfo {
```
aljo242:
```suggestion
func (c *ChainConfig) ToEvmCoinInfo() evmtypes.EvmCoinInfo {
```

server/config/config.go
aljo242: Can we unit test this?

server/config/toml.go
```diff
@@ -113,4 +113,22 @@ certificate-path = "{{ .TLS.CertificatePath }}"
 
 # Key path defines the key.pem file path for the TLS configuration.
 key-path = "{{ .TLS.KeyPath }}"
+
+###############################################################################
+###                            Chain Configuration                          ###
+###############################################################################
```
aljo242: Say i had a chain on  v0.4.0.  What is the process of "migrating" my config to this new format?

tests/integration/x/precisebank/test_burn_integration.go
```diff
@@ -238,7 +238,15 @@ func (s *KeeperIntegrationTestSuite) TestBurnCoinsRemainder() {
...
 
 	// Burn 0.1 until balance is 0
+	// Add safety guard to prevent infinite loop
+	const maxIterations = 1000
+	iterationCount := 0
 	for {
+		iterationCount++
+		if iterationCount > maxIterations {
+			s.Failf("Test exceeded maximum iterations", "reached %d iterations without balance reaching zero", maxIterations)
+			break
```
aljo242: should we move this to another PR? what is this?

almk-dev (commented):
Clearing obsolete files marked as viewed in review progress.

evmd/ante/evm_antehandler_benchmark_test.go
```diff
@@ -1,3 +1,4 @@
+
 package ante_test
```
aljo242: weird spacing

evmd/tests/integration/ibc_test.go
```diff
@@ -1,3 +1,4 @@
+
 package integration
```
aljo242: we could make the diff a lot smaller by removing

testutil/config/dynamic_config_test.go
```diff
@@ -0,0 +1,132 @@
+//go:build test
+// +build test
```
aljo242: shouldn't need

config/evm_app_options.go
```diff
@@ -28,14 +34,16 @@ func EvmAppOptionsWithConfig(
...
 	}
 
-	if err := EvmAppOptionsWithConfigWithReset(chainID, chainsCoinInfo, cosmosEVMActivators, false); err != nil {
-		return err
+	coinInfo, found := chainsCoinInfo[chainID]
+	if !found {
+		return fmt.Errorf("unknown chain id: %d", chainID)
 	}
 
-	sealed = true
```
aljo242, line 35: do we no longer care about this `sealed` variable? what is its purpose?

evmd/cmd/evmd/config/config.go
```diff
@@ -1,36 +1,43 @@
...
-	EVMChainID: {
-		Denom:         "atest",
-		ExtendedDenom: "atest",
-		DisplayDenom:  "test",
-		Decimals:      evmtypes.EighteenDecimals,
-	},
+// GetEvmdEvmCoinInfo returns appropriate EvmCoinInfo for evmd based on chainID.
+// This replaces the old hardcoded ChainsCoinInfo map with a function that
+// creates configurations on demand.
+func GetEvmdEvmCoinInfo(chainID uint64) evmtypes.EvmCoinInfo {
```
aljo242: We still have these types of config utilities being exported from multiple places in the code.  One of the purposes of  this refactor should be that this is just in one unified place in our library if anyone needs to use some example config

config/evm_app_options.go
almkx: I believe it was added as a half-measure to solve chainID issues way in the past. It looks like it's acting like a simple lock (like a `sync.Once` but worse) so that further calls to chain configurations can't change it after initialization. It was likely to prevent different areas of the code initializing with different chain configs.

I'm testing with this flag removed and it seems to be stable without it, which is encouraging that we are no longer attempting to do configs in multiple places, but the actual full impact of removing is hard to pin down.

evmd/cmd/evmd/cmd/testnet.go
almkx: Yes, talked with Tyler about this and resolved this was likely set to `stake` by accident. It should use the `TEST_DENOM` const set above.

.github/workflows/codeql-analysis.yml
almkx: Tested time to completion with the higher core machine and it made no speedup in completion, so I'm setting it to 4, since it pins to 4 cpus anyways.

server/config/config.go
almkx: ChainConfig was removed in favor of using the exported CoinInfo

almkx: https://github.com/cosmos/evm/pull/436/commits/e85f0f07695c02f904ccb15038e9d0f76ad2884c#diff-a86ad716c06504885be192e4f8f48ce81c073d5a709fae87fb12c8484ae8df2f

tests/integration/x/precisebank/test_burn_integration.go
almkx: This was added as I was debugging the integration tests permanently hanging. Removed: https://github.com/cosmos/evm/pull/436/commits/1fda943cb3a5da031a2de618ff30dfaf123e4769

server/config/toml.go
almkx: All they would need to do if generate this config using the template in `toml.go` with their preexisting chain config (coin info), and it will take the place of this "standard" toml file. 

Not sure where to best document this.

.github/workflows/codeql-analysis.yml
aljo242: i never actually knew the -N was how many CPUs. good to know

evmd/cmd/evmd/config/evmd_config.go
```diff
@@ -92,6 +93,7 @@ type EVMAppConfig struct {
 	EVM     cosmosevmserverconfig.EVMConfig
 	JSONRPC cosmosevmserverconfig.JSONRPCConfig
 	TLS     cosmosevmserverconfig.TLSConfig
+	Chain   evmtypes.EvmCoinInfo
```
aljo242: This feels confusing that all of the other configs are defined in `cosmosevmserverconfig` but this one comes from another package (`evmtypes`).  Then below, the `DefaultEvmCoinInfo` function actually does come from the `cosmosevmserverconfig` package

evmd/cmd/evmd/config/evmd_config.go
```diff
@@ -117,14 +119,24 @@ func InitAppConfig(denom string, evmChainID uint64) (string, interface{}) {
 	evmCfg := cosmosevmserverconfig.DefaultEVMConfig()
 	evmCfg.EVMChainID = evmChainID
 
+	// Use the default chain configuration as a fallback
+	chainCfg := cosmosevmserverconfig.DefaultEvmCoinInfo()
```
aljo242: Isn't this just a `coinConfig` not really a `chainConfig`?

server/config/toml.go
aljo242: Folks generate this `toml` when they run `appd init` so this would be run at whatever initial version they are using.  So if i  was on `v0.4` when i ran `appd init` i would not have this section in my config.  What would happen in that scenario?

evmd/cmd/evmd/config/config.go
```diff
@@ -1,36 +1,41 @@
...
-	},
-	// EVMChainID provides a chain ID used for internal testing
-	EVMChainID: {
-		Denom:         "atest",
-		ExtendedDenom: "atest",
-		DisplayDenom:  "test",
-		Decimals:      evmtypes.EighteenDecimals,
-	},
+// GetEvmCoinInfo returns appropriate EvmCoinInfo for evmd based on chainID.
+func GetEvmCoinInfo(chainID uint64) evmtypes.EvmCoinInfo {
```
aljo242: Could we just export something like this once from the library instead of here and how it is duplicated in `testutil`?

server/config/migration/v0.50-app.toml
```diff
@@ -328,6 +328,24 @@ certificate-path = ""
 # Key path defines the key.pem file path for the TLS configuration.
 key-path = ""
 
+###############################################################################
```
aljo242: What is this file?

testutil/config/example_migration.md
```diff
@@ -0,0 +1,119 @@
+# Chain Configuration Migration Guide
```
aljo242: The title of this file is pretty generic.  How are we pointing people to this file?

Feels like it needs to be linked in the changelog or in an overall "migration guide" 

we have stuff [here](https://github.com/cosmos/evm/tree/main/docs/migrations) for previous releases

testutil/constants/constants.go
```diff
@@ -129,3 +97,48 @@ var (
 		},
 	}
 )
+
+// GetExampleChainCoinInfo returns appropriate EvmCoinInfo for testing based on chainID.
+// Deprecated: Use testutil/config.CreateEvmCoinInfoFromDynamicConfig() instead.
+// This function is kept for backward compatibility but should be replaced
+// with dynamic configuration generation.
```
aljo242: if we are deprecating this and keeping it for compatibility, then why not just keep the global map type?

Folks will be broken because we've changed the API

aljo242: would also make the diff much smaller

x/vm/types/config.go
```diff
@@ -11,6 +11,7 @@ import (
 	"fmt"
 
 	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/log"
```
aljo242: why does this file need to be behind `test` build flags?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - [\#467](https://github.com/cosmos/evm/pull/467) Replace GlobalEVMMempool by passing to JSONRPC on initiate.
 - [\#352](https://github.com/cosmos/evm/pull/352) Remove the creation of a Geth EVM instance, stateDB during the AnteHandler balance check.
 - [\#496](https://github.com/cosmos/evm/pull/496) Simplify mempool instantiation by using configs instead of objects.
+- [\#436](https://github.com/cosmos/evm/pull/436) Consolidate and clean up chain config
```
technicallyty: this feels a bit too ambiguous for the change being made. if i was an evm user, i wouldn't really know what needs to be actionized from this.

evmd/cmd/evmd/config/config.go
```diff
@@ -1,36 +1,41 @@
...
+		return evmtypes.EvmCoinInfo{
+			Denom:         "atest",
+			ExtendedDenom: "atest",
+			DisplayDenom:  "test",
+			Decimals:      evmtypes.EighteenDecimals,
+		}
+	default:
+		// Default fallback - return the default configuration
+		return *config.DefaultEvmCoinInfo()
+	}
```
technicallyty: this seems like it could be a potential footgun. what types of issues could cascade from this?

evmd/cmd/evmd/config/evmd_config.go
technicallyty: not sure about this field name

evmd/cmd/evmd/config/evmd_config.go
technicallyty: agree here, the whole EVMAppConfig is really what id think the chainCfg is.

server/config/toml.go
technicallyty: if we can automate the updating of their app.toml in an upgrade handler, we should definitely do that. im still going through the PR as i write this comment so i'm not sure what is possible yet.

testutil/config/dynamic_config.go
technicallyty, line 1: is this used for anything?

testutil/config/example_migration.md
technicallyty: we should definitely put what versions this is relevant for here

testutil/config/example_migration.md
```diff
@@ -0,0 +1,119 @@
...
+            ExtendedDenom: "atest",
+            DisplayDenom:  "test",
+            Decimals:      evmtypes.SixDecimals,
+        }
+    // ... more hardcoded cases
+    }
+}
+```
+
+## Dynamic Configuration for Tests
```
technicallyty: this seems a bit confusing to me. is there a reason we have all this dynamic stuff -- it seems like you could just skirt all this and directly construct an EvmCoinInfo type, no?

testutil/constants/constants.go
technicallyty: agree here, it seems counter intuitive to break something and immediately deprecate it. its also still used extensively throughout the code. should that be updated to the non-deprecated version?

testutil/integration/evm/network/config.go
```diff
@@ -118,14 +118,7 @@ type ConfigOption func(*Config)
 // WithChainID sets a custom chainID for the network. Changing the chainID
 // change automatically also the EVM coin used. It panics if the chainID is invalid.
 func WithChainID(chainID testconstants.ChainID) ConfigOption {
-	evmCoinInfo, found := testconstants.ExampleChainCoinInfo[chainID]
```
technicallyty, line 121: regarding the change of maps to functions for the coin info, the map version did have the benefit of giving the user a presence check for free. with the function as is, it doesnt seem like there is a way for the user to check that the thing they were looking for is actually there. but, im not too sure how important that is.

x/vm/types/config.go
```diff
@@ -46,7 +47,8 @@ func (ec *EVMConfigurator) Configure() error {
 }
 
 func (ec *EVMConfigurator) ResetTestConfig() {
-	panic("this is only implemented with the 'test' build flag. Make sure you're running your tests using the '-tags=test' flag.")
+	// no-op
+	log.Warn("this is only implemented with the 'test' build flag. Make sure you're running your tests using the '-tags=test' flag.")
```
technicallyty: why change to warn?

server/config/toml.go
Cordtus: I have thought a long time about how to "nudge" the user when their config files are outdated but I can't think of any very nice ways.
Missing parameters just go to the default setting AFAIK.

evmd/cmd/evmd/config/evmd_config.go
almkx: Agreed, I think `Coin` might be better here: https://github.com/cosmos/evm/pull/436/commits/cc8ccac8296de11739b1b91af1363db4d077f72d

x/vm/types/config.go
almkx: This is mainly for the CodeQL warning. It doesn't like any calls that have panics in them, as it considers that unsafe. The only way to get that pass is to mark that file with a `test` go tag, but that causes other problems and requires more changes and moving things around. 

I thought this might be a good compromise since in practice it would never use the non-test version of the method anyways.

evmd/cmd/evmd/config/evmd_config.go
almkx: https://github.com/cosmos/evm/pull/436/commits/cc8ccac8296de11739b1b91af1363db4d077f72d

x/vm/types/config.go
technicallyty: I've actually ran into this panic triggering when using the GoLand test runner (just clicking the ▶️ button next to tests). we should be able to make CodeQL skip analyzing the panic with a comment, similar to `//nolint` directives

evmd/cmd/evmd/config/evm_app_options.go
```diff
@@ -1,14 +1,26 @@
...
-func EvmAppOptions(chainID uint64) error {
-	return evmconfig.EvmAppOptionsWithConfig(chainID, ChainsCoinInfo, cosmosEVMActivators)
+// for the Cosmos EVM chain using the provided chain configuration.
+func EvmAppOptions(chainConfig ChainConfig) error {
+	return evmconfig.EvmAppOptionsWithConfig(chainConfig.ChainInfo.EVMChainID, chainConfig.CoinInfo, cosmosEVMActivators)
+}
+
+// EvmAppOptionsFromConfig creates an EVM options function that uses the provided chain configuration.
+func EvmAppOptionsFromConfig(chainConfig ChainConfig) evmconfig.EVMOptionsFn {
+	return func(chainID uint64) error {
```
cloudgray, line 16: Argument `chainID` is not used.

I think we should modify evmconfig.EVMOptionsFn as `func() error` type and remove this argument.

Currently, it is still confusing for devs in [app.goL271-L274](https://github.com/cosmos/evm/blob/stack-1409-clean-up-chain-config/evmd/app.go#L271-L274)

Because now, we are using EVMChainID from config, but looking at app.go, devs can confused that the other chainID input is used.

evmd/cmd/evmd/cmd/root.go
```diff
@@ -81,140 +79,288 @@ func NewRootCmd() *cobra.Command {
...
+
+// initSDKConfig initializes the SDK configuration
+func initSDKConfig() {
+	cfg := sdk.GetConfig()
+	evmdconfig.SetBip44CoinType(cfg)
+	cfg.Seal()
+}
+
+// createBasicApp creates a basic app instance for extracting encoding configuration
+func createBasicApp(nodeHome string) *evmd.EVMD {
```
cloudgray: Argument nodeHome is not used. 
How about removing it?

I think HomeDir is set in L82.

evmd/cmd/evmd/cmd/root.go
```diff
@@ -81,140 +79,288 @@ func NewRootCmd() *cobra.Command {
...
+	)
+	if err != nil {
+		return clientCtx, err
+	}
+	return clientCtx.WithTxConfig(txConfig), nil
+}
 
-	return cfg
+// loadChainConfigFromContext loads chain configuration from the command context and flags.
+func loadChainConfigFromContext(cmd *cobra.Command, clientCtx client.Context) (evmdconfig.ChainConfig, error) {
```
cloudgray: Argument clientCtx is not used.
How about removing it?

evmd/cmd/evmd/cmd/root.go
```diff
@@ -81,140 +79,288 @@ func NewRootCmd() *cobra.Command {
...
-		if err := evmdconfig.EvmAppOptions(evmdconfig.EVMChainID); err != nil {
-			panic(err)
+	return evmd.NewExampleApp(
+		log.NewNopLogger(),
+		dbm.NewMemDB(),
+		nil,
+		true,
+		simtestutil.EmptyAppOptions{},
+		1, // Temporary chain ID, only for codec extraction
+		noOpEvmAppOptions,
```
cloudgray: I have suggestions.

### Option 1. Lets make evmAppOptions as one of baseAppOptions.

Because evmAppOptions is used for global variable configuration and it returns nothing, I think we can add this function as one of baseAppOptions and let is executed in BaseApp instantiation.

```go
evmAppOptions := func(_ *baseapp.BaseApp) {
    if err := evmdconfig.EvmAppOptions(chainConfig); err != nil {
        panic(err)
    }
}
```

And we can remove `evmAppOptions config.EVMOptionsFn` argument from NewExampleApp method.

Then we don't have to make weird noOpAppOptions for cmd client instantiation.

In noremal case, we can input `evmAppOptions` as one of baseAppOptions like below.

```go
evmAppOptions := func(_ *baseapp.BaseApp) {
    if err := evmdconfig.EvmAppOptions(chainConfig); err != nil {
        panic(err)
    }
}

baseappOptions := []func(*baseapp.BaseApp){
    // ...
    evmAppOptions
}

app := evmd.NewExampleApp(
    //...
    baseappOptions
)
```

### Option 2. Let's just add condition in app.go

```go
// initialize the Cosmos EVM application configuration
if evmAppOptions != nil && err := evmAppOptions(evmChainID); err != nil {
    panic(err)
}
```

In this case, we can just set nil for evmAppOptions argument.

evmd/cmd/evmd/cmd/creator.go
```diff
@@ -88,14 +88,20 @@ func (a appCreator) newApp(
 		baseapp.SetIAVLCacheSize(cast.ToInt(appOpts.Get(server.FlagIAVLCacheSize))),
 	}
```
cloudgray, line 90: This is comment about `newApp` and `exportApp` functions.

These functions are not used.
Instead, `newApp` and `exportApp` in evmd/cmd/evmd/root.go are used.

The two set of functions are functionally identical.

I think we can remove this file entirely.

evmd/tests/integration/create_app.go
```diff
@@ -21,6 +20,15 @@ import (
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 )
 
+// testEvmAppOptionsWithReset creates an EvmAppOptions function for testing with reset
+func testEvmAppOptionsWithReset(evmChainID uint64) error {
+	// Use default chain configuration for testing
+	chainConfig := testconfig.DefaultChainConfig
+	configurator := evmtypes.NewEVMConfigurator()
+	configurator.ResetTestConfig()
+	return configurator.WithEVMCoinInfo(chainConfig.CoinInfo).Configure()
```
cloudgray, line 29: @almk-dev (cc. @vladjdk)
ChainConfig including chainID should be set, but currently, it is not set.
I think this is the reason why `invalid chainID 0` panic occurs in all integration tests.

On the other hand, start cmd works well. 
I guess It is because, chainConfig is configured well when evmd starts from cmd client.

You can refer to [how cmd client configure chainConfig](https://github.com/cosmos/evm/blob/stack-1409-clean-up-chain-config/config/app_options.go#L38-L67)

testutil/integration/evm/network/setup.go
```diff
@@ -463,12 +463,15 @@ func setDefaultMintGenesisState(cosmosEVMApp evm.EvmApp, genesisState cosmosevmt
 }
 
 func setDefaultErc20GenesisState(cosmosEVMApp evm.EvmApp, evmChainID uint64, genesisState cosmosevmtypes.GenesisState) cosmosevmtypes.GenesisState {
```
cloudgray, line 464: Argument evmChainID is not used. 
How about removing it?

x/vm/types/denom.go
```diff
@@ -77,13 +77,30 @@ func (d Decimals) ConversionFactor() math.Int {
...
-// EvmCoinInfo struct holds the name and decimals of the EVM denom. The EVM denom
-// is the token used to pay fees in the EVM.
-//
-// TODO: move to own file? at least rename file because it's unclear to use "denom"
-type EvmCoinInfo struct {
-	Denom         string
-	ExtendedDenom string
-	DisplayDenom  string
-	Decimals      Decimals
+func (d Decimals) GetDenomPrefix() string {
```
cloudgray: I think we should decide between 2 options and modify other parts for consistency.
1. Only support 1, 2, 3, 6, 9, 12, 15, 18 decimals
2. Support <= 18 decimals

For example, [Validate() method](https://github.com/cosmos/evm/blob/stack-1409-clean-up-chain-config/x/vm/types/denom.go#L59-L67) regards decimals less than or equal to 18 as valid.

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@
 - [\#568](https://github.com/cosmos/evm/pull/568) Avoid unnecessary block notifications when the event bus is already set up.
 - [\#511](https://github.com/cosmos/evm/pull/511) Minor code cleanup for `AddPrecompileFn`.
 - [\#544](https://github.com/cosmos/evm/pull/544) Parse logs from the txResult.Data and avoid emitting EVM events to cosmos-sdk events.
+- [\#436](https://github.com/cosmos/evm/pull/436) Refactor and consolidate chain config helpers and constants for test suites
+>>>>>>> main
```
cloudgray: lets remove this

tests/integration/ante/test_evm_unit_06_account_verification.go
```diff
@@ -27,13 +27,17 @@ import (
 func (s *EvmUnitAnteTestSuite) TestVerifyAccountBalance() {
 	// Setup
 	keyring := testkeyring.New(2)
+	chainConfig := testconfig.ChainConfig{
```
cloudgray, line 30: These setup Code are repeated in many test files.

I suggest adding `(s *EvmUnitAnteTestSuite) Setup()` method.

server/config/toml.go
```diff
@@ -113,4 +113,22 @@ certificate-path = "{{ .TLS.CertificatePath }}"
...
 # Key path defines the key.pem file path for the TLS configuration.
 key-path = "{{ .TLS.KeyPath }}"
+
+###############################################################################
+###                             Coin Configuration                          ###
+###############################################################################
+
+[coin]
+
+# Denom defines the base denomination used in the chain
```
cloudgray: I think we'd better adding detailed description.

```suggestion
# Denom is the base denomination of the native coin used in the chain’s EVM for paying gas fees. If the native coin has 18 decimals, then Denom must be the same denomination as ExtendedDenom.
```

evmd/test_helpers.go
```diff
@@ -44,8 +45,17 @@ func init() {
...
+	// Note: Bech32 prefixes are now managed by the configuration system
+}
+
+// testEvmAppOptions creates an EvmAppOptions function for testing
+func testEvmAppOptions(evmChainID uint64) error {
+	// Use default chain configuration for testing
+	chainConfig := testconfig.DefaultChainConfig
+	configurator := evmtypes.NewEVMConfigurator()
+	configurator.ResetTestConfig()
+	return configurator.WithEVMCoinInfo(chainConfig.CoinInfo).Configure()
```
cloudgray, line 58: ditto (https://github.com/cosmos/evm/pull/436#discussion_r2336056927)

evmd/tests/network/network.go
```diff
@@ -100,6 +100,15 @@ type Config struct {
 	PrintMnemonic     bool   // print the mnemonic of first validator as log output for testing
 }
 
+// testEvmAppOptions creates an EvmAppOptions function for testing
+func testEvmAppOptions(evmChainID uint64) error {
+	// Use default chain configuration for testing
+	chainConfig := testconfig.DefaultChainConfig
+	configurator := evmtypes.NewEVMConfigurator()
+	configurator.ResetTestConfig()
+	return configurator.WithEVMCoinInfo(chainConfig.CoinInfo).Configure()
```
cloudgray, line 109: ditto (https://github.com/cosmos/evm/pull/436#discussion_r2336056927)

ante/evm/mono_decorator_test.go
```diff
@@ -144,8 +146,16 @@ func toMsgSlice(msgs []*evmsdktypes.MsgEthereumTx) []sdk.Msg {
 }
 
 func TestMonoDecorator(t *testing.T) {
-	chainID := uint64(config.EighteenDecimalsChainID)
-	require.NoError(t, config.EvmAppOptions(chainID))
+	chainConfig := testconfig.DefaultChainConfig
```
cloudgray, line 149: [suggestion] I think we could commonize default setup code (L149-L156) in testutil/somewhere dir.

ante/evm/fee_checker_test.go
```diff
@@ -57,9 +56,15 @@ func TestSDKTxFeeChecker(t *testing.T) {
 	//      with extension option
 	//      without extension option
 	//      london hardfork enableness
-	chainID := uint64(config.EighteenDecimalsChainID)
+	chainConfig := testconfig.DefaultChainConfig
```
cloudgray, line 59: ditto (https://github.com/cosmos/evm/pull/436#discussion_r2336374525)

cloudgray (commented):
I left additional comments.

Overall, because the config values are declared and configured as global variables, I think code duplication occurs in various places and maintenance becomes difficult. This refactoring PR has already grown in size, so making additional large-scale changes here doesn’t seem appropriate. However, at some point, it would be better to separate the config and pass it as an argument individually to the app, gRPC server, and JSON-RPC server, etc...

x/vm/types/denom.go
almkx: I've updated the Validate method to make use of this. The <18 is too simplistic and has gaps in usability as many of those decimals are non standard and don't even have SI prefixes. 

My opinion is that we should strictly stick to 1, 2, 3, 6, 9, 12, 15, 18 decimals.

evmd/cmd/evmd/config/config.go
```diff
@@ -1,68 +1,24 @@
...
-	config.SetBech32PrefixForValidator(Bech32PrefixValAddr, Bech32PrefixValPub)
-	config.SetBech32PrefixForConsensusNode(Bech32PrefixConsAddr, Bech32PrefixConsPub)
+type ChainConfig struct {
+	ChainInfo ChainInfo
+	CoinInfo  evmtypes.EvmCoinInfo
+}
+
+type ChainInfo struct {
+	ChainID    string
+	EVMChainID uint64
```
Eric-Warehime, line 21: This is a bit confusing as we also have similarly named and purposed types in the vm module types.
```
// EvmConfig contains the specific EVM configuration details
type EvmConfig struct {
	ChainConfig  *ChainConfig
	CoinInfo     *EvmCoinInfo
	ExtendedEIPs map[int]func(*vm.JumpTable)
}
```

testutil/config/config.go
```diff
@@ -1,65 +1,39 @@
...
-	},
-	EVMChainID: {
-		Denom:         ExampleChainDenom,
-		ExtendedDenom: ExampleChainDenom,
-		DisplayDenom:  ExampleDisplayDenom,
-		Decimals:      evmtypes.EighteenDecimals,
-	},
+type ChainConfig struct {
+	ChainID   string
+	EvmConfig *evmtypes.EvmConfig
```
Eric-Warehime, line 9: Why do we need to redefine this here--we have similar types in evmd and in vm types.

evmd/cmd/evmd/config/loader.go
Eric-Warehime, line 1: I think this is directionally correct, but it could be improved if it was made a bit more generic to fit whatever params we want to pull from app options/flags. 

I think [dydx's setup](https://github.com/dydxprotocol/v4-chain/blob/main/protocol/app/flags/flags.go) is pretty clean. Maybe if we organized it similarly, it would be easier to track how parameters were initialized and passed into the app--and eliminate some of our duplicated fields tracking chain IDs, denom, decimals, etc. in multiple places.

x/vm/types/coin_info.go
```diff
@@ -0,0 +1,82 @@
...
+
+// EvmCoinInfo struct holds the display name and decimal precisions of the EVM denom
+type EvmCoinInfo struct {
+	// DisplayDenom defines the display denomination shown to users
+	DisplayDenom string `mapstructure:"display-denom"`
+	// Decimals defines the precision/decimals for the base denomination (1-18)
+	Decimals Decimals `mapstructure:"decimals"`
+	// ExtendedDecimals defines the precision/decimals for the extended denomination (typically 18 decimals for atto-denom)
+	ExtendedDecimals Decimals `mapstructure:"extended-decimals"`
+}
```
Eric-Warehime, line 23: I do consider these as things that should just be in module state--and potentially already are. Maybe you can pull things like Decimals or the Bond Denom from the bank/staking modules? Or maybe we should be storing these in precisebank and having an explicit dependency on that in the vm module?  


But it probably it makes sense to have these in module Params somewhere, and we can keep things like the eip-155 chain id and hardfork rules in user-supplied params.

x/vm/types/denom.go
cloudgray: I agree to use 1, 2, 3, 6, 9, 12, 15, 18 decimals.

But I cannot find changes of [Validate method and constants](https://github.com/cosmos/evm/blob/stack-1409-clean-up-chain-config/x/vm/types/denom.go#L13-L67)

PR #442, opened 2025-08-13
fix: avoid nil pointer by checking error in gov precompile FromResponse

mmsqe:
# Description

Closes: #XXXX

---

## Author Checklist

I have...


precompiles/gov/types.go
```diff
@@ -373,12 +373,12 @@ func ParseVotesArgs(method *abi.Method, args []interface{}) (*govv1.QueryVotesRe
 	}, nil
 }
 
-func (vo *VotesOutput) FromResponse(res *govv1.QueryVotesResponse) *VotesOutput {
+func (vo *VotesOutput) FromResponse(res *govv1.QueryVotesResponse) (*VotesOutput, error) {
```
aljo242, line 376: We should add test cases for these error cases

aljo242 (approved):

precompiles/gov/types.go
mmsqe: btw, the flaky gov test was resolved by https://github.com/cosmos/evm/pull/454 due to the random address might contains `val` which trigger the error

PR #467, opened 2025-08-14
chore: ensure SetGlobalEVMMempool is thread-safe and only set once

mmsqe:
keep singleton behavior on concurrent initialization

# Description

Closes: #XXXX

---

## Author Checklist

I have...


mempool/registry.go
```diff
@@ -1,18 +1,23 @@
...
 
 // SetGlobalEVMMempool sets the global ExperimentalEVMMempool instance.
-// This should only be called during application initialization.
-// In testing builds, it allows resetting by not returning an error.
-func SetGlobalEVMMempool(mempool *ExperimentalEVMMempool) error {
-	globalEVMMempool = mempool
-	return nil
+// This is guaranteed to only set the instance once for the lifetime of the process,
+// even if called multiple times. Should only be called during application initialization.
+func SetGlobalEVMMempool(mempool *ExperimentalEVMMempool) {
```
yihuang: there are still the risk that a temporary app is created after the main one.

mmsqe: not sure, do you mean [L19](https://github.com/MANTRA-Chain/mantrachain/blob/release/v5.0.0-rcx/cmd/mantrachaind/main.go#L19) is before L20?

yihuang: if it creates a new temp app anywhere, it'll override the mempool instance used by the main app.

mmsqe: why override if can set once?

yihuang: ok, then the problem is if the temp app is created before the main one, then we'll have the issue.

mmsqe: yea, just not sure how to trigger with 
```if err := svrcmd.Execute(cmd.NewRootCmd(), clienthelpers.EnvPrefix, app.DefaultNodeHome); err != nil {```

aljo242 (approved):
LGTM wdyt @vladjdk

PR #471, opened 2025-08-16
fix: avoid insufficient funds error when fund and send tx within same block

mmsqe:
# Description

* get SpendableCoin from sdk ctx to reflect uncommitted balance change
* for test [info](https://github.com/MANTRA-Chain/mantrachain/pull/412) about `failed to broadcast transaction: insufficient funds for gas * price + value`

Closes: #XXXX

---

## Author Checklist

I have...

### DIFF
--- a/mempool/txpool/legacypool/legacypool.go
+++ b/mempool/txpool/legacypool/legacypool.go
@@ -41,6 +41,7 @@ import (
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/holiman/uint256"
 
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/evm/mempool/txpool"
 )
 
@@ -133,6 +134,8 @@ type BlockChain interface {
 
 	// StateAt returns a state database for a given root hash (generally the head).
 	StateAt(root common.Hash) (vm.StateDB, error)
+
+	GetLatestCtx() (sdk.Context, error)
 }
 
 // Config are the configuration parameters of the transaction pool.
@@ -256,6 +259,8 @@ type LegacyPool struct {
 	changesSinceReorg int // A counter for how many drops we've performed in-between reorg.
 
 	BroadcastTxFn func(txs []*types.Transaction) error
+
+	SpendableCoin func(ctx sdk.Context, addr common.Address) *uint256.Int
 }
 
 type txpoolResetRequest struct {
@@ -601,6 +606,8 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction) error {
 			}
 			return nil
 		},
+		SpendableCoin: pool.SpendableCoin,
+		GetLatestCtx:  pool.chain.GetLatestCtx,
 	}
 	if err := txpool.ValidateTransactionWithState(tx, pool.signer, opts); err != nil {
 		return err
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/ethereum/go-ethereum/common"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/holiman/uint256"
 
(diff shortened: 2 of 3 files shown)


mempool/txpool/legacypool/legacypool.go
```diff
@@ -1570,7 +1588,7 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			log.Trace("Removed old pending transaction", "hash", hash)
 		}
 		// Drop all transactions that are too costly (low balance or out of gas), and queue any invalids back for later
-		drops, invalids := list.Filter(pool.currentState.GetBalance(addr), gasLimit)
+		drops, invalids := list.Filter(pool.SpendableCoin(ctx, addr), gasLimit)
```
vladjdk: Just doing a preemptive review here - as of the audit fixes, the [StateDB uses SpendableBalances](https://github.com/cosmos/evm/blob/b9bdbe6a335aeb05a4228c8bb24b992a1f84a82c/x/vm/keeper/statedb.go#L33) for its accounts and state objects, so we're fine using GetBalance here directly and keep a smaller diff.

mmsqe: yes, I felt weired to call keeper inside mempool.

mempool/README.md
```diff
@@ -211,7 +206,7 @@ ERROR unable to publish transaction nonce=40 expected=12: invalid sequence
 ERROR unable to publish transaction nonce=41 expected=12: invalid sequence
 ```
 
-**Real-World Testing**: The [`tests/systemtests/Counter/script/SimpleSends.s.sol`](../../tests/systemtests/Counter/script/SimpleSends.s.sol) script demonstrates typical Ethereum tooling behavior - it sends 10 sequential transactions in a batch, which naturally arrive out of order and create nonce gaps. With the default Cosmos mempool, this script would fail with sequence errors. With the EVM mempool, all transactions are queued locally and promoted as gaps are filled, allowing the script to succeed.
+**Real-World Testing**: The [`tests/systemtests/Counter/script/SimpleSends.s.sol`](../tests/systemtests/Counter/script/SimpleSends.s.sol) script demonstrates typical Ethereum tooling behavior - it sends 10 sequential transactions in a batch, which naturally arrive out of order and create nonce gaps. With the default Cosmos mempool, this script would fail with sequence errors. With the EVM mempool, all transactions are queued locally and promoted as gaps are filled, allowing the script to succeed.
```
vladjdk, line 226: Can we add the Event Bus to the integration guide?

vladjdk:
Let's also remove this check:

https://github.com/cosmos/evm/blob/7b8b9f5480c2133ea5ef7d9fd61e37ff5a462830/mempool/mempool.go#L206-L208

evmd/app.go
```diff
@@ -1145,6 +1151,40 @@ func (app *EVMD) SetClientCtx(clientCtx client.Context) {
 	app.clientCtx = clientCtx
 }
 
+// SetEventBus sets the application's CometBFT event bus to listen for new block header event.
+func (app *EVMD) SetEventBus(eventBus *cmttypes.EventBus) {
```
vladjdk: Could we add these functions into the mempool instead? I'm always wary of increasing the amount of EVM-specific logic that chains need to add into app.go.

I was thinking that we could add the eventBus as an object into the mempool and call `evmApp.Mempool().SetEventBus...` alongside some nil checks for the mempool.

Same with Close(), so that we can fully abstract this away from app integrations.

In this case, we won't even have to add anything into the integration guide.

mempool/README.md
vladjdk: Closing this in exchange for abstracting this logic into the mempool code instead.

evmd/app.go
mmsqe: We still need to add logic for unsubscribing eventbus inside app Close()

vladjdk: That's fine, let's move the logic for Close() into the mempool and call it from app then

PR #479, opened 2025-08-18
feat: allow PostTxProcessing to run on failures and persist data

facundomedica:
# Description

This change allows PostTxProcessing to be ran at all times (unless there's a previous error at the antehandler level of course), and to persist data even if the executed tx is reverted. The status in the receipt allows PostProcessingTx to know wether or not the tx has succeeded and act accordingly.

How it works? If the tx succeeds PostTxProcessing will use the same cached context as the tx, but if the tx fails we'll create a new cached context using the original one and use that for PostTxProcessing.

There are 3 new integration test cases:
 1. Happy path, tx succeeds, PostTxProcessing succeeds. The transfer occurs and we can see the balance change.
 2. Tx succeeds but PostTxProcessing errors. This will cause the entire tx to be reverted, so no balance change is expected.
 3. Tx fails but PostTxProcessing succeeds. This is the newly introduced case, in which the tx fails (because of it trying to over-spend), but the PostTxProcessing succeeds and is able to make state changes (in the test we mint a coin with a specific denom).

Note that this is only possible because of how cosmos/evm works with failing EVM transactions,  they fail only at the EVM level not at Cosmos level and we can still persist data.

Closes: #XXXX

---

## Author Checklist

I have...

### DIFF
--- a/x/vm/keeper/keeper.go
+++ b/x/vm/keeper/keeper.go
@@ -217,6 +217,11 @@ func (k *Keeper) PostTxProcessing(
 	return k.hooks.PostTxProcessing(ctx, sender, msg, receipt)
 }
 
+// HasHooks returns true if hooks are set
+func (k *Keeper) HasHooks() bool {
+	return k.hooks != nil
+}
+
 // ----------------------------------------------------------------------------
 // Log
 // ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


vladjdk (approved):
Thanks for the submission!

Approving this as a temporary fix. Ideally, we want to remove PostTxProcessing in the future, as it's fairly heavy and has seen little use. 

For the purpose of observability on EVM events, we should look at a longer-term fix. I opened an issue to track this in #482.

vladjdk:
@facundomedica check the unit tests and lints here

facundomedica:
@vladjdk I've added a check for hooks, we were always calling commitFn because PostTxProcessing would return a nil err when hooks aren't set. Now I only call PostTxProcessing if hooks are set, so we don't call commitFn for nothing. Tests and linter are passing.

BTW, we can remove the nil checks for commitFn, leaving it as is so we I don't add more noise to the PR.

Issue #254, opened 2025-06-30
Rename TokenPair to TokenMapping


vladjdk:
Pairs imply a swap between two different assets. In our case, the asset is the same, but with different representations, so a mapping more accurately describes the actual structure.

Ref: [https://github.com/cosmos/evm-internal/pull/32/files#r2176022095](https://github.com/cosmos/evm-internal/pull/32/files#r2176022095)

PR #480, opened 2025-08-18
fix: rename TokenPair to TokenMapping across repo

almk-dev:
# Description
> Pairs imply a swap between two different assets. In our case, the asset is the same, but with different representations, so a mapping more accurately describes the actual structure.

Closes: https://github.com/cosmos/evm/issues/254

Please review for comprehensive renaming (including go, proto, docs, etc.) and well as any functionality changes (there should not be).

## Author Checklist

I have...


almk-dev:
Closing as the changes are more breaking than expected. Aligned with @aljo242 that we will instead just add comments explaining the unclear `TokenPair` naming.

Issue #493, opened 2025-08-22
[Bug]: tx with future nonce after EIP-7702 tx get rejected with gapped nonce


mmsqe:
temporary rejection is not aligned with go-ethereum:
 
* submit an eip 7702 [transaction](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/3003b8cae1fd7fe79e9d05cd27b3859b649b200f/integration_tests/test_eip7702.py#L44) from an account with delegation or pending delegation
* submit a [transaction](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/3003b8cae1fd7fe79e9d05cd27b3859b649b200f/integration_tests/test_eip7702.py#L53) from the same account with a future (gapped) nonce
* 2nd transaction is rejected due to the nonce gap

PR #494, opened 2025-08-22
fix: align temporary reject check for included tx from txpool

mmsqe:
# Description

Closes: https://github.com/cosmos/evm/issues/493

---

## Author Checklist

I have...


rpc/backend/utils.go
```diff
@@ -291,3 +294,22 @@ func GetHexProofs(proof *crypto.ProofOps) []string {
...
+
+// HandleBroadcastRawLog checks the RawLog for known temporary rejection or nonce gap error.
+// NOTE: make sure it sync with the latest go-ethereum logic when upgrade:
+// https://github.com/ethereum/go-ethereum/blob/master/core/txpool/locals/errors.go#L29
+func HandleBroadcastRawLog(rawLog string, txHash common.Hash) (bool, string) {
+	if strings.Contains(rawLog, mempool.ErrNonceGap.Error()) {
+		return true, "transaction queued due to nonce gap"
+	}
+	switch rawLog {
+	case legacypool.ErrOutOfOrderTxFromDelegated.Error(),
```
vladjdk, line 307: Wouldn't these get outright rejected? I'm not sure I understand what temporary rejection means in this setting. The reason we treat nonce gaps differently is because they get added into the queue on the EVM side despite being rejected on the Cosmos side, so we don't want to return errors on those.

However, if a transaction is underpriced, there are nonce gaps on 7702 accounts, etc., these get discarded from the pool whatsoever, behaving differently from our nonce gap handling.

Could you reference the geth code that returns a hash on these and somewhere where we I read up on temporary rejections?

vladjdk: Looks like I just found it: https://github.com/cosmos/go-ethereum/blob/9089f9461cc2a25703ee247256e1d0c48b64f815/core/txpool/locals/errors.go#L29-L28

Could we use this function instead? Should improve maintainability

vladjdk (changes requested):

rpc/backend/sign_tx.go
```diff
@@ -106,15 +104,13 @@ func (b *Backend) SendTransaction(args evmtypes.TransactionArgs) (common.Hash, e
 	syncCtx := b.ClientCtx.WithBroadcastMode(flags.BroadcastSync)
 	rsp, err := syncCtx.BroadcastTx(txBytes)
 	if rsp != nil && rsp.Code != 0 {
+		if shouldSkip, msg := HandleBroadcastRawLog(rsp.RawLog, txHash); shouldSkip {
```
vladjdk, line 107: We don't have a local TX tracker here, so we should just be returning the error until we implement that: https://github.com/cosmos/go-ethereum/blob/f70aaa8399ccee429804eecf3fc4c6fd8d9e6cab/eth/api_backend.go#L325-L341

I believe returning the error without a tx tracker is still considered EVM compatible, and we can create a separate issue to add one in.

rpc/backend/utils.go
mmsqe: specified the [reason](https://github.com/cosmos/evm/pull/494/commits/d5f869dcc1ed5e07368fac357709166f48962d95)

aljo242:
@mmsqe looks like we have conflicts - thanks for the update

PR #495, opened 2025-08-22
fix: allow mempool interrupt while it is still ongoing

almk-dev:
# Description

Closes: #XXXX

---

## Author Checklist

I have...

### DIFF
--- a/mempool/txpool/txpool.go
+++ b/mempool/txpool/txpool.go
@@ -17,10 +17,12 @@
 package txpool
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"math/big"
 	"sync"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core"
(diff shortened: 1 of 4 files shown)


mempool/txpool/txpool.go
```diff
@@ -120,14 +129,42 @@ func New(gasTip uint64, chain BlockChain, subpools []SubPool) (*TxPool, error) {
...
+	if timeout == 0 {
+		// Force immediate shutdown without waiting
+		log.Info("Force shutting down transaction pool immediately")
+	} else {
+		// Try graceful shutdown with timeout
+		errc := make(chan error, 1)
+		go func() {
+			select {
+			case p.quit <- errc:
+				errc <- <-errc
```
mmsqe: why do we take the channel result and send back to the same channel here?

mempool/txpool/txpool.go
```diff
@@ -120,14 +129,42 @@ func New(gasTip uint64, chain BlockChain, subpools []SubPool) (*TxPool, error) {
...
+			case <-p.shutdownCtx.Done():
+				errc <- errors.New("shutdown context cancelled")
```
mmsqe: seems that this will always be triggered due to p.shutdownCancel call right?

vladjdk:
It still hangs for a few seconds after I run a catalyst task from this repo: https://github.com/skip-mev/catalyst/tree/technicallyty/STACK-1368-eth-txs.

Is it possible to make this as instant as it was pre-mempool?

server/start.go
```diff
@@ -132,8 +132,8 @@ which accepts a path for the resulting pprof file.
 				return err
 			}
 
-			withTM, _ := cmd.Flags().GetBool(srvflags.WithCometBFT)
```
almkx, line 135: This is a simple rename symbol I did while reading through this file.

Issue #460, opened 2025-08-14
Reorganize Configs for Mempool


vladjdk:
Instead of providing entire mempools as inputs into the mempool option, the configs for each mempool should be exposed. We don't recommend setting entirely different mempools.

PR #496, opened 2025-08-22
(chore):Simplify Mempool Config

vladjdk:
closes: #460 

Not a breaking change - uses configs instead of the actual objects during initialization, with sensible defaults still provided.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -67,11 +67,11 @@ type (
 // It allows customization of the underlying mempools, verification functions,
 // and broadcasting functions used by the sdkmempool.
 type EVMMempoolConfig struct {
-	TxPool        *txpool.TxPool
-	CosmosPool    sdkmempool.ExtMempool
-	AnteHandler   sdk.AnteHandler
-	BroadCastTxFn func(txs []*ethtypes.Transaction) error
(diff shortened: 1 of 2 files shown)


aljo242:
Can we add a changelog and make a follow up task for a guide for migrating your code?

mempool/mempool.go
```diff
@@ -67,11 +67,11 @@ type (
 // It allows customization of the underlying mempools, verification functions,
 // and broadcasting functions used by the sdkmempool.
 type EVMMempoolConfig struct {
-	TxPool        *txpool.TxPool
-	CosmosPool    sdkmempool.ExtMempool
-	AnteHandler   sdk.AnteHandler
-	BroadCastTxFn func(txs []*ethtypes.Transaction) error
-	BlockGasLimit uint64 // Block gas limit from consensus parameters
+	LegacyPoolConfig *legacypool.Config
+	CosmosPoolConfig *sdkmempool.PriorityNonceMempoolConfig[math.Int]
```
vladjdk, line 71: I kept this as a math.Int because I didn't feel like making this generic would be entirely useful since we usually rely on gas/fees for priorities. Could be wrong here though.

almk-dev (approved):
LGTM

aljo242 (approved):

PR #512, opened 2025-08-25
test(mempool): add integration test

cloudgray:
# Description

## Integration test added for appside mempool

- Fix test helper functions to generate valid transaction that has valid signature, gasPrice, gasLimit to pass CheckTx
- Call CheckTx for generated transactions and verify expected transactions are included in mempool in expected order.
- Call PrepareProposal and verify expected transactions are included in proposal in expected order.
- Modify existing unit test for mempool interface to use valid transaction instead of failing dummy transactions

### CheckTx reject valid tx for some case

When two transactions (`Tx1` and `Tx2`) with the same account and the same nonce but different GasPrice sequentially go through the CheckTx logic, replacement does not occur, and the later transaction is treated as an error.

- Expected behavior: `Tx2`, which has a higher GasPrice, should replace `Tx1`, which has a lower GasPrice, already in the mempool.
- Actual behavior: Only `Tx1` remains in the mempool.
- Error return point: [anteHandler](https://github.com/cloudgray/evm/blob/test%2Fappside-mempool/ante/evm/09_increment_sequence.go#L32-L35)

## Consideration for code review

- some test cases are bypassed for the found issue above "CheckTx reject valid tx for some case" issue. After fixing the issue, that test cases should be tested.

- `mempool.Remove()` can remove EVM transactoin only in case that tx is invalid After fixing tx gernerating helper functions to generate valida transaction, false positive test case that tries to remove tx from mempool was deleted.

Closes: #516

---

## Author Checklist

I have...


cloudgray:
@zsystm 
> In Geth’s mempool, when two txs with the same account/nonce are submitted, the one with the higher gas price replaces the earlier one, right?

Yes, more specifically, the following conditions apply, and these conditions are also applied in the same way to both geth([list.go:307-325](https://github.com/ethereum/go-ethereum/blob/dd1ebac11757484575ee779381af32f29ce3fbe4/core/txpool/legacypool/list.go#L307-L325)) and cosmos/evm ([list.go:307-325](https://github.com/ethereum/go-ethereum/blob/dd1ebac11757484575ee779381af32f29ce3fbe4/core/txpool/legacypool/list.go#L307-L325)).

1. **Price Bump Requirement**: The new transaction must have a gas price that's at least X% higher than the existing one (where X is the priceBump parameter, typically 10% by default)
2. **Both Fee Cap and Tip Must Be Higher**: For EIP-1559 transactions, both the GasFeeCap and GasTipCap must be higher than the existing transaction AND meet the percentage threshold

tests/integration/mempool/test_helper.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// checkTxs call abci CheckTx for a transaction
+func (s *IntegrationTestSuite) checkTx(tx sdk.Tx) error {
+	txBytes, err := s.network.App.GetTxConfig().TxEncoder()(tx)
+	if err != nil {
+		return fmt.Errorf("failed to encode cosmos tx: %w", err)
+	}
+
+	_, err = s.network.App.CheckTx(&abci.RequestCheckTx{
```
vladjdk: Let's make sure that no changes are actually persisted to the context (i.e. sequence incrementing) after a simple CheckTx is done. I a normal environment, changes would actually save until the finalization anyways.

tests/integration/mempool/test_mempool_integration_abci.go
```diff
@@ -0,0 +1,453 @@
...
+				// Input txs in order
+				inputTxs := []sdk.Tx{mediumFeeTx, lowFeeTx, highFeeTx}
+
+				// Expected txs in order
+				expectedTxs := []sdk.Tx{highFeeTx}
+				expTxHashes := s.getTxHashes(expectedTxs)
+
+				return inputTxs, expTxHashes
+			},
+			bypass: true,
```
vladjdk, line 101: Reverting the increment from `checkTxs` should allow the test to pass.

tests/integration/mempool/test_mempool_integration_abci.go
```diff
@@ -0,0 +1,453 @@
...
+				expTxHashes := s.getTxHashes(expectedTxs)
+
+				return txs, expTxHashes
+			},
+			verifyFunc: func(mpool mempool.Mempool) {
+				// After replacement, both nonces 0 and 1 should be pending
+				count := mpool.CountTx()
+				s.Require().Equal(2, count, "After replacement, both transactions should be pending")
+			},
+			bypass: true,
```
vladjdk, line 399: ditto

tests/integration/mempool/test_mempool_integration_abci.go
```diff
@@ -0,0 +1,453 @@
...
+				// Input txs in order
+				inputTxs := []sdk.Tx{cosmosTx, evmTx}
+
+				// Expected txs in order
+				expectedTxs := []sdk.Tx{evmTx, cosmosTx}
+				expTxHashes := s.getTxHashes(expectedTxs)
+
+				return inputTxs, expTxHashes
+			},
+			bypass: true,
```
vladjdk, line 122: ditto

tests/integration/mempool/test_mempool_integration_abci.go
```diff
@@ -0,0 +1,453 @@
...
```
vladjdk, line 142: ditto

tests/integration/mempool/test_mempool_integration_abci.go
```diff
@@ -0,0 +1,453 @@
...
+				// Input txs in order
+				inputTxs := []sdk.Tx{evmTx, cosmosTx}
+
+				// Expected txs in order
+				expectedTxs := []sdk.Tx{cosmosTx, evmTx}
+				expTxHashes := s.getTxHashes(expectedTxs)
+
+				return inputTxs, expTxHashes
+			},
+			bypass: true,
```
vladjdk, line 162: ditto

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - [\#467](https://github.com/cosmos/evm/pull/467) Replace GlobalEVMMempool by passing to JSONRPC on initiate.
 - [\#352](https://github.com/cosmos/evm/pull/352) Remove the creation of a Geth EVM instance, stateDB during the AnteHandler balance check.
 - [\#496](https://github.com/cosmos/evm/pull/496) Simplify mempool instantiation by using configs instead of objects.
+- [\#512](https://github.com/cosmos/evm/pull/512) Add integration test for apside mempool
```
aljo242:
```suggestion
- [\#512](https://github.com/cosmos/evm/pull/512) Add integration test for appside mempool.
```

aljo242 (approved):

Issue #505, opened 2025-08-24
Problem: dynamic erc20 precompiles are heavy


yihuang:
## Proposal

Implement an unified bank precompile to manage all native denoms:

```solidity
interface IBankPrecompile {
    function name(string memory denom) external view returns (string memory);
    function symbol(string memory denom) external view returns (string memory);
    function decimals(string memory denom) external view returns (uint8);
    function totalSupply(string memory denom) external view returns (uint256);
    function balanceOf(address account, string memory denom) external view returns (uint256);

    function transferFrom(address from, address to, uint256 value, string memory denom) external returns (bool);
}
```

The only mutable function is `transferFrom`, it only authorize two use cases:
1. `msg.sender == from`
    user are allowed to transfer their own funds for any denom.

2. `msg.sender` is the deterministic contract constructed from the standard implementation with the specific denom.
    This contract is allowed to manage any user's funds only for the specific denom.

## Standard ERC20 Implementation

```solidity
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.0;

interface IERC20 {
    event Transfer(address indexed from, address indexed to, uint256 value);
    event Approval(address indexed owner, address indexed spender, uint256 value);
    function totalSupply() external view returns (uint256);
    function balanceOf(address account) external view returns (uint256);
    function transfer(address to, uint256 value) external returns (bool);
    function allowance(address owner, address spender) external view returns (uint256);
    function approve(address spender, uint256 value) external returns (bool);
    function transferFrom(address from, address to, uint256 value) external returns (bool);
}

interface IERC20Metadata is IERC20 {
    function name() external view returns (string memory);
    function symbol() external view returns (string memory);
    function decimals() external view returns (uint8);
}

interface IERC20Errors {
    error ERC20InvalidSender(address sender);
    error ERC20InvalidReceiver(address receiver);
    error ERC20InsufficientAllowance(address spender, uint256 allowance, uint256 needed);
    error ERC20InvalidApprover(address approver);
    error ERC20InvalidSpender(address spender);
}

interface IBankPrecompile {
    function name(string memory denom) external view returns (string memory);
    function symbol(string memory denom) external view returns (string memory);
    function decimals(string memory denom) external view returns (uint8);
    function totalSupply(string memory denom) external view returns (uint256);
    function balanceOf(address account, string memory denom) external view returns (uint256);

    function transferFrom(address from, address to, uint256 value, string memory denom) external returns (bool);
}

[...]

PR #517, opened 2025-08-26
Problem: dynamic ERC20 precompiles are heavy

yihuang:
# Description

Closes: #505

Re-design bank precompile to avoid dynamic precompiles.

## Notes

* Single byte method index, instead of 4bytes hash based selector.
* Packed ABI format.
* Manual encode/decode code instead of `github.com/ethereum/go-ethereum/accounts/abi` library.

Closes: #XXXX

---

## Author Checklist

I have...


aljo242:
We need some more time to review this.

But my initial thoughts are that I like the `init` pattern for loading the precompiles.  That seems like it should be its own separate PR that we can review first

yihuang:
> We need some more time to review this.
> 
> But my initial thoughts are that I like the `init` pattern for loading the precompiles. That seems like it should be its own separate PR that we can review first

those things are another PR: https://github.com/cosmos/evm/pull/577

this PR should only be merged after that one.

Issue #537, opened 2025-08-27
Optimize Gas Estimation


vladjdk:
Currently, our binary search estimation in EstimateGas runs the same transaction at least 20 times. Refer to [https://github.com/ethereum/go-ethereum/blob/v1.16.2/eth/gasestimator/gasestimator.go](https://github.com/ethereum/go-ethereum/blob/v1.16.2/eth/gasestimator/gasestimator.go) for optimizations made to shorten the amount of times we estimate.

In particular, [short-circuit on plain sends to non-contracts](https://github.com/ethereum/go-ethereum/blob/dd1ebac11757484575ee779381af32f29ce3fbe4/eth/gasestimator/gasestimator.go#L117), and calculate an [optimisticGasLimit](https://github.com/ethereum/go-ethereum/blob/dd1ebac11757484575ee779381af32f29ce3fbe4/eth/gasestimator/gasestimator.go#L147).

PR #538, opened 2025-08-27
perf: optimize gas estimation

vladjdk:
Optimizes gas estimation by short-circuiting on simple sends and calculating an optimistic gas limit based on the first transaction's gas used.

closes: #537

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - [\#492](https://github.com/cosmos/evm/pull/492) Duplicate case switch to avoid empty execution block
 - [\#509](https://github.com/cosmos/evm/pull/509) Allow value with slashes when query token_pairs
 - [\#495](https://github.com/cosmos/evm/pull/495) Allow immediate SIGINT interrupt when mempool is not empty
+- [\#545](https://github.com/cosmos/evm/pull/545) Check if mempool is not nil before accepting nonce gap error tx.
 
 ### IMPROVEMENTS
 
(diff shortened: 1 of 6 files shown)


x/vm/keeper/grpc_query.go
```diff
@@ -427,10 +439,34 @@ func (k Keeper) EstimateGasInternal(c context.Context, req *types.EthCallRequest
...
 			return nil, fmt.Errorf("gas required exceeds allowance (%d)", gasCap)
 		}
+		// If no larger allowance is available, fail fast
+		return nil, fmt.Errorf("gas required exceeds allowance (%d)", hi)
 	}
+
+	// There's a fairly high chance for the transaction to execute successfully
+	// with gasLimit set to the first execution's usedGas + gasRefund. Explicitly
+	// check that gas amount and use as a limit for the binary search.
+	optimisticGasLimit := (result.MaxUsedGas + ethparams.CallStipend) * 64 / 63
```
technicallyty, line 451: in geth, before setting the optimisticGasLimit, they set the value of `lo` using the result above. should that be added here?

https://github.com/ethereum/go-ethereum/blob/79807bc3b16ee1dadb506c87917fcf042d4e186d/eth/gasestimator/gasestimator.go#L137-L142

Eric-Warehime (approved):

x/vm/keeper/grpc_query.go
aljo242: @vladjdk seems like a good addition to make right?

aljo242: created follow up issue

Issue #563, opened 2025-08-28
Investigate the removal of GetBlockchain.NotifyNewBlock() in EndBlocker


vladjdk:
We no longer rely on EndBlock to do this, and we instead use a subscription. Remove this from the EndBlocker in x/vm.

[https://github.com/cosmos/evm/blob/f164197bce4f2b2b69b50141019be801ebfe40d9/x/vm/keeper/abci.go#L46](https://github.com/cosmos/evm/blob/f164197bce4f2b2b69b50141019be801ebfe40d9/x/vm/keeper/abci.go#L46)

PR #568, opened 2025-08-29
feat: avoid unnecessary block notifications when the event bus is already set up

mmsqe:
this change should have been included in https://github.com/cosmos/evm/commit/642c8e989405b8604b61001c61cfaa6928a35ec1

# Description

Closes: https://github.com/cosmos/evm/issues/563

---

## Author Checklist

I have...


almk-dev (approved):
Thanks!

testutil/integration/evm/network/abci.go
```diff
@@ -49,6 +51,9 @@ func (n *IntegrationNetwork) finalizeBlockAndCommit(duration time.Duration, txBy
 	if err != nil {
 		return nil, err
 	}
+	if evmMempool, ok := n.app.GetMempool().(*evmmempool.ExperimentalEVMMempool); ok {
+		evmMempool.GetBlockchain().NotifyNewBlock()
+	}
```
aljo242: why do we have to add this call here?

mmsqe: need notify for unit test since it doesn't run a node to fire event

PR #580, opened 2025-09-02
test(systemtests): add appside mempool e2e test

cloudgray:
# Description

Closes: #XXXX

---

## Author Checklist

I have...


mempool/mempool.go
```diff
@@ -386,7 +386,7 @@ func (m *ExperimentalEVMMempool) SelectBy(goCtx context.Context, i [][]byte, f f
 
 // SetEventBus sets CometBFT event bus to listen for new block header event.
 func (m *ExperimentalEVMMempool) SetEventBus(eventBus *cmttypes.EventBus) {
-	if m.HasEventBus() {
+	if m.eventBus != nil {
```
aljo242: why did we make this change? seems out of scope for this

cloudgray: Oh, It seems that when I merge conflict from main branch, something wrong happens.
I'll rollback that changes.

cloudgray: fixed! (https://github.com/cosmos/evm/pull/580/commits/d0892134be98d4c173a7fbc5f8c0fc846f75a92f)

tests/systemtests/clients/types.go
```diff
@@ -0,0 +1,53 @@
...
+	AccountNumber uint64
+	PrivKey       *ethsecp256k1.PrivKey
+}
+
+type TxPoolResult struct {
+	Pending map[string]map[string]*RPCTransaction `json:"pending"`
+	Queued  map[string]map[string]*RPCTransaction `json:"queued"`
+}
+
+type RPCTransaction struct {
```
vladjdk: We have this type [here](https://github.com/cosmos/evm/blob/4d93f2d04225b48c38427bfcbc8378d3f8b49092/rpc/types/types.go#L36-L60). Let's use that instead of copying.

tests/systemtests/clients/types.go
```diff
@@ -0,0 +1,53 @@
...
+	PrivKey *ecdsa.PrivateKey
+}
+
+type CosmosAccount struct {
+	AccAddress    sdk.AccAddress
+	AccountNumber uint64
+	PrivKey       *ethsecp256k1.PrivKey
+}
+
+type TxPoolResult struct {
```
vladjdk, line 22: Ditto, same file as the RPCTransaction.

tests/systemtests/go.mod
```diff
@@ -1,11 +1,17 @@
-module github.com/evmos/tests/systemtests
+module github.com/cosmos/evm/tests/systemtests
 
 go 1.24.4
 
 require (
+	cosmossdk.io/math v1.5.3
 	cosmossdk.io/systemtests v1.3.0
-	github.com/ethereum/go-ethereum v1.15.5
+	github.com/cometbft/cometbft/v2 v2.0.0-rc1
```
vladjdk, line 8: Why are we using V2 and RCs here?

tests/systemtests/mempool/interface.go
```diff
@@ -0,0 +1,42 @@
...
+	"testing"
+	"time"
+
+	"github.com/cosmos/evm/tests/systemtests/suite"
+)
+
+type TestSuite interface {
+	// Test Lifecycle
+	BeforeEach(t *testing.T)
+	AfterEach(t *testing.T)
```
vladjdk:
```suggestion
	AfterEachCase(t *testing.T)
```

tests/systemtests/mempool/interface.go
```diff
@@ -0,0 +1,42 @@
...
+	JustAfterEach(t *testing.T)
```
vladjdk:
```suggestion
	AfterEachAction(t *testing.T)
```

tests/systemtests/mempool/interface.go
```diff
@@ -0,0 +1,42 @@
...
+
+	// Tx
+	SendTx(t *testing.T, nodeID string, accID string, nonceIdx uint64, gasPrice *big.Int, gasTipCap *big.Int) (*suite.TxInfo, error)
+	SendEthTx(t *testing.T, nodeID string, accID string, nonceIdx uint64, gasPrice *big.Int, gasTipCap *big.Int) (*suite.TxInfo, error)
```
vladjdk, line 19: How does this differ from the DynamicFeeTx?

tests/systemtests/suite/test_suite.go
```diff
@@ -0,0 +1,150 @@
...
+	require.NoError(t, err)
+
+	s.baseFee = currentBaseFee
+}
+
+// JustAfterEach checks the expected mempool state right after each test case
+func (s *SystemTestSuite) JustAfterEach(t *testing.T) {
+	// Check pending txs exist in mempool or already committed
+	for _, txInfo := range s.GetExpPendingTxs() {
+		err := s.CheckPendingOrCommitted(txInfo.DstNodeID, txInfo.TxHash, txInfo.TxType, time.Second*10)
```
vladjdk: Can we actually slow block times down to check for pending, but **not committed** transactions? The way that these are sent around is actually fundamentally different, and I'm more curious about the former behaviour.

Committed transactions happen via [consensus round gossip](https://github.com/cometbft/cometbft/blob/f981aba653f39b44330bacadf72cf82881e955bf/consensus/reactor.go#L440-L446) and are [sent via built blocks](https://github.com/cometbft/cometbft/blob/848dd9fb4a0596f1b20430315491bbcb24cd8185/consensus/state.go#L1236-L1254).

Unconfirmed transcations are propagated differently, via the [mempool reactor](https://github.com/cometbft/cometbft/blob/f205d8496923fc8543036f4d8da57c3a45042cef/mempool/reactor.go#L183-L258).

We can assume that block propagation functions properly, so we only really care about the pending transaction gossip via the mempool reactor. That is exactly what we should test for. We could probably slow down blocktimes enough to test for the uncommitted, but still pending transactions.

tests/systemtests/suite/test_suite.go
```diff
@@ -0,0 +1,150 @@
...
+
+		for i := range s.Nodes() {
+			pendingTxHashes := pendingHashes[i]
+			queuedTxHashes := queuedHashes[i]
+
+			if s.Node(i) == txInfo.DstNodeID {
+				ok := slices.Contains(queuedTxHashes, txInfo.TxHash)
+				require.True(t, ok, fmt.Sprintf("tx %s is not contained in queued txs in mempool", txInfo.TxHash))
+			} else {
+				ok := slices.Contains(pendingTxHashes, txInfo.TxHash)
```
vladjdk: I like this logic :). This is exactly what I wanted to test for.

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+
+					_, err = s.SendTx(t, s.Node(2), "acc0", 2, s.BaseFee(), nil)
+					require.NoError(t, err, "failed to send tx")
+					tx6, err := s.SendTx(t, s.Node(2), "acc0", 2, s.BaseFeeX2(), big.NewInt(1))
+					require.NoError(t, err, "failed to send tx")
+
+					s.SetExpPendingTxs(tx2, tx4, tx6)
+				},
+			},
+			bypass: true,
```
vladjdk, line 54: Why are we bypassing these?

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+				func(s TestSuite) {
+					_, err := s.SendTx(t, s.Node(0), "acc0", 0, s.BaseFee(), nil)
+					require.NoError(t, err, "failed to send tx")
+					tx2, err := s.SendTx(t, s.Node(0), "acc0", 0, s.BaseFeeX2(), big.NewInt(1))
+					require.NoError(t, err, "failed to send tx")
+
+					s.SetExpPendingTxs(tx2)
+				},
+			},
+			bypass: true,
```
vladjdk, line 30: Ditto

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+				func(s TestSuite) {
+					_, err := s.SendEthTx(t, s.Node(0), "acc0", 0, s.BaseFee(), nil)
+					require.NoError(t, err, "failed to send tx")
+					tx2, err := s.SendCosmosTx(t, s.Node(1), "acc0", 0, s.BaseFeeX2(), big.NewInt(1))
+					require.NoError(t, err, "failed to send tx")
+
+					s.SetExpPendingTxs(tx2)
+				},
+			},
+			bypass: true,
```
vladjdk, line 167: Ditto on all of these. If this is from https://github.com/cosmos/evm/pull/591, let's merge that in first and un-bypass these tests.

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+		},
+		{
+			name: "single pending tx (high prio evm tx first) %s",
```
vladjdk, line 170: Couldn't there theoretically be a race condition here where Node(1) is the proposer and commits the transaction, making Node(0)'s transaction technically invalid?

My recurring thought is that we should slow down blocktimes to prevent this from happening.

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+			},
+			bypass: true,
+		},
+		{
+			name: "single pending tx (low prio cosmos tx first) %s",
+			actions: []func(s TestSuite){
+				func(s TestSuite) {
+					_, err := s.SendCosmosTx(t, s.Node(0), "acc0", 0, s.BaseFee(), nil)
+					require.NoError(t, err, "failed to send tx")
+					tx2, err := s.SendEthTx(t, s.Node(1), "acc0", 0, s.BaseFeeX2(), big.NewInt(1))
```
vladjdk: Ditto

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+			},
+			bypass: true,
+		},
+		{
+			name: "single pending tx (high prio cosmos tx first) %s",
+			actions: []func(s TestSuite){
+				func(s TestSuite) {
+					tx1, err := s.SendCosmosTx(t, s.Node(0), "acc0", 0, s.BaseFeeX2(), big.NewInt(1))
+					require.NoError(t, err, "failed to send tx")
+					_, err = s.SendEthTx(t, s.Node(1), "acc0", 0, s.BaseFee(), nil)
```
vladjdk: Ditto

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+		actions []func(s TestSuite)
+		bypass  bool
+	}{
+		{
+			name: "single pending tx (low prio evm tx first) %s",
+			actions: []func(s TestSuite){
+				func(s TestSuite) {
+					_, err := s.SendEthTx(t, s.Node(0), "acc0", 0, s.BaseFee(), nil)
+					require.NoError(t, err, "failed to send tx")
+					tx2, err := s.SendCosmosTx(t, s.Node(1), "acc0", 0, s.BaseFeeX2(), big.NewInt(1))
```
vladjdk: Ditto

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+			},
+			bypass: true,
+		},
+		{
+			name: "single pending tx (high prio evm tx first) %s",
+			actions: []func(s TestSuite){
+				func(s TestSuite) {
+					tx1, err := s.SendEthTx(t, s.Node(0), "acc0", 0, s.BaseFeeX2(), big.NewInt(1))
+					require.NoError(t, err, "failed to send tx")
+					_, err = s.SendCosmosTx(t, s.Node(1), "acc0", 0, s.BaseFee(), nil)
```
vladjdk: Ditto

tests/systemtests/mempool/test_replacement.go
```diff
@@ -0,0 +1,425 @@
...
+					txHash, err := s.SendEthLegacyTx(t, s.Node(0), s.Acc(0), 0, s.BaseFee())
+					require.NoError(t, err, "failed to send tx")
+
+					s.SetExpPendingTxs(txHash)
+					s.PromoteExpTxs(1)
+				},
+			},
+		},
+		{
+			name: "legacy should always not tx replace dynamic fee tx",
```
vladjdk:
```suggestion
			name: "legacy should never replace dynamic fee tx",
```

tests/systemtests/suite/query.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if result.TxResult.Code != 0 {
+		return fmt.Errorf("tx(%s) is committed but failed: %v", result.Hash.String(), err)
+	}
+
+	return nil
+}
+
+// CheckPendingOrCommitted checks if the given tx is either pending or committed within the timeout duration
+func (s *SystemTestSuite) CheckPendingOrCommitted(
```
vladjdk: Let's do `CheckPending` and `CheckCommitted` as two separate calls as they function differently in Comet.

tests/systemtests/mempool/interface.go
cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/e566f7a4460811f0a067f5557f6b75f2c9f0a037)

tests/systemtests/mempool/interface.go
cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/e566f7a4460811f0a067f5557f6b75f2c9f0a037)

tests/systemtests/mempool/interface.go
cloudgray: It calls SendEthLegacyTx or SendEthDynamicFeeTx according to testOptions. [[ref](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/suite/tx.go#L37-L41)]

It is for using same test cases with various test options (e.g. EthLegacyTx, EthDynamicFeeTx). [[ref](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/mempool/test_replacement.go#L108-L131)]
I wanted to prevent test code becoming too long.

tests/systemtests/suite/test_suite.go
cloudgray: Agree. I'll try that!

tests/systemtests/suite/test_suite.go
cloudgray: Thanks for your detailed guide before starting this PR. 😀

tests/systemtests/mempool/test_replacement.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2348936711)

tests/systemtests/mempool/test_replacement.go
cloudgray: Exactly. Ok, then I'll handle comments of https://github.com/cosmos/evm/pull/591 first.

tests/systemtests/mempool/test_replacement.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2348936711)

tests/systemtests/mempool/test_replacement.go
cloudgray: Oh, you are right.
I think I missed this because this and [other similar 3 test cases](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/mempool/test_replacement.go#L169-L210) are bypassed.

I remember that this is the reason why I write [these four test cases](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/mempool/test_replacement.go#L211-L293) with nonce-gapped txs.

I'll fix.

tests/systemtests/mempool/test_replacement.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2348959553)

tests/systemtests/mempool/test_replacement.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2348959553)

tests/systemtests/mempool/test_replacement.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2348959553)

tests/systemtests/mempool/test_replacement.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2348959553)

cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/58e97d8583e5cd06603b31506b88c7ff9a53ac62)

tests/systemtests/mempool/test_replacement.go
cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/58e97d8583e5cd06603b31506b88c7ff9a53ac62)

tests/systemtests/mempool/test_replacement.go
cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/58e97d8583e5cd06603b31506b88c7ff9a53ac62)

tests/systemtests/mempool/test_replacement.go
cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/58e97d8583e5cd06603b31506b88c7ff9a53ac62)

tests/systemtests/mempool/test_replacement.go
cloudgray: applied (https://github.com/cosmos/evm/pull/580/commits/58e97d8583e5cd06603b31506b88c7ff9a53ac62)

I made test cases send tx to same node

tests/systemtests/go.mod
cloudgray: It is because [cosmossdk.io/systemtests v1.3.0 is dependent with cometbft v2.0.0-rc-1.](https://github.com/cosmos/cosmos-sdk/blob/d6c085d451c29a63046e06ee055cdc21bd34cd20/go.mod#L22)

Actually, I tried to downgrade systemtests to [v1.2.1 that is dependent with cometbft v0.38.17](https://github.com/cosmos/cosmos-sdk/blob/d1dfc5176609a21d6d50542f59020c55c5ce5e31/go.mod#L21).

But I got issue.
I remember that some flags currently being used do not work.

tests/systemtests/clients/types.go
cloudgray: For this, maybe this PR (#600) should be merged first.

You can see this line ([go.mod](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/go.mod#L192))

This is temporary fix for resolving Issue #601.
But after temporary fix, when I refer to some cosmos/evm package (to refer to `rpctypes.RPCTransaction`), I faced build error with dependency problem like below.

```
# github.com/cosmos/evm/x/vm/types
../../x/vm/types/opcodes_hooks.go:11:5: undefined: vm.OpCodeHooks
../../x/vm/types/config.go:37:15: undefined: vm.ExtendActivators
../../x/vm/types/logs.go:122:23: log.BlockTimestamp undefined (type *"github.com/ethereum/go-ethereum/core/types".Log has no field or method BlockTimestamp)
FAIL    github.com/cosmos/evm/tests/systemtests [build failed]
FAIL    github.com/cosmos/evm/tests/systemtests/clients [build failed]
FAIL    github.com/cosmos/evm/tests/systemtests/mempool [build failed]
FAIL    github.com/cosmos/evm/tests/systemtests/suite [build failed]
```

Maybe this is because of temporary replacing cosmos/evm package.

So, we need to merge PR #600 and remove replacement of cosmos/evm pacakges.

tests/systemtests/clients/types.go
cloudgray: ditto (https://github.com/cosmos/evm/pull/580#discussion_r2345092004)

tests/systemtests/suite/query.go
cloudgray: Agree!, I'll handle this comment tomorrow!

tests/systemtests/mempool/test_replacement.go
cloudgray: applied! (https://github.com/cosmos/evm/pull/580/commits/58e97d8583e5cd06603b31506b88c7ff9a53ac62)

tests/systemtests/suite/query.go
cloudgray: I just removed checking for committed tx from `CheckPendingOrCommitted` and made checking operation concurrent logic. (https://github.com/cosmos/evm/pull/580/commits/23d361d485c89965a55e69e47845884e963e0522)

It is because, lets say there are `tx1`, `tx2`, `tx3`,  for some test cases, 3 txs are committed in different block.
But for the other cases, tx1 and tx2 can be committed in one block. and tx3 in the other. In this case, if I check pending txs sequentially, we succeed to check tx1 but fail to check tx2 because tx2 can be already committed.

To solve this situation, I modified this verification method as concurrent function.

FYI, we don't have to add `CheckCommitted`. Checking commit status is already conducted in AfterEachCase. ([ref](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/suite/test_suite.go#L93-L97))

tests/systemtests/suite/test_suite.go
cloudgray: applied! 
ditto(https://github.com/cosmos/evm/pull/580#discussion_r2355138587)

tests/systemtests/suite/query.go
vladjdk: Very nice, checked your changes. Thank you!

vladjdk:
Merged #591 and #601. Let's fix the type import and disable bypassing. Rest lgtm!

tests/systemtests/clients/types.go
cloudgray: @vladjdk 

## Summary 

I’ve reconsidered, and I think it makes more sense to define and use `RPCTransaction` as a separate type with only the `Hash` field.

```go
type EthRPCTransaction struct{
	Hash common.Hash `json:"hash"`
}
```

## Detail
It would be better to manage things modularly so that the `cosmos/evm/tests/systemtests` package does not depend on the `cosmos/evm` package. Since `cosmos/evm/tests/systemtests` uses a separately built evmd binary, I believe this is the proper way to run `systemtests` together with `evmd`.

As you mentioned, if `RPCTransaction` is referenced from the cosmos/evm/rpc package, then even after updating cosmos/evm to the latest version, there will still be build errors because cosmos/evm and systemtests would be referencing different versions of cometbft. This kind of issue could easily recur whenever new test code is added.

Currently, the only part of the `cosmos/evm/tests/systemtests` package that references cosmos/evm is the use of `ethsecp256k1` in the Cosmos client for signing, which I added. ([ref1](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/clients/types.go#L7), [ref2](https://github.com/cloudgray/evm/blob/teest%2Fappside-mempool-e2e/tests/systemtests/clients/cosmosclient.go#L13))


## Proposed Solution
Therefore, my proposed solution is as follows:

Redefine `RPCTransaction` locally in `cosmos/evm/tests/systemtests`, keeping only the actually used field, `RPCTransaction.Hash`.

Reimplement the ethsecp256k1 package specifically for the Cosmos client, thereby completely removing the dependency on cosmos/evm.

WDYT?

tests/systemtests/mempool/test_replacement.go
cloudgray: @vladjdk  I separated issue! (https://github.com/cosmos/evm/issues/634)

Issue #449, opened 2025-08-13
Add MinimumGasPrices and BlockGas to Mempool from configs


vladjdk:
Get app.toml and client.toml data into app.go so that they can be used in the mempool.

Refs: 

[https://github.com/cosmos/evm/pull/387#discussion_r2272186869](https://github.com/cosmos/evm/pull/387#discussion_r2272186869)<br>[https://github.com/cosmos/evm/pull/387#discussion_r2267063628](https://github.com/cosmos/evm/pull/387#discussion_r2267063628)

PR #582, opened 2025-09-02
feat: add block max_gas and min_tip ingestion to app.go mempool config

almk-dev:
# Description

> Get app.toml and client.toml data into app.go so that they can be used in the mempool.

Instead of setting `minTip` in the mempool config to `minimum-gas-price` directly, we decided to have a separate configurable field instead, called `min_tip`.  

Closes: [#449](https://github.com/cosmos/evm/issues/449)

Review whether this matches the original intent regarding usefulness for the app-side mempool implementation. Ensure there are no regressions in accessing genesis config attributes before blockchain initialization. 

---

## Author Checklist
I have...


aljo242:
@almk-dev needs a changelog

evmd/app.go
```diff
@@ -765,10 +767,21 @@ func NewExampleApp(
...
+		// Get the block gas limit from genesis file
+		blockGasLimit := getBlockGasLimitFromGenesis(appOpts, logger)
+
+		// Get MinGasPrices from app configuration
+		minGasPricesStr := cast.ToString(appOpts.Get(sdkserver.FlagMinGasPrices))
+		minGasPrices, err := sdk.ParseDecCoins(minGasPricesStr)
+		if err != nil {
+			logger.With("error", err).Info("failed to parse min gas prices, using empty DecCoins")
+			minGasPrices = sdk.DecCoins{}
+		}
```
aljo242: All of this logic is in the `emvd` package.  Could we move some portion of it to the `evm` library and export from there?

evmd/app.go
```diff
@@ -1207,3 +1220,50 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
...
+
+	logger.Debug(
+		"extracted block gas limit from genesis using SDK AppGenesisFromFile",
+		"genesis_path", genesisPath,
+		"max_gas", maxGas,
+		"block_gas_limit", blockGasLimit,
+	)
+
+	return blockGasLimit
+}
```
aljo242: can we test this?

evmd/app.go
almkx: https://github.com/cosmos/evm/pull/582/commits/4205566599f351840e404bcfffecdc0379e802a8

evmd/app.go
almkx: https://github.com/cosmos/evm/pull/582/commits/841bded5725dbd783caff8794a52524b80c3170a

aljo242 (approved):

mempool/mempool.go
```diff
@@ -447,8 +450,14 @@ func (m *ExperimentalEVMMempool) getIterators(goCtx context.Context, i [][]byte)
 
 	m.logger.Debug("getting iterators")
 
+	var minTip *uint256.Int
+	gasPrice := m.minGasPrices.AmountOf(m.evmDenom)
+	if !gasPrice.IsZero() {
+		minTip = uint256.MustFromBig(gasPrice.BigInt())
```
almkx: After a bunch of testing, I determined that the root cause of the stalled solidity tests is in the `uint256.MustFromBig()` call, which results in an overflow and panic. 

Interestingly, the behavior of this seems to persist between multiple runs of `make test-solidity`, depending on what `local_node` was run with. For example, if `./local_node.sh -y` was run with single line removed, then `make test-solidity` will always pass, regardless of whether this line was re-introduced.

This implies that `./local_node.sh -y` is persisting some state some where (it doesn't seem to be any different in `.evmd/config/*`), that is used by the solidity tests.

mempool/mempool.go
```diff
@@ -447,8 +450,14 @@ func (m *ExperimentalEVMMempool) getIterators(goCtx context.Context, i [][]byte)
...
+	}
+
 	pendingFilter := txpool.PendingFilter{
-		MinTip:       nil,
+		MinTip:       minTip,
```
Eric-Warehime: The reason the tests are now failing is that with this change you're actually requiring a priority fee per gas to be supplied for every transaction in order for it to be included in a block. 
This means that our solidity test transactions for creation of contracts just sit in the mempool and never get included.

I know that the miner in geth has some logic which matches this with the min gas, but it's not a protocol rule, and in reth it's defaulted to 0 as well https://github.com/paradigmxyz/reth/blob/36e39ebe3dd5199c8f11a705ae8c0a2bd534ede2/crates/transaction-pool/src/config.rs#L108

I think we can add this as a field to the mempool config and just default it to 0  instead of min gas prices. That should make all of the tests pass too.

Eric-Warehime (changes requested):

mempool/mempool.go
almkx: Makes sense! Thanks.

Do you have any idea what causes the "stickiness" with the local node behavior?

almkx: I don't think we actually need to discard setting the minTip in code. The previous local_node flag was broken: https://github.com/cosmos/evm/pull/582/commits/732bb3ba4c027d646672062a5676f72ce3a7b7d0

mempool/mempool.go
almkx: 😓 This was the critical change: https://github.com/cosmos/evm/pull/582/commits/732bb3ba4c027d646672062a5676f72ce3a7b7d0#diff-98ae6d1cbaf65ca1fce0086341cdaab6bc96800bf2632d8de91cd6b18c7d140eR361

mempool/mempool.go
almkx: https://github.com/cosmos/evm/pull/582/commits/732bb3ba4c027d646672062a5676f72ce3a7b7d0

I think this is a better solution (don't know why that argument was set like that in the first place). This way we don't have different defaults for mempool and the behavior is more predictable.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - [\#568](https://github.com/cosmos/evm/pull/568) Avoid unnecessary block notifications when the event bus is already set up.
 - [\#511](https://github.com/cosmos/evm/pull/511) Minor code cleanup for `AddPrecompileFn`.
 - [\#544](https://github.com/cosmos/evm/pull/544) Parse logs from the txResult.Data and avoid emitting EVM events to cosmos-sdk events.
+- [\#582](https://github.com/cosmos/evm/pull/582) Add block max_gas and minimum_gas_prices ingestion from toml/json
```
aljo242: I don't understand what the changelog entry means?  I think mentioning that they are from flags  / configs is clearer and to spell out what the flags and specific config files / fields are

Eric-Warehime (approved):
Ty!

config/server_app_options.go
```diff
@@ -0,0 +1,92 @@
...
+		minGasPrices = sdk.DecCoins{}
+	}
+
+	return minGasPrices
+}
+
+// GetMinTip reads the min tip from the app options, set from app.toml
+// This field is also known as the minimum priority fee
+func GetMinTip(appOpts servertypes.AppOptions, logger log.Logger) *uint256.Int {
+	flag := "min-tip"
```
aljo242: should this be a constant?

config/server_app_options.go
```diff
@@ -0,0 +1,92 @@
...
+	minTipUint64 := cast.ToUint64(appOpts.Get(flag))
+	minTip := uint256.NewInt(minTipUint64)
+
+	if minTip.Cmp(uint256.NewInt(0)) >= 0 { // zero or positive
+		return minTip
+	}
+
+	return nil
```
aljo242, line 93: so if this is negative (or incorrectly configured) we silently fail by returning a `nil`?

evmd/app.go
```diff
@@ -84,7 +85,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/codec/types"
 	"github.com/cosmos/cosmos-sdk/runtime"
 	runtimeservices "github.com/cosmos/cosmos-sdk/runtime/services"
-	"github.com/cosmos/cosmos-sdk/server"
+	sdkserver "github.com/cosmos/cosmos-sdk/server"
```
aljo242, line 88: why did we make this change?

evmd/app.go
```diff
@@ -765,10 +766,15 @@ func NewExampleApp(
...
+		blockGasLimit := evmconfig.GetBlockGasLimit(appOpts, logger)
+		// Get GetMinTip from app.toml or cli flag configuration
+		mipTip := evmconfig.GetMinTip(appOpts, logger)
+
 		mempoolConfig := &evmmempool.EVMMempoolConfig{
 			AnteHandler:   app.GetAnteHandler(),
-			BlockGasLimit: 100_000_000,
+			BlockGasLimit: blockGasLimit,
+			MinTip:        mipTip,
 		}
```
aljo242, line 778: feels like this whole function could be factored out

mempool/mempool.go
```diff
@@ -29,7 +29,10 @@ import (
 
 var _ sdkmempool.ExtMempool = &ExperimentalEVMMempool{}
 
-const SubscriberName = "evm"
+const (
+	SubscriberName        = "evm"
+	fallbackBlockGasLimit = 100_000_000
```
aljo242, line 36: can we doc this? when do we fall back to this value?

CHANGELOG.md
almkx: @aljo242 This should be clearer: https://github.com/cosmos/evm/pull/582/commits/9457bfa69d4692e664f6fb5bbb5cc35261623e5e

evmd/app.go
almkx: To distinguish with the in-repo `server` package, which is also used frequently and makes things hard to read imo

config/server_app_options.go
almkx: Good catch, this was a placeholder while I added the constants to `srvflags` and I missed repointing it to the new vars. Fixed: https://github.com/cosmos/evm/pull/582/commits/77bb5a250e86a723755a910c200d315ce1f980b2

config/server_app_options.go
almkx: Hm, yeah we should log it but fall back to nil since that matches the default behavior. Fixed: https://github.com/cosmos/evm/pull/582/commits/77bb5a250e86a723755a910c200d315ce1f980b2

mempool/mempool.go
almkx: Added docs and also updated the log to a `warn`: https://github.com/cosmos/evm/pull/582/commits/48f26f37b230813540b1c627db8882ce12ea5e46

That link also points to where we use the fallback (in the `NewExperimentalEVMMempool()` constructor).

evmd/app.go
almkx: I agree, but it's likely not worth doing as it will be rewritten by the `app.go` refactor. We can revisit when this PR's changes are merged into that branch.

aljo242: i would at least make an issue to track

almkx: Done: https://linear.app/cosmoslabs/issue/APPS-9/refactor-out-mempool-configuration-in-appgo-into-separate-helper

server/config/config.go
```diff
@@ -66,6 +66,9 @@ const (
 	// DefaultEVMChainID is the default EVM Chain ID if one is not provided
 	DefaultEVMChainID = 262144
 
+	// DefaultEVMMinTip is the default minimum priority fee for the mempool
```
Cordtus, line 69: Should this be up in the mempool section?

almkx: Since't it's an EVM prefix, I think it is easier to find with the other EVM prefex vars here.

almk-dev:
Relevant docs:
https://cosmos-docs-pre-050.mintlify.app/docs/next/documentation/getting-started/node-configuration

PR #588, opened 2025-09-09
feat: bridge go-ethereum metrics to prometheus

technicallyty:
# Description

---

## Author Checklist

I have...


metrics/geth_prom_bridge.go
Eric-Warehime, line 1: The bridge code unfortunately looks brittle--there's probably a lot of changes in prom and geth that could cause issues with it in the future. Doesn't seem like there's a better way of translating from one type to the other though. 

Do you think it would be a better approach to just go into the txpool that was forked into this repo and replace all of the geth metrics with SDK-compatible metrics instead of trying to translate everything?

technicallyty: id say if you find the idea of all of geths metrics being automatically joined to the SDK's prom instance, we should stick with the bridge. 

if we just want to quickly get txpool's metrics and nothing else, i can just swap them to use the SDK's telemetry package. 

and thetn theres still the other alternative of using geth's own prometheus package and just putting all its metrics out via its own server/port

technicallyty: fwiw im a fan of the last option - it touches the least amount of code, and seems like it wouldnt really cause much pain. i think we just tell overseer to scrape another port and it should just work?

technicallyty:
TODO: scrap bridge. put metrics on new port.

Eric-Warehime (approved):

PR #587, opened 2025-09-04
feat: implement object stores

mmsqe:
# Description

Closes: #XXXX

---

## Author Checklist

**All** items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.

I have...

- [ ] tackled an existing issue or discussed with a team member
- [ ] left instructions on how to review the changes
- [ ] targeted the `main` branch

PR #589, opened 2025-09-09
feat: Block STM

Eric-Warehime:
# Description

Draft for integrating block stm.

Closes: #XXXX

---

## Author Checklist

I have...


x/vm/store/snapshotkv/store.go
```diff
@@ -5,24 +5,23 @@ import (
...
 // Store manages a stack of nested cache store to
 // support the evm `StateDB`'s `Snapshot` and `RevertToSnapshot` methods.
 type Store struct {
 	// Store of the initial state before transaction execution
-	initialStore storetypes.CacheKVStore
+	initialStore storetypes.CacheWrap
 
 	// Stack of cached store
-	cacheStores []storetypes.CacheKVStore
+	cacheStores []storetypes.CacheWrap
```
mmsqe, line 18: actually I felt a bit hacky like https://github.com/cosmos/evm/pull/587/commits/04f5ae28058294cee9cb968b472b0b794bc6d1df when doing cast for snapshotKVStore

x/vm/keeper/keeper.go
```diff
@@ -126,14 +134,23 @@ func NewKeeper(
...
+		storeKeys:        storeKeys,
 	}
 }
 
+func (k *Keeper) EnableVirtualFeeCollection() {
+	/*
+		if types.GetEVMCoinDecimals() != types.EighteenDecimals {
+			panic("virtual fee collection is only enabled for 18 decimal fee denoms")
+			}
+	*/
```
Eric-Warehime: Need to fix this--should prevent using virtual fee collection if you don't have an 18 decimal gas token. i.e. if you are using the precisebank keeper.

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@
 - [\#733](https://github.com/cosmos/evm/pull/733) Avoid rejecting tx with unsupported extension option for ExtensionOptionDynamicFeeTx.
 - [\#736](https://github.com/cosmos/evm/pull/736) Add InitEvmCoinInfo upgrade to avoid panic when denom is not registered.
 
+* [\#658](https://github.com/cosmos/evm/pull/658) Fix race condition between legacypool's RemoveTx and runReorg.
+
```
Eric-Warehime: This is not correct.

evmd/app.go
```diff
@@ -35,6 +37,8 @@ import (
 	feemarkettypes "github.com/cosmos/evm/x/feemarket/types"
 	ibccallbackskeeper "github.com/cosmos/evm/x/ibc/callbacks/keeper"
 
+	// NOTE: override ICS20 keeper to support IBC transfers of ERC20 tokens
+
```
Eric-Warehime: Probably remove this

ante/cosmos.go
```diff
@@ -40,6 +40,5 @@ func newCosmosAnteHandler(ctx sdk.Context, options HandlerOptions) sdk.AnteHandl
 		ante.NewSigVerificationDecorator(options.AccountKeeper, options.SignModeHandler),
 		ante.NewIncrementSequenceDecorator(options.AccountKeeper),
 		ibcante.NewRedundantRelayDecorator(options.IBCKeeper),
-		evmante.NewGasWantedDecorator(options.EvmKeeper, options.FeeMarketKeeper, &feemarketParams),
```
technicallyty, line 43: why is this removed?

ante/evm/01_setup_ctx.go
```diff
@@ -50,12 +41,5 @@ func SetupContextAndResetTransientGas(ctx sdktypes.Context, tx sdktypes.Tx, evmK
 	newCtx := evmante.BuildEvmExecutionCtx(ctx).
 		WithGasMeter(storetypes.NewInfiniteGasMeter())
 
-	// Reset transient gas used to prepare the execution of current cosmos tx.
-	// Transient gas-used is necessary to sum the gas-used of cosmos tx, when it contains multiple eth msgs.
-	//
-	// TODO: add more context here to explain why gas used is reset. Not clear
-	// from docstring.
-	evmKeeper.ResetTransientGasUsed(ctx)
```
technicallyty, line 58: same question

ante/evm/11_emit_event.go
```diff
@@ -3,62 +3,23 @@ package evm
...
-// TxIndex should be included in the header vote extension as part of ABCI++
-func EmitTxHashEvent(ctx sdk.Context, msg *evmtypes.MsgEthereumTx, blockTxIndex, msgIndex uint64) {
+func EmitTxHashEvent(ctx sdk.Context, msg *evmtypes.MsgEthereumTx, blockTxIndex uint64) {
 	// emit ethereum tx hash as an event so that it can be indexed by CometBFT for query purposes
 	// it's emitted in ante handler, so we can query failed transaction (out of block gas limit).
 	ctx.EventManager().EmitEvent(
 		sdk.NewEvent(
 			evmtypes.EventTypeEthereumTx,
 			sdk.NewAttribute(evmtypes.AttributeKeyEthereumTxHash, msg.Hash().String()),
-			sdk.NewAttribute(evmtypes.AttributeKeyTxIndex, strconv.FormatUint(blockTxIndex+msgIndex, 10)), // #nosec G115
```
technicallyty, line 61: what was the point of the msgIndex? does this break anyone who might be relying on this event?

evmd/go.mod
```diff
@@ -297,4 +297,6 @@ replace (
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.9.1
 	// replace broken goleveldb
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
+	// use btree fork for bstm
+	github.com/tidwall/btree => github.com/cosmos/btree v0.0.0-20250924232609-2c6195d95951
```
technicallyty, line 301: nit: do we actually need this replace? could we just use the cosmos/btree everywhere?

go.mod
```diff
@@ -290,6 +290,8 @@ replace (
 	github.com/gin-gonic/gin => github.com/gin-gonic/gin v1.9.1
 	// replace broken goleveldb
 	github.com/syndtr/goleveldb => github.com/syndtr/goleveldb v1.0.1-0.20210819022825-2ae1ddf74ef7
+	// use btree fork for bstm
+	github.com/tidwall/btree => github.com/cosmos/btree v0.0.0-20250924232609-2c6195d95951
```
technicallyty, line 294: ditto

x/vm/keeper/keeper.go
```diff
@@ -130,14 +134,21 @@ func NewKeeper(
...
 		storeKeys:        storeKeys,
 	}
 }
 
+// EnableVirtualFeeCollection switches fee deduction for evm transactions to use the virtual fee collection of the
+// bank keeper via the object store.
+// Note: Do NOT use this if your chain does not have an 18 decimal point precision gas token.
+func (k *Keeper) EnableVirtualFeeCollection() {
+	k.virtualFeeCollection = true
+}
```
technicallyty, line 150: what happens if you don't have an 18 decimal gas token? does BSTM just not work? do we have tests for either scenario?

x/vm/keeper/state_transition.go
```diff
@@ -223,25 +211,25 @@ func (k *Keeper) ApplyTransaction(ctx sdk.Context, tx *ethtypes.Transaction) (*t
...
-		BlockNumber:       big.NewInt(ctx.BlockHeight()),
-		TransactionIndex:  txConfig.TxIndex,
+		Type:            tx.Type(),
+		PostState:       nil,
+		Logs:            ethLogs,
+		TxHash:          txConfig.TxHash,
+		ContractAddress: contractAddr,
+		GasUsed:         res.GasUsed,
+		BlockHash:       common.BytesToHash(ctx.HeaderHash()),
+		BlockNumber:     big.NewInt(ctx.BlockHeight()),
```
technicallyty: is CumulativeGasUsed set elsewhere? does this break anyone?

evmd/go.mod
Eric-Warehime: Yeah maybe we can just use the cosmos package directly now in the SDK? Probably some tech debt that we should figure out. I'll make a ticket for this.

x/vm/keeper/keeper.go
Eric-Warehime: The virtual fee collection does not integrate with precisebank to track fractional balances.
I originally had a panic in here that would be triggered based on the evm token denom, but that gets a NPE when called during app creation, so I'm not sure where to put this check. I've just added docstrings for now.

x/vm/keeper/state_transition.go
Eric-Warehime: Good call, I think the main use case for all of these fields in the receipt is to send them to the post tx processing hooks. I've added back in all of the fields except for the bloom data which isn't updated until the end of the block anymore so intermediate changes to the block bloom wouldn't change things and the correct value will be calculated based on the included logs.

x/vm/keeper/keeper_test.go
```diff
@@ -41,8 +41,10 @@ func TestKeeperTestSuite(t *testing.T) {
 
 func (suite *KeeperTestSuite) SetupTest() {
 	key := storetypes.NewKVStoreKey(vmtypes.StoreKey)
-	transientKey := storetypes.NewTransientStoreKey(vmtypes.TransientKey)
-	testCtx := testutil.DefaultContextWithDB(suite.T(), key, storetypes.NewTransientStoreKey("transient_test"))
+	okey := storetypes.NewObjectStoreKey(vmtypes.ObjectKey)
```
aljo242:
```suggestion
	oKey := storetypes.NewObjectStoreKey(vmtypes.ObjectKey)
```

evmd/app.go
```diff
@@ -268,7 +270,7 @@ func NewExampleApp(
 		txConfig:          txConfig,
 		interfaceRegistry: interfaceRegistry,
 		keys:              keys,
-		tkeys:             tkeys,
+		okeys:             okeys,
```
aljo242:
```suggestion
		oKeys:             oKeys,
```

evmd/app.go
```diff
@@ -162,9 +161,9 @@ type EVMD struct {
 	pendingTxListeners []evmante.PendingTxListener
 
 	// keys to access the substores
-	keys    map[string]*storetypes.KVStoreKey
-	tkeys   map[string]*storetypes.TransientStoreKey
-	memKeys map[string]*storetypes.MemoryStoreKey
+	keys  map[string]*storetypes.KVStoreKey
+	tkeys map[string]*storetypes.TransientStoreKey
```
aljo242: do we need these anymore?

x/vm/keeper/keeper.go
technicallyty: so this essentially means BSTM does not work on your chain unless you have an 18 dec token?

Eric-Warehime: This version doesn't integrate with precisebank, which means mismatches between evm token and cosmos token for bank precompiles will not properly calculate fractional balances.

Issue #590, opened 2025-09-09
Fix(mempool): CheckTxHandler should handle tx with "invalid sequence" error 


cloudgray:
## Problem

The AnteHandler rejects a valid transaction whose nonce is higher than the committed state’s nonce but lower than the nonce of a transaction already present in the mempool.

## Problem Details

In Ethereum’s mempool, a transaction (`TX1`) with the same nonce as an existing transaction (`TX0`) is allowed.
If `TX1` has a higher priority, it replaces `TX0`.

However, in cosmos/evm, the AnteHandler rejects `TX1`.
As a result, the cosmos/evm mempool cannot handle transactions the same way Ethereum’s mempool does.

[ante/evm/09_increment_sequence.go](https://github.com/cosmos/evm/blob/67950ad934f3393cf6a507dc1b5ad2aa46edaa60/ante/evm/09_increment_sequence.go#L25-L36)

```go
	if txNonce != accountNonce {
		if txNonce > accountNonce {
			return errorsmod.Wrapf(
				mempool.ErrNonceGap,
				"tx nonce: %d, account accountNonce: %d", txNonce, accountNonce,
			)
		}
		return errorsmod.Wrapf(
			errortypes.ErrInvalidSequence,
			"invalid nonce; got %d, expected %d", txNonce, accountNonce,
		)
	}

```

While `ErrNonceGap` is handled by CheckTxHandler and that tx is inserted into mempool as "queued transaction", ErrInvalidSequence is not handled appropriately.

However, there is already validation logic that checks whether tx nonce is lower than nonce with committed state, not uncommitted state. ([mempool/txppool/validation.go](https://github.com/cosmos/evm/blob/67950ad934f3393cf6a507dc1b5ad2aa46edaa60/mempool/txpool/validation.go#L223-L226))
```go
	next := opts.State.GetNonce(from)
	if next > tx.Nonce() {
		return fmt.Errorf("%w: next nonce %v, tx nonce %v", core.ErrNonceTooLow, next, tx.Nonce())
	}
```

This validation is called when new tx is added to mempool. ([mempool/txpool/legacypool/legacypool.go](https://github.com/cosmos/evm/blob/67950ad934f3393cf6a507dc1b5ad2aa46edaa60/mempool/txpool/legacypool/legacypool.go#L688-L692))

## Proposed Solution

[...]

PR #591, opened 2025-09-09
fix(mempool): checkTxHandler handles "invalid nonce" tx

cloudgray:
# Description

Fix `CheckTxHandler` to allow tx with "invalid sequence" error.

The purpose of this pr is aligning appside mempool with geth mempool.

Because current anteHandler rejects valid tx whose nonce is lower than tx in mempool, `CheckTxHandler` should handle this error and insert that tx into mempool.

The reason why does not modify anteHandler but `CheckTxHandler` is, if appside mempool is not enabled, validation of anteHandler is correct. So, we should modify `CheckTxHandler` that is used for appside mempool.

Closes: #590 

---

## Author Checklist

I have...

### DIFF
--- a/mempool/check_tx.go
+++ b/mempool/check_tx.go
@@ -18,7 +18,7 @@ func NewCheckTxHandler(mempool *ExperimentalEVMMempool) types.CheckTxHandler {
 		gInfo, result, anteEvents, err := runTx(request.Tx, nil)
 		if err != nil {
 			// detect if there is a nonce gap error (only returned for EVM transactions)
-			if errors.Is(err, ErrNonceGap) {
+			if errors.Is(err, ErrNonceGap) || errors.Is(err, sdkerrors.ErrInvalidSequence) {
 				// send it to the mempool for further triage
 				err := mempool.InsertInvalidNonce(request.Tx)
 				if err != nil {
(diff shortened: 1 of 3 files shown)


mempool/check_tx.go
```diff
@@ -18,7 +18,7 @@ func NewCheckTxHandler(mempool *ExperimentalEVMMempool) types.CheckTxHandler {
 		gInfo, result, anteEvents, err := runTx(request.Tx, nil)
 		if err != nil {
 			// detect if there is a nonce gap error (only returned for EVM transactions)
-			if errors.Is(err, ErrNonceGap) {
+			if errors.Is(err, ErrNonceGap) || errors.Is(err, sdkerrors.ErrInvalidSequence) {
```
vladjdk: Could we add a test that confirms that transactions that have already been included in a block will not be included in the mempool?

cloudgray: Sure. I'll add test cases tomorrow.

cloudgray: test case added! (https://github.com/cosmos/evm/pull/591/commits/839d35cd4f0fd39c212057f5210b5cb04bafea1c)

PR #595, opened 2025-09-10
fix: respect max_gas_wanted for mempool

### LINKED ISSUE #593: Bad `Maxtx-gas-want` implementation that may lead to proposal rejection of normal nodes.
https://github.com/cosmos/evm/blob/67950ad934f3393cf6a507dc1b5ad2aa46edaa60/ante/evm/08_gas_consume.go#L20
# introduce
In the following code, maxtxgaswant is used to limit the upper limit of messages.  However, in the actual implementation, it did not realize its function. On the contrary, due to the difference of the cumulative results of gaswant during checktx and processhandler, it may cause the normal node to send a proposal that was rejected by processhandler because it contained sub-message transactions beyond maxtxgaswant.

# Risk scenario
Specifically, for those nodes whose maxtxgaswant is not 0, malicious users can send a message that the tx contains a msg which gas greater than maxtxgaswant, so that the gas accumulation of the transaction in the checktx phase is smaller and enters the transaction pool. And during the processhandler, the proposal may be rejected because of the normal calculation of gas and the higher gaswant than before.

mmsqe:
# Description

* max-tx-gas-wanted was broken since sdk 50
* depends on https://github.com/cosmos/cosmos-sdk/pull/25338
* for test info: pytest -vv -s [test_max_gas_wanted.py](https://github.com/MANTRA-Chain/mantrachain-e2e/blob/gas_wanted/integration_tests/test_max_gas_wanted.py) --chain-config evmd

Closes: https://github.com/cosmos/evm/issues/593

---

## Author Checklist

I have...


vladjdk:
While we wait for the SDK change to be reviewed, could you fix the go.mod merge conflicts?

aljo242 (changes requested):
blocked on the cosmos-sdk replace directive: both go.mod files replace `cosmos/cosmos-sdk` with `github.com/mmsqe/cosmos-sdk v0.46.0-beta2.0.20260526155725-0c80f02aefb0`. needs cosmos/cosmos-sdk#25338 to land upstream before the replace can be dropped and this merged. the fix itself looks correct.

PR #598, opened 2025-09-12
perf: reduce number of query context calls

vladjdk:
Reduces the amount of times CreateQueryContext is called by caching the context once every block and reusing it.

closes: #604


aljo242:
@vladjdk can we add a changelog entry for this?

Eric-Warehime (approved):

Issue #446, opened 2025-08-13
Align JSON-RPC API responses/arguments with Geth


zsystm:
Several supported JSON-RPC APIs differ from Geth in their returned fields or supported features.  
This issue tracks the adjustments needed to achieve full compatibility.

---

## `eth_feeHistory`

| Type                   | Field(s) |
|------------------------|----------|
| **Missing field**      | `withdrawals ([]*Withdrawal)`, `header.excessBlobGas (*uint64)`, `header.blobGasUsed (*common.Hash)`, `header.parentBeaconBlockRoot (*common.Hash)`, `requestsHash (*common.Hash)`, `header.withdrawalsRoot (*common.Hash)` |
| **Always empty value** | `receiptsRoot (*common.Hash)` |
| **Extra (not in Geth)**| `totalDifficulty (*hexutil.Big)` |

---

## `eth_getBlockByNumber`

| Type                   | Field(s) |
|------------------------|----------|
| **Missing field**      | `withdrawals ([]*Withdrawal)`, `header.excessBlobGas (*uint64)`, `header.blobGasUsed (*common.Hash)`, `header.parentBeaconBlockRoot (*common.Hash)`, `requestsHash (*common.Hash)`, `header.withdrawalsRoot (*common.Hash)` |
| **Always empty value** | `receiptsRoot (*common.Hash)` |
| **Extra (not in Geth)**| `totalDifficulty (*hexutil.Big)` |

---

## `eth_getTransactionByBlockNumberAndIndex`

| Type              | Field(s) |
|-------------------|----------|
| **Missing field** | `yParity (*hexutil.Uint64)` |

---

## `debug_traceBlock`, `debug_traceBlockByNumber`, `debug_traceBlockByHash`

| Type                  | Note |
|-----------------------|------|
| **Missing feature**   | `StateOverrides` not supported (unlike Geth) |

---

**Recommendation:**  
- Add missing fields to match Geth’s return structure.  
- Remove extra fields not present in Geth, unless intentionally diverging (document if so).  
- Implement missing features where applicable (e.g., `StateOverrides`).

zsystm:
@cloudgray Please check `txpool `namespace` also. There should be differences between evmd and geth.

cloudgray:
I'll separate Issue for `debug` namespace apis.

It is because 
1. state override is releated to currently open issue #336 and pr #337.
2. current pr #599 for eth namespace apis is already heavy.

Issue #447, opened 2025-08-13
Implement eth_getHeaderByHash and eth_getHeaderByNumber


zsystm:
These JSON-RPC methods exist in Geth but are missing in our implementation.

They are required for EVM compatibility and should be implemented to return block header data in the same structure and format as Geth.

PR #599, opened 2025-09-15
fix(rpc): align jsonrpc apis with geth v1.16.3

cloudgray:
# Description

Closes: #446  #447

## Overview

The primary purpose of this PR was to make the return values of the Ethereum APIs compatible with go-ethereum v1.16.3.

However, as I progressed with the work, I found that methods for obtaining the same values had been developed in a disorganized way without clear separation of layers. It was obvious that leaving them as-is would make it very difficult to maintain compatibility with go-ethereum.

Therefore, I refactored the backend methods to be properly separated by layer `comet block query → comet-to-eth conversion → eth-to-rpc conversion`, making it easier to later migrate the `comet block subscription → comet-to-eth conversion & indexing` process to the indexer.

Please take this into consideration during the code review.

### 1. Add missing apis 

- `eth_getHeaderByNumber`
- `eth_getHeaderByHash`

#### Block Header

| Type                                | Field(s)   | Note        |
|------------------------|----------|----------|
| **Missing field**              | `withdrawals ([]*Withdrawal)`, `header.excessBlobGas (*uint64)`, `header.blobGasUsed (*common.Hash)`, `header.parentBeaconBlockRoot (*common.Hash)`, `requestsHash (*common.Hash)`, `header.withdrawalsRoot (*common.Hash)` | Add missing fields |
| **Always empty value**  | `receiptsRoot (*common.Hash)` | Fill with valid value  |
| **Invalid field**              | `transactionsRoot (common.Hash)`, `Time` | `transactionRoot`: Remove cosmos tx from root hash calculation, `Time`: change from local timestamp to UTC timestamp  |
| **Extra (not in Geth)**   | `totalDifficulty (*hexutil.Big)` | Remove |

Related APIs

- `eth_getHeaderByNumber`
- `eth_getHeaderByHash`
- `eth_getBlockByNumber`
- `eth_getBlockByHash`

### Block Body

| Type                                | Field(s)   | Note        |
|------------------------|----------|----------|
| **Invalid field**              | `size`     | change from comet block size to eth block size |

Related APIs

- `eth_getBlockByNumber`
- `eth_getBlockByHash`

### Transaction

| Type                                | Field(s)   | Note        |
|------------------------|----------|----------|
| **Missing field**            | `yParity`     | fill with valid value |

Related APIs

- `eth_getBlockByNumber`
- `eth_getBlockByHash`
- `eth_getTransactionByHash`
- `eth_getTransactionByBlockHashAndIndex`
- `eth_getTransactionByBlockNumberAndIndex`

#### Before

- Internal header queries only used for apis other than eth_getHeaderByNumber & eth_getHeaderByHash returns invalid headers.
- It is because [this method gets data used for making header only from comet header](https://github.com/cloudgray/evm/blob/enhance%2Fjsonrpc-compat-eth/rpc/types/utils.go#L58-L62), it cannot get many fields (`transactionRoot`, `receiptsRoot`, `bloom`, `gasUsed`, ...) that needs block body.

#### After

- [Use common block query method](https://github.com/cloudgray/evm/blob/enhance%2Fjsonrpc-compat-eth/rpc/backend/headers.go#L81-L91) to get all data to make valid header

Limitation

- Websocket api [still use invalid block headers](https://github.com/cloudgray/evm/blob/enhance%2Fjsonrpc-compat-eth/rpc/stream/rpc.go#L165-L167).
- It is because, to get block results and tx data in stream package, additional moification is needed.
- This work could be meaningless if indexer were improved and [`RPCStream`](https://github.com/cosmos/evm/blob/enhance%2Fjsonrpc-compat-eth/rpc/stream/rpc.go#L54-L67) does not subscribe cometBlockHeader but ethBlockHeader.
- So currently, websocket api will still publish block header that is missing many fields like receiptsRoot.

### 4. Fix `txpool` apis

- Handle EIP-7702 SetCode TxCase using common [`NewRPCTransaction`](https://github.com/cloudgray/evm/blob/enhance%2Fjsonrpc-compat-eth/rpc/types/utils.go#L272-L285)
- Before, the tx converting method existed separately and it does not support EIP-7702 SetCodeType tx case.

#### Before

```go
EthPendingBlockNumber  = BlockNumber(-2)
EthLatestBlockNumber   = BlockNumber(-1)
EthEarliestBlockNumber = BlockNumber(0)
```

#### After

```go
EthEarliestBlockNumber  = BlockNumber(-5)
EthSafeBlockNumber      = BlockNumber(-4)
EthFinalizedBlockNumber = BlockNumber(-3)
EthLatestBlockNumber    = BlockNumber(-2)
EthPendingBlockNumber   = BlockNumber(-1)
```

## Author Checklist

I have...

### DIFF
--- a/tests/jsonrpc/README.md
+++ b/tests/jsonrpc/README.md
@@ -57,15 +57,15 @@ go build .
 
 ### evmd Endpoints
 
-- **JSON-RPC**: <LOCAL_HOST>:8545
-- **WebSocket**: <LOCAL_HOST>:8546  
-- **Cosmos REST**: <LOCAL_HOST>:1317
-- **Tendermint RPC**:<LOCAL_HOST>:26657
+- **JSON-RPC**: http://localhost:8545
+- **WebSocket**: http://localhost:8546  
+- **Cosmos REST**: http://localhost:1317
+- **Tendermint RPC**:http://localhost:26657
 - **gRPC**: localhost:9090
 
 ### geth Endpoints
 
-- **JSON-RPC**: <LOCAL_HOST>:8547
+- **JSON-RPC**: http://localhost:8547
 - **WebSocket**: ws://localhost:8548
 
 ## Scripts Structure
@@ -123,4 +123,4 @@ The scripts use the following defaults:
 
 - Verify the container is running: `docker ps | grep evmd-jsonrpc-test`
 - Check if ports are bound: `docker port evmd-jsonrpc-test`
-- Test with curl: `curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' <LOCAL_HOST>:8545`
+- Test with curl: `curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"eth_chainId","params":[],"id":1}' http://localhost:8545`
--- a/mlc_config.json
+++ b/mlc_config.json
@@ -5,6 +5,18 @@
   "ignorePatterns": [
     {
       "pattern": "^https://twitter.com/.*"
+    },
+    {
+      "pattern": "^http://localhost.*"
+    },
+    {
+      "pattern": "^https://localhost.*"
+    },
+    {
+      "pattern": "^http://127\\.0\\.0\\.1.*"
+    },
+    {
+      "pattern": "^https://127\\.0\\.0\\.1.*"
     }
   ]
 }


rpc/namespaces/ethereum/eth/filters/api.go
```diff
@@ -47,7 +47,7 @@ type Backend interface {
 	CometBlockResultByNumber(height *int64) (*coretypes.ResultBlockResults, error)
 	GetLogs(blockHash common.Hash) ([][]*ethtypes.Log, error)
 	GetLogsByHeight(*int64) ([][]*ethtypes.Log, error)
-	BlockBloom(blockRes *coretypes.ResultBlockResults) (ethtypes.Bloom, error)
+	BlockBloomFromCometBlock(blockRes *coretypes.ResultBlockResults) (ethtypes.Bloom, error)
```
aljo242, line 50: why did we need to change this name?

rpc/namespaces/ethereum/eth/api.go
```diff
@@ -125,9 +125,29 @@ func NewPublicAPI(logger log.Logger, backend backend.EVMBackend) *PublicAPI {
...
+func (e *PublicAPI) GetHeaderByNumber(ethBlockNum rpctypes.BlockNumber) (map[string]interface{}, error) {
+	e.logger.Debug("eth_getHeaderByNumber", "number", ethBlockNum)
+	return e.backend.GetHeaderByNumber(ethBlockNum)
+}
+
+// GetHeaderByHash returns the requested header by hash.
+func (e *PublicAPI) GetHeaderByHash(hash common.Hash) (map[string]interface{}, error) {
+	e.logger.Debug("eth_getHeaderByHash", "hash", hash.Hex())
+	return e.backend.GetHeaderByHash(hash)
+}
```
aljo242, line 146: should we define `type Header = map[string]any` to be used here?

tests/jsonrpc/README.md
```diff
@@ -57,15 +57,15 @@ go build .
...
 
-- **JSON-RPC**: http://localhost:8545
-- **WebSocket**: ws://localhost:8546  
-- **Cosmos REST**: http://localhost:1317
-- **Tendermint RPC**: http://localhost:26657
+- **JSON-RPC**: <LOCAL_HOST>:8545
+- **WebSocket**: <LOCAL_HOST>:8546  
+- **Cosmos REST**: <LOCAL_HOST>:1317
+- **Tendermint RPC**:<LOCAL_HOST>:26657
 - **gRPC**: localhost:9090
```
aljo242, line 64: why not update this too?

tests/jsonrpc/README.md
```diff
@@ -57,15 +57,15 @@ go build .
...
 
 ### geth Endpoints
 
-- **JSON-RPC**: http://localhost:8547
+- **JSON-RPC**: <LOCAL_HOST>:8547
 - **WebSocket**: ws://localhost:8548
```
aljo242, line 69: same

rpc/namespaces/ethereum/eth/filters/api.go
cloudgray: I wanted to separate data conversion functions and real rpc backend query functions. 

In case of `BlockBloom`, there is no api like `eth_getBlockBloom` and it is just util function that converse cometbft block result to ethereum block bloom.

It should be removed from Backend interface. But now, to avoid confusion, I just rename function.

I think we need to discuss how we can store ethereum backend data like block header, transactionRoots, receiptsRoot, blockBlooms, ... at indexer level after `v0.5.0`. Then I think this conversion function can be located in indexer/utils.

rpc/namespaces/ethereum/eth/api.go
cloudgray: I don't think so.

I think the effect of defining a type is that it forces other packages or clients to import and use that type, while also allowing the use of interfaces like Validate, Marshal/Unmarshal, and Encode/Decode.

However, headers, blocks, and other data returned by the Ethereum RPC server can vary in their fields depending on the ChainConfig (hardfork config), and even when the fields are the same, as with transactions, the representation may differ (e.g., common.Hash or a full Tx struct) depending on the query parameters (e.g. `fullTx`). Because of this, I don’t think we can enforce a type here, nor would it provide much benefit.

This design is consistent with the current go-ethereum implementation, and I don’t see a reason to implement it differently.

tests/jsonrpc/README.md
cloudgray: It is because ci (markdown link check) fails ([ref](https://github.com/cosmos/evm/actions/runs/17969156735/job/51109061564))

It says this link is not available in github-ci instance.

I don't know the reason why it did not fail in past but fail currently...

tests/jsonrpc/README.md
cloudgray: same (https://github.com/cosmos/evm/pull/599#discussion_r2386505476)

tests/jsonrpc/README.md
vladjdk: Let's just set the CI to ignore

vladjdk (approved):
Approved with one reply

tests/jsonrpc/README.md
cloudgray: @vladjdk 
I added ignored pattern for markdown link check ci. (https://github.com/cosmos/evm/pull/599/commits/888b5fee1dd0f49d65eb1c9212ef36ee853140bf)

Issue #522, opened 2025-08-26
[Spike] Investigate nil pointer error in gentxs


vladjdk:
Because we use [noop configurators before we start the chain](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/evmd/cmd/evmd/cmd/root.go#L60-L68), there may be issues in gentxs and exporting genesis.

Investigate whether these issues exist, or are even a problem after [EVM-201](https://linear.app/cosmoslabs/issue/EVM-201/evmd-04-configs-issues) is merged in.

Issue #371, opened 2025-07-29
evmd should support flexible Chain IDs for external test suites


zsystm:
**Description:**
While attempting to integrate ethereum/tests into cosmos/evm, we encountered a blocking issue:
Currently, `evmd` only allows instantiating with a chain ID that exists in a fixed set (`ChainsCoinInfo` map). This prevents us from running test cases where the chain ID is dynamically set via test configuration.

For example, the test suite defines chain IDs like 1, 1234, 1337, etc., which are not present in the hardcoded set. Unless we modify each test file or fork the suite, we cannot use it as-is — which is not ideal.

Suggestion:
We should update `evmd` to accept arbitrary chain IDs in test mode or provide a fallback configuration so that tests from ethereum/tests can run without modifying the upstream suite.

Current limitation:
```go
var ChainsCoinInfo = map[uint64]evmtypes.EvmCoinInfo{
	EighteenDecimalsChainID: {
		Denom:         ExampleChainDenom,
		ExtendedDenom: ExampleChainDenom,
		DisplayDenom:  ExampleDisplayDenom,
		Decimals:      evmtypes.EighteenDecimals,
	},
	CosmosChainID: {
		Denom:         "atest",
		ExtendedDenom: "atest",
		DisplayDenom:  "test",
		Decimals:      evmtypes.EighteenDecimals,
	},
	EVMChainID: {
		Denom:         "atest",
		ExtendedDenom: "atest",
		DisplayDenom:  "test",
		Decimals:      evmtypes.EighteenDecimals,
	},
}
```

Impact:
This limitation blocks compatibility with standard Ethereum conformance tests unless we hardcode or fork chain IDs, which defeats the purpose of using a shared suite.

aljo242:
We believe that we should removed this fixed chain ID config entirely and allow everything to be configured via `app.toml` and validated at runtime

PR #608, opened 2025-09-16
refactor!: clean up chain config

almk-dev:
closes: #616, #371 and #522


almk-dev:
@Eric-Warehime @vladjdk please take a look for a first pass in terms of direction and generally use useful the new patterns are before I mark as r4r

PR #633, opened 2025-09-17
feat: separate geth metrics server

technicallyty:
# Description

separating the geth metrics server from the sdk server so it plays more nicely with alloy. 

calling this a feat and just updating the changelog line since the previous never made it into a release.

## Reviewers
please see my comments below

---

## Author Checklist

I have...


server/config/config.go
```diff
@@ -69,6 +70,9 @@ const (
 	// DefaultEVMMinTip is the default minimum priority fee for the mempool
 	DefaultEVMMinTip = 0
 
+	// DefaultGethMetricsAddr is the default port for the geth metrics server.
+	DefaultGethMetricsAddr = `"127.0.0.1:8080"`
```
technicallyty: weirdly this had to be wrapped in tick marks because otherwise it would write it without quotes and then the toml unmarshal would error.

metrics/geth.go
```diff
@@ -0,0 +1,56 @@
...
+	"time"
+
+	gethmetrics "github.com/ethereum/go-ethereum/metrics"
+	gethprom "github.com/ethereum/go-ethereum/metrics/prometheus"
+
+	"cosmossdk.io/log"
+)
+
+// StartGethMetricServer starts the geth metrics server on the specified address.
+func StartGethMetricServer(ctx context.Context, log log.Logger, addr string) error {
```
technicallyty, line 16: this is meant to mirror pretty closely how the API server is started in cosmos sdk here: https://github.com/cosmos/cosmos-sdk/blob/727faf04848461f248f3a84edfc73c657464cdb1/server/api/server.go#L112

server/config/config.go
technicallyty: also picked 8080 arbitrarily here. if you think it should be a different port please let me know!

nadim-az (approved):

metrics/geth.go
```diff
@@ -0,0 +1,56 @@
...
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", gethprom.Handler(gethmetrics.DefaultRegistry))
+
+	server := &http.Server{
```
Eric-Warehime, line 20: I guess we should set some default timeouts here.

server/start.go
```diff
@@ -9,9 +9,8 @@ import (
 	"path/filepath"
 	"runtime/pprof"
 
-	gethmetrics "github.com/ethereum/go-ethereum/metrics"
+	evmmetrics "github.com/cosmos/evm/metrics"
```
Eric-Warehime: Needs `gci`

Eric-Warehime (approved):
Needs some http Server config settings and formatting. Otherwise lgtm.

server/config/config.go
Eric-Warehime: 8080 is super commonly used. We should pick something different.

metrics/geth.go
technicallyty: added 10s timeouts

server/config/config.go
technicallyty: switched to 8100

PR #635, opened 2025-09-18
refactor: move default precompiles out of evmd config

vladjdk:
closes: #636 

Moves the precompile instantiation to be in line with keeper instantiation, and moved the default precompile builder to the EVM repo instead of making people include this in their own repos.


Eric-Warehime (approved):

evmd/app.go
```diff
@@ -486,6 +487,18 @@ func NewExampleApp(
 		&app.ConsensusParamsKeeper,
 		&app.Erc20Keeper,
 		tracer,
+	).WithStaticPrecompiles(
+		precompiletypes.DefaultStaticPrecompiles(
+			*app.StakingKeeper,
+			app.DistrKeeper,
+			app.BankKeeper,
```
yihuang, line 494: It changed from the `PreciseBankKeeper` to normal `BankKeeper` here, is that intentional?

PR #661, opened 2025-09-24
Vlad/remove app options

vladjdk:
# Description

Closes: #XXXX

---

## Author Checklist

I have...


evmd/cmd/evmd/cmd/testnet.go
```diff
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/cosmos/evm/config"
+	network2 "github.com/cosmos/evm/testutil/integration/evm/network"
+	evmtypes "github.com/cosmos/evm/x/vm/types"
```
aljo242: stinky imports and formatting

evmd/tests/integration/create_app.go
```diff
@@ -2,6 +2,8 @@ package integration
 
 import (
 	"encoding/json"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	srvflags "github.com/cosmos/evm/server/flags"
```
aljo242: need a `lint-fix` format

aljo242 (approved):
lgtm with nits

evmd/cmd/evmd/cmd/root.go
```diff
@@ -130,7 +126,7 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
-			customAppTemplate, customAppConfig := config.InitAppConfig(config.BaseDenom, config.EVMChainID) // TODO:VLAD - Remove this
+			customAppTemplate, customAppConfig := config.InitAppConfig(config.TestExtendedDenom, config.EVMChainID) // TODO:VLAD - Remove this
```
almkx: Using `Test*` variables in `root.go` seems like an anti-pattern and it seems that `DefaultExtendedDenom` would be easier to understand.

ante/evm/mono_decorator_test.go
```diff
@@ -147,7 +147,6 @@ func toMsgSlice(msgs []*evmsdktypes.MsgEthereumTx) []sdk.Msg {
 
 func TestMonoDecorator(t *testing.T) {
 	chainID := uint64(config.EighteenDecimalsChainID)
-	require.NoError(t, config.EvmAppOptions(chainID))
```
almkx, line 150: Was this necessary before (i.e. tests break without this call)? Or was it removed because it was entirely extraneous?

docs/migrations/v0.4.0_to_v0.5.0_UNRELEASED.md
```diff
@@ -113,6 +113,73 @@ function can now happen *alongside the keeper instantiation*, and not after. The
...
+
+`evmd_config.go`, `chain_id.go`, `config.go`, `constants.go` have been moved to
+`github.com/cosmos/evm/config` and may be removed to your repo.
+
+#### UpgradeHandler
+
+As the configs have been moved to state and genesis, you must include an UpgradeHandler if your chain does
+not satisfy the following conditions.
+1. Your EVM Denom set in the `x/vm` params must have `DenomMetadata` registered for it in `x/bank`.
+2. Your EVM Denom must have a display denom associated with in in the `DenomMetadata`.
```
almkx:
```suggestion
2. Your EVM Denom must have a display denom associated with it in `DenomMetadata`.
```

docs/migrations/v0.4.0_to_v0.5.0_UNRELEASED.md
```diff
@@ -113,6 +113,73 @@ function can now happen *alongside the keeper instantiation*, and not after. The
...
+   1. The display denom for the EVM Denom must have an accurate decimal value (i.e. for `uatom`, `atom` must have a decimal value of 6.
```
almkx, line 173: This is a bit confusing as the "display" denom as it is used in the SDK isn't quite the same definition as it is used in this repo. 

For example, in the `evm` repo, the display denom would always be `atom`, while the denom for different decimal precisions would be `aatom`, `uatom`, etc. 

Is it worth specifying these definitions somewhere?

evmd/tests/integration/create_app.go
```diff
@@ -55,9 +57,8 @@ func SetupEvmd() (ibctesting.TestingApp, map[string]json.RawMessage) {
 		dbm.NewMemDB(),
 		nil,
 		true,
-		simutils.EmptyAppOptions{},
+		NewAppOptionsWithFlagHomeAndChainID("", constants.ExampleEIP155ChainID),
```
almkx, line 59: I think the different usages of `Test*`, `Example*`, and just descriptive config variables is confusing. It would be easier if default values were named accordingly, like `DefaultChainID`. 

Otherwise, it reads like we and using an example value that was created for testing convenience, as the default value used when creating an app.

evmd/test_helpers.go
```diff
@@ -79,7 +81,7 @@ func Setup(t *testing.T, chainID string, evmChainID uint64) *EVMD {
 	acc := authtypes.NewBaseAccount(senderPrivKey.PubKey().Address().Bytes(), senderPrivKey.PubKey(), 0, 0)
 	balance := banktypes.Balance{
 		Address: acc.GetAddress().String(),
-		Coins:   sdk.NewCoins(sdk.NewCoin(sdk.DefaultBondDenom, math.NewInt(100000000000000))),
+		Coins:   sdk.NewCoins(sdk.NewCoin(types.DefaultEVMExtendedDenom, math.NewInt(100000000000000))),
```
almkx, line 84: Re: point above, it seems strange that we use the default here, but an "example" in the actual app creator.

mempool/mempool.go
```diff
@@ -54,8 +53,6 @@ type (
 		logger        log.Logger
 		txConfig      client.TxConfig
 		blockchain    *Blockchain
-		bondDenom     string
```
almkx, line 57: Is it worth standardizing the different terms (i.e. bond denom, staking denom, base denom)? A base denom can be used for staking, but shouldn't necessarily be referred to like that to avoid confusion. 

Either way, the relationship between the two should be made clear in a docstring

proto/cosmos/evm/vm/v1/evm.proto
```diff
@@ -348,3 +353,10 @@ message Preinstall {
   // code in hex format for the preinstall contract
   string code = 3;
 }
+
+message EvmCoinInfo {
+  string denom = 1;
+  string extended_denom = 2;
+  string display_denom = 3;
+  uint32 decimals = 4;
+}
```
almkx, line 362: This is inconsistent with what "display" denom means in the SDK metadata fields.

tests/integration/precompiles/ics20/test_integration.go
```diff
@@ -319,14 +320,14 @@ func TestPrecompileIntegrationTestSuite(t *testing.T, evmAppCreator ibctesting.A
...
 			// check contract balance
-			contractBalance := evmAppA.GetBankKeeper().GetBalance(
+			contractBalance := evmAppA.GetEVMKeeper().GetBalance(
 				s.chainA.GetContext(),
-				ics20CallerAddr.Bytes(),
-				sourceBondDenom,
+				ics20CallerAddr,
 			)
-			Expect(contractBalance.Amount).To(Equal(fundAmt), "Contract balance should be equal to the fund amount")
+			Expect(contractBalance.ToBig()).To(Equal(fundAmt.Mul(math.NewInt(1e12)).BigInt()), "Contract balance should be equal to the fund amount")
```
almkx: Split to two variables as this is hard to read now.

tests/integration/precompiles/staking/test_staking.go
```diff
@@ -381,7 +381,7 @@ func (s *PrecompileTestSuite) TestRun() {
 				s.Require().NoError(err, "failed to pack input")
 				return input
 			},
-			19103, // use enough gas to avoid out of gas error
+			21295, // use enough gas to avoid out of gas error
```
almkx, line 384: Why did this change?

tests/integration/precompiles/werc20/test_integration.go
```diff
@@ -556,7 +556,7 @@ func TestPrecompileIntegrationTestSuite(t *testing.T, create network.CreateEvmAp
 						ChainID:    is.network.GetChainID(),
 						EVMChainID: is.network.GetEIP155ChainID().Uint64(),
 					}]
-					Expect(decimals).To(Equal(uint8(coinInfo.Decimals)), "expected different decimals")
+					Expect(decimals).To(Equal(uint8(coinInfo.Decimals)), "expected different decimals") //nolint:gosec // G115
```
almkx, line 559: Both of these are ideally `uint32`. Should we track a future task in getting these to match?

testutil/integration/evm/network/amounts.go
```diff
@@ -21,15 +21,15 @@ func DefaultInitialAmounts() InitialAmounts {
 	baseCoinInfo := testconstants.ExampleChainCoinInfo[defaultChain]
 
 	return InitialAmounts{
-		Base: GetInitialAmount(baseCoinInfo.Decimals),
```
almkx, line 24: Nit: another place where it's inconsistent that `base`:`evm` is to `bonded`:`extended`

testutil/integration/evm/network/coins.go
```diff
@@ -40,7 +40,7 @@ func DefaultChainCoins() ChainCoins {
 func getCoinInfo(coinInfo evmtypes.EvmCoinInfo) CoinInfo {
 	return CoinInfo{
```
almkx, line 41: We should (eventually) get rid of this duplicate CoinInfo struct as well in the network setup.

x/vm/keeper/coin_info.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+// GetEvmCoinInfo returns the EVM Coin Info stored in the module
+func (k Keeper) GetEvmCoinInfo(ctx sdk.Context) (coinInfo types.EvmCoinInfo) {
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(types.KeyPrefixEvmCoinInfo)
+	if bz == nil {
+		return coinInfo
+	}
+	k.cdc.MustUnmarshal(bz, &coinInfo)
```
almkx, line 16: Is this a risk of panic and halting the chain if the coin info fields are malformed, since now we no longer have the config validation?

x/vm/types/config.go
```diff
@@ -58,3 +55,22 @@ func GetEthChainConfig() *geth.ChainConfig {
 func GetChainConfig() *ChainConfig {
 	return chainConfig
 }
+
+// SetChainConfig allows to set the `chainConfig` variable modifying the
+// default values. The method is private because it should only be called once
+// in the EVMConfigurator.
+func SetChainConfig(cc *ChainConfig) error {
```
almkx, line 62: Why is this outside `Configure()` now? `EVMConfigurator` still has the `chainConfig` field.

x/vm/types/config_testing.go
```diff
@@ -23,13 +21,9 @@ var testChainConfig *ChainConfig
...
-	if ec.sealed {
-		return fmt.Errorf("error configuring EVMConfigurator: already sealed and cannot be modified")
-	}
-
-	if err := setTestChainConfig(ec.chainConfig); err != nil {
-		return err
-	}
+	//if ec.sealed {
+	//	return fmt.Errorf("error configuring EVMConfigurator: already sealed and cannot be modified")
+	//}
```
almkx: Commented code. Also, we can remove the seal field entirely if it's not useful with the `test` tag either (there's only one occurance).

x/vm/genesis.go
```diff
@@ -19,7 +20,9 @@ func InitGenesis(
 	ctx sdk.Context,
 	k *keeper.Keeper,
 	accountKeeper types.AccountKeeper,
+	bankKeeper types.BankKeeper,
 	data types.GenesisState,
+	initializer *sync.Once,
```
almkx, line 25: Nit: Is `initializer` a good name here? Something can be re-initialized, so it's not for sure that it can only happen once. Something like `onceMutex` is more literal but I'm not convinced that's better either. 🤷

x/vm/module.go
```diff
@@ -129,6 +136,15 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	types.RegisterQueryServer(cfg.QueryServer(), am.keeper)
 }
 
+func (am AppModule) PreBlock(goCtx context.Context) (appmodule.ResponsePreBlock, error) {
```
almkx, line 139: So we are re-applying this at the start of every block? Can this be a possible attack vector, i.e. we somehow change the param values between blocks?

Is the expectation that the `sync.Once` mutex does the heavy lifting here in ensuring that we never *actually* try to set it more than once? If that's the case, I'm not clear why we need this pre block logic, since we are are returning the same hardcoded struct each time, and we already do it once in `InitGenesis()`. 🤔

x/vm/module.go
```diff
@@ -176,3 +192,78 @@ func (am AppModule) IsAppModule() {}
...
+
+	if err := vmKeeper.SetEvmCoinInfo(ctx, coinInfo); err != nil {
+		panic(err)
+	}
+
+	configurator := types.NewEVMConfigurator()
+	// if withReset {
+	//	// reset configuration to set the new one
+	//	configurator.ResetTestConfig()
+	//}
```
almkx: Comment cleanup

almk-dev (commented):
I this this is a well considered subset of changes to be included from all the goals of the original PR, plus some nice new improvements. 

Generally looks good! Some thoughts and minor comments.

ante/evm/mono_decorator_test.go
vladjdk: This was necessary before, as this would implicitly set the global vars

x/vm/module.go
vladjdk: InitGenesis is not run in the case of chains restarting after already having run once. PreBlock was necessary to make sure the globals are reinitialized

tests/integration/precompiles/staking/test_staking.go
vladjdk: I believe this is because the params got slightly bigger, so when we save a multistore snapshot, the size increases, and this is the only test to use an *exact* value, as opposed to other tests overestimating. However, not entirely sure

proto/cosmos/evm/vm/v1/evm.proto
vladjdk: This is consistent to bank metadata fields and how they're meant to be used.

ante/evm/mono_decorator_test.go
vladjdk: Fixed: https://github.com/cosmos/evm/pull/661/commits/6ec0eeafbefcfb793bc2f2d778ea8c7398fe6f23

evmd/test_helpers.go
vladjdk: Replaced with default: https://github.com/cosmos/evm/pull/661/commits/c6e2c56f575587299e3f2bfe62e74a988e93ab39

evmd/cmd/evmd/cmd/root.go
vladjdk: Replaced with default: https://github.com/cosmos/evm/pull/661/commits/c6e2c56f575587299e3f2bfe62e74a988e93ab39

docs/migrations/v0.4.0_to_v0.5.0_UNRELEASED.md
vladjdk: Definition is in the ADR: https://docs.cosmos.network/main/build/architecture/adr-024-coin-metadata

mempool/mempool.go
vladjdk: Base denom is probably the way to go: https://github.com/cosmos/evm/issues/693

tests/integration/precompiles/ics20/test_integration.go
vladjdk: Fixed: https://github.com/cosmos/evm/pull/661/commits/af5caaaeb13a2624607eed1b96fce8612c3acba1

tests/integration/precompiles/werc20/test_integration.go
vladjdk: https://github.com/cosmos/evm/pull/661/commits/e8ead269cdde5c33a80ff1949e764e057909339c

testutil/integration/evm/network/amounts.go
vladjdk: Added the issue #693

x/vm/keeper/coin_info.go
vladjdk: Likely not - it's marshalled every time it's set: https://github.com/cosmos/evm/blob/79d63fc3504dc4020fe6ad39bd58ab7c8fc24461/x/vm/keeper/coin_info.go#L21

x/vm/types/config.go
vladjdk: Removed! The chainConfig field wasn't actually doing anything: https://github.com/cosmos/evm/pull/661/commits/cc38dfbc1771a555360cb4811e0484c5699a0610****

x/vm/genesis.go
vladjdk: I'm fine with the name personally

x/vm/module.go
vladjdk: https://github.com/cosmos/evm/pull/661/commits/97e8c3897cf00eab07592b6ba0af758d5a3ae62b

x/vm/types/config_testing.go
vladjdk: https://github.com/cosmos/evm/pull/661/commits/97e8c3897cf00eab07592b6ba0af758d5a3ae62b

evmd/cmd/evmd/cmd/root.go
```diff
@@ -131,7 +127,7 @@ func NewRootCmd() *cobra.Command {
 				return err
 			}
 
-			customAppTemplate, customAppConfig := config.InitAppConfig(config.BaseDenom, config.EVMChainID) // TODO:VLAD - Remove this
+			customAppTemplate, customAppConfig := config.InitAppConfig(types.DefaultEVMExtendedDenom, config.EVMChainID) // TODO:VLAD - Remove this
```
cloudgray, line 129: I think denom should be `DefaultEVMDenom` instead of `DefaultEVMExtendedDenom`.

Is there any reason that `DefaultEVMExtendedDenom` is used?

evmd/app.go
```diff
@@ -211,7 +211,6 @@ func NewExampleApp(
 	loadLatest bool,
 	appOpts servertypes.AppOptions,
 	evmChainID uint64, // TODO:VLAD - Remove this
-	evmAppOptions evmconfig.EVMOptionsFn, // TODO:VLAD - Remove this
 	baseAppOptions ...func(*baseapp.BaseApp),
 ) *EVMD {
 	encodingConfig := evmosencoding.MakeConfig(evmChainID) // TODO:VLAD - Remove chain id from this
```
cloudgray: For encoding config (EIP155 ChainID for EIP712 data format), chainId from argument is used. 
But for evmKeeper, evmChainId from srvFlag is used.

Considering the TODO comment, chainId from argument will be removed.
How about remove this in this PR?

testutil/ibc/chain.go
```diff
@@ -150,7 +152,29 @@ func NewTestChainWithValSet(tb testing.TB, isEVM bool, coord *Coordinator, chain
...
+	metadata := []banktypes.Metadata{{
+		Description: "",
+		DenomUnits: []*banktypes.DenomUnit{
+			{
+				Denom:    types.DefaultEVMExtendedDenom,
+				Exponent: 0,
+				Aliases:  nil,
+			},
+			{
+				Denom:    types.DefaultEVMDisplayDenom,
```
cloudgray, line 164: I think DefaultDEMDenom is correct in this case. 
WDYT?

x/precisebank/module.go
```diff
@@ -78,7 +78,7 @@ func (AppModuleBasic) ValidateGenesis(cdc codec.JSONCodec, config client.TxEncod
 	if err != nil {
 		return err
 	}
-	return gs.Validate()
```
cloudgray, line 81: Why was this validation removed?

x/vm/types/config.go
cloudgray: Is the reason for setting chainConfig outside of `Configure` letting ChainConfig be configurable after sealing?

x/vm/types/config_testing.go
```diff
@@ -71,6 +66,25 @@ func setTestChainConfig(cc *ChainConfig) error {
 	return nil
 }
 
+// SetChainConfig allows to set the `chainConfig` variable modifying the
+// default values. The method is private because it should only be called once
+// in the EVMConfigurator.
+func SetChainConfig(cc *ChainConfig) error {
```
cloudgray, line 72: I think this method and `setTestChainConfig` is functionally almost same and `setTestChainConfig` is not used. How about removing `setTestChainConfig`?

x/vm/types/denom.go
```diff
@@ -54,7 +54,7 @@ var ConversionFactor = map[Decimals]math.Int{
 }
 
 // Decimals represents the decimal representation of a Cosmos coin.
-type Decimals uint8
+type Decimals uint32
```
cloudgray, line 57: Is there any reason changing the type of Decimals?
I'm just curious.

testutil/ibc/chain.go
vladjdk: This should actually be the display here

evmd/app.go
vladjdk: Fixed!

PR #665, opened 2025-09-25
feat(encoding): include an address codec implementation

srdtrk:
# Description

closes: #633
closes: EVM-322

---

## Author Checklist

I have...


utils/address_codec.go
```diff
@@ -0,0 +1,56 @@
...
+		return bz, nil
+	default:
+		return nil, sdkerrors.ErrUnknownAddress.Wrapf("unknown address format: %s", text)
+	}
+}
+
+// BytesToString decodes bytes to text
+func (bc EvmCodec) BytesToString(bz []byte) (string, error) {
+	return common.BytesToAddress(bz).Hex(), nil
+}
```
srdtrk, line 63: We could also use bech32 in BytesToString. I see some pros to this. Such as this codec becoming a potential drop in replacement for the default bech32 address codec

aljo242: we should document that this will return the hex address in the godoc

utils/address_codec.go
```diff
@@ -0,0 +1,56 @@
...
+
+var _ address.Codec = (*EvmCodec)(nil)
+
+// NewEvmCodec returns a new EvmCodec with the given bech32 prefix
+func NewEvmCodec(prefix string) address.Codec {
+	return EvmCodec{prefix}
+}
+
+// StringToBytes encodes text to bytes
```
aljo242: we should document the behavior more explicitly in this godoc

utils/address_codec.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"cosmossdk.io/core/address"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/bech32"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+// EvmCodec defines an address codec for EVM compatible cosmos modules
+type EvmCodec struct {
```
aljo242: This lgtm but I think this should live elsewhere (not in a `utils` package and also be piped into our `evmd` application

aljo242: @vladjdk for thoughts on locations etc

vladjdk: Good location would be `encoding/codec`, since this is a codec after all :)

vladjdk: Also agree on adding this into `evmd`

srdtrk: If we are going to pipe this into our `evmd`. Then we might have a better case for https://github.com/cosmos/evm/pull/665#discussion_r2379208030 since if we used bech32, this codec would be a better drop-in replacement for the default cosmos sdk one

aljo242: @vladjdk thoughts?

vladjdk: Yeah lgtm! @srdtrk

srdtrk: After talking to vlad, we decided to use bech32 in BytesToString, and pass it to evmd

PR #686, opened 2025-10-02
feat: support unordered txs

alexanderbez:
## Changelog

Allow for EVM txs to utilize the unordered tx feature of the SDK. It works by essentially mimicking and using the same values/fields as the native `SigVerificationDecorator`.

* Refactor `IncrementNonce` to be a method on `MonoDecorator`
* Refactor `IncrementNonce` to consider unordered txs
* Add `BuildTxWithFactory` on `MsgEthereumTx`

Note:

* This targets the v0.4.x line directly, the latest release. I will create a PR against `main` if this is approved.
* We assume a single signer always.
* Right now, only EVM txs sent via `tx evm raw` are supported. In other words, EVM txs that are intercepted via the EVM JSON RPC (and mempool?) do not support setting unordered values. There is no way, AFAIK, that allows for granular setting of custom tx fields via the RPC. If there is a clean approach, we can do that in a future separate PR.

### DIFF
--- a/tests/integration/ante/test_evm_unit_09_increment_sequence.go
+++ b/tests/integration/ante/test_evm_unit_09_increment_sequence.go
@@ -56,6 +56,13 @@ func (s *EvmUnitAnteTestSuite) TestIncrementSequence() {
 		},
 	}
 
+	md := evm.NewEVMMonoDecorator(
+		unitNetwork.App.GetAccountKeeper(),
+		unitNetwork.App.GetFeeMarketKeeper(),
+		unitNetwork.App.GetEVMKeeper(),
+		0,
+	)
+
 	for _, tc := range testCases {
 		s.Run(tc.name, func() {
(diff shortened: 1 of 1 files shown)


x/vm/types/msg.go
```diff
@@ -323,18 +325,29 @@ func (msg *MsgEthereumTx) BuildTx(b client.TxBuilder, evmDenom string) (signing.
...
 		return nil, err
 	}
+
 	builder.SetFeeAmount(fees)
 	builder.SetGasLimit(msg.GetGas())
-	tx := builder.GetTx()
-	return tx, nil
+	builder.SetTimeoutHeight(txf.TimeoutHeight())
+	builder.SetTimeoutTimestamp(txf.TimeoutTimestamp())
+	builder.SetUnordered(txf.Unordered())
```
yihuang, line 341: These fields are not signed by user, right? would that make it malleable by malicious thirdparty?

PR #698, opened 2025-10-09
refactor(config): Expose mempool configuration flags and move mempool app configuration to helper

almk-dev:
# Description
- Moves out the mempool configurations into a separate helper in `evmd/mempool.go`
- Expose mempool configuration flags and sets them in the evmd app.go mempool configuration helper

Closes: [APPS-9](https://linear.app/cosmoslabs/issue/APPS-9/refactor-out-mempool-configuration-in-appgo-into-separate-helper)
Closes: [STACK-1519](https://linear.app/cosmoslabs/issue/STACK-1519/expose-mempool-config-in-apptoml)

---

## Author Checklist

### DIFF
--- a/tests/systemtests/eip712/test_suite.go
+++ b/tests/systemtests/eip712/test_suite.go
@@ -144,6 +144,7 @@ func DefaultNodeArgs() []string {
 	apiEnable := "--api.enable=true"
 	jsonrpcApi := "--json-rpc.api=eth,txpool,personal,net,debug,web3"
 	jsonrpcAllowUnprotectedTxs := "--json-rpc.allow-unprotected-txs=true"
+	maxTxs := "--mempool.max-txs=0"
 
-	return []string{jsonrpcApi, chainID, evmChainID, apiEnable, jsonrpcAllowUnprotectedTxs}
+	return []string{jsonrpcApi, chainID, evmChainID, apiEnable, jsonrpcAllowUnprotectedTxs, maxTxs}
 }


server/config/config.go
```diff
@@ -151,6 +151,65 @@ type EVMConfig struct {
...
+	if c.AccountQueue < 1 {
+		return fmt.Errorf("account queue must be at least 1, got %d", c.AccountQueue)
+	}
+	if c.GlobalQueue < 1 {
+		return fmt.Errorf("global queue must be at least 1, got %d", c.GlobalQueue)
+	}
+	if c.Lifetime < 1 {
+		return fmt.Errorf("lifetime must be at least 1 nanosecond, got %s", c.Lifetime)
+	}
+	return nil
```
almkx, line 212: @vladjdk please look over these defaults and validation rules, as there are somewhat arbitrary to start.

vladjdk: Defaults match legacypool defaults, so lgtm!

server/config/toml.go
```diff
@@ -28,6 +28,30 @@ min-tip = {{ .EVM.MinTip }}
 # GethMetricsAddress defines the addr to bind the geth metrics server to. Default 127.0.0.1:8100.
 geth-metrics-address = "{{ .EVM.GethMetricsAddress }}"
 
+# Mempool configuration for EVM transactions
+[evm.mempool]
```
vladjdk, line 32: Have you tested this locally to confirm that these pass through to change the defaults? Confirming a local test is fine here

config/server_app_options.go
```diff
@@ -92,3 +103,46 @@ func GetMinTip(appOpts servertypes.AppOptions, logger log.Logger) *uint256.Int {
...
+		legacyConfig.Lifetime = lifetime
+	}
+
+	return &legacyConfig
+}
+
+func GetCosmosPoolMaxTx(appOpts servertypes.AppOptions, logger log.Logger) int {
+	if appOpts == nil {
+		logger.Error("app options is nil, using default cosmos pool max tx of 0 (uncapped)")
+		return 0
```
vladjdk, line 149: Why was this -1 by default in the current version?

config/server_app_options.go
```diff
@@ -82,6 +88,11 @@ func GetMinGasPrices(appOpts servertypes.AppOptions, logger log.Logger) sdk.DecC
 // GetMinTip reads the min tip from the app options, set from app.toml
 // This field is also known as the minimum priority fee
 func GetMinTip(appOpts servertypes.AppOptions, logger log.Logger) *uint256.Int {
+	if appOpts == nil {
```
vladjdk, line 96: Why did we change the behaviour here?

config/server_app_options.go
```diff
@@ -21,6 +22,11 @@ import (
 // GetBlockGasLimit reads the genesis json file using AppGenesisFromFile
 // to extract the consensus block gas limit before InitChain is called.
 func GetBlockGasLimit(appOpts servertypes.AppOptions, logger log.Logger) uint64 {
+	if appOpts == nil {
```
vladjdk, line 25: Ditto

config/server_app_options.go
almkx: I suppose the default behavior is to have a no-op mempool then? Is that's the case, I can match it to be `-1`.

vladjdk: I think so -- that's what we currently see in app.go

server/config/toml.go
almkx: Good catch, I missed adding the server bindings.
Fixed: https://github.com/cosmos/evm/pull/698/commits/6800093b73ca2633aeb1f632856d496b032db64e

Confirmed locally, using dummy 7 values:
```
12:57AM INF starting ABCI with CometBFT module=server
------------------------------------------------------
legacyConfig: {Locals:[] NoLocals:false Journal:transactions.rlp Rejournal:1h0m0s PriceLimit:7 PriceBump:7 AccountSlots:77 GlobalSlots:7777 AccountQueue:77 GlobalQueue:777 Lifetime:7h7m7s}
------------------------------------------------------
mempool.max-tx: 7
12:57AM INF starting node with ABCI CometBFT in-process module=server
```

config/server_app_options.go
almkx: The behavior isn't really changed, I just wanted to an earlier nil check and return the same default, since `AppOptions` is an interface and not guaranteed to be initialized.

config/server_app_options.go
almkx: > The behavior isn't really changed, I just wanted to an earlier nil check and return the same default, since AppOptions is an interface and not guaranteed to be initialized.

config/server_app_options.go
almkx: Fixed: https://github.com/cosmos/evm/pull/698/commits/6800093b73ca2633aeb1f632856d496b032db64e

Issue #520, opened 2025-08-26
[Spike] Investigate EIP-712


vladjdk:
Investigate and answer the following questions:

* Because we no longer enforce EIP-155 on a consensus level, does [this line](https://github.com/cosmos/evm/blob/3977168b066e13dda869721bf0b7bab8b057f6ad/ethereum/eip712/encoding.go#L30) allow users to get around replay protection if they use EIP-712 messages?
* What does [this line](https://github.com/cosmos/evm/blob/3977168b066e13dda869721bf0b7bab8b057f6ad/encoding/config.go#L49) do? Do we need it? Can we remove it? Document this in the code in a more descriptive way.
* Are we confident in the testing right now? Can we include some E2E tests with various cases?

PR #700, opened 2025-10-10
test: eip712 e2e tests

vladjdk:
closes: #520

### DIFF
--- a/tests/systemtests/clients/cosmosclient.go
+++ b/tests/systemtests/clients/cosmosclient.go
@@ -192,7 +192,7 @@ func (c *CosmosClient) GetBalance(nodeID string, address sdk.AccAddress, denom s
 		Denom:   denom,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to query balance: %v", err)
+		return nil, fmt.Errorf("failed to query balance: %w", err)
 	}
 
 	return res.Balance.Amount.BigInt(), nil
--- a/tests/systemtests/eip712/test_suite.go
+++ b/tests/systemtests/eip712/test_suite.go
@@ -82,7 +82,7 @@ func (s *SystemTestSuite) SendBankSendWithEIP712(
 		gasPrice,
 	)
 	if err != nil {
-		return "", fmt.Errorf("failed to send bank send with EIP-712: %v", err)
+		return "", fmt.Errorf("failed to send bank send with EIP-712: %w", err)
 	}
 
 	return resp.TxHash, nil


tests/systemtests/clients/config.go
```diff
@@ -22,7 +21,7 @@ import (
 
 const (
 	ChainID    = "local-4221"
-	EVMChainID = "262144"
+	EVMChainID = 4221
```
almkx, line 24: This was changed just to match the `ChainID` value, right? Can we use the evm config constants here instead, unless there's a good reason to hardcode it?

tests/systemtests/clients/cosmosclient.go
```diff
@@ -183,20 +181,43 @@ func (c *CosmosClient) UnconfirmedTxs(nodeID string) (*coretypes.ResultUnconfirm
...
+	c.ClientCtx = c.ClientCtx.WithClient(c.RpcClients[nodeID])
+
+	queryClient := banktypes.NewQueryClient(c.ClientCtx)
+
+	res, err := queryClient.Balance(context.Background(), &banktypes.QueryBalanceRequest{
+		Address: address.String(),
+		Denom:   denom,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to query balance: %v", err)
```
almkx:
```suggestion
		return nil, fmt.Errorf("failed to query balance: %w", err)
```

tests/systemtests/eip712/eip712_utils.go
```diff
@@ -0,0 +1,148 @@
...
+	signMode := signing.SignMode_SIGN_MODE_LEGACY_AMINO_JSON
+
+	txBuilder := cosmosClient.ClientCtx.TxConfig.NewTxBuilder()
+
+	txBuilder.SetGasLimit(150_000)
+	txBuilder.SetFeeAmount(sdk.NewCoins(sdk.NewCoin("atest", sdkmath.NewIntFromBigInt(gasPrice).MulRaw(150_001))))
+
+	err := txBuilder.SetMsgs(msg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to set messages: %v", err)
```
almkx, line 75: Nit: Replace all %v with %w in this file:

```suggestion
		return nil, fmt.Errorf("failed to set messages: %w", err)
```

tests/systemtests/eip712/eip712_utils.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	txBuilder.SetMemo("")
```
almkx: It's not clear to me why this step is needed. Maybe a clarifying comment.

tests/systemtests/eip712/test_suite.go
```diff
@@ -0,0 +1,149 @@
...
+		nodeID,
+		accID,
+		from,
+		to,
+		sdkmath.NewIntFromBigInt(amount),
+		nonce,
+		gasPrice,
+	)
+	if err != nil {
+		return "", fmt.Errorf("failed to send bank send with EIP-712: %v", err)
```
almkx: Nit: Replace all %v with %w in this file:
```suggestion
		return "", fmt.Errorf("failed to send bank send with EIP-712: %w", err)
```

tests/systemtests/clients/config.go
vladjdk: It was just done to use something other than the default value. Using defaults could allow faulty config propagation to slip through.

almkx: Not worth adding to this PR, but that intent might be clearer with a random chain ID generator function for use in tests.

vladjdk: Good idea!

tests/systemtests/eip712/eip712_utils.go
vladjdk: Ended up removing this as it was just a copy from the integration tests

almkx: :big:

Issue #722, opened 2025-10-14
Fix unit test CI flakes


almk-dev:
Solves for the following cases from the parent issue:
- Unit tests no longer have flaky race condition failures

PR #732, opened 2025-10-16
fix(ci): solve gas meter race condition in integration tests

almk-dev:
# Description
Addresses the following tests:
- `TestCheckTxHandlerForCommittedAndLowerNonceTxs` race and flake
  - `runReorg()`/`StateAt` race
- `TestBlockchainRaceCondition` nil pointer error

Adds teardowns to integration `evmd` integration tests and added locks for state reads and commits. There are two sets of locking functions for the same `testingCommitMu` mutex:
- `BeginCommit()`/`EndCommit()`: This is only used in network integration tests and is a loose lock around `app.Commit()`, so it cannot happen in prod.
- `BeginRead()`/`EndRead()`: This is a tighter lock that is actually called inside `LegacyPool.runReorg()`, so we have two versions of the lock, separated with build tag. It actually holds the lock if we use the `-race` flag, but is a no-op otherwise. The lock will never be held in prod, as we directly run the binary and don't use the go toolchain to execute (so the `RACE` env var is not applicable).
  - The explicit sync is only necessary in testing because of rapid commits and block progression (like calling `NextBlock()` 100 times in a loop without delay), whereas block times are at fixed or regular cadences in production.
  - Also, the `CacheContext()` call all but guarantees against dirty reads in practice (in prod). However, the race detector looks for explicit locks and holds whenever there are goroutines, since it only checks for access patterns and doesn't evaluate logical consistency like block times. 
    - That's why it's necessary to add a `race`-only lock to appease the race detector, even if it's not actually a concern in practice. 

Adds the `count=1` flag to Makefile test targets so that flakes are discovered earlier (locally). 

Closes: #722

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### BUG FIXES
 
 - [\#748](https://github.com/cosmos/evm/pull/748) Fix DynamicFeeChecker in Cosmos ante handler to respect NoBaseFee feemarkets' parameter.
+- [\#769](https://github.com/cosmos/evm/pull/769) Fix erc20 ibc middleware to not to validate sender address format.
 - [\#756](https://github.com/cosmos/evm/pull/756) Fix error message typo in NewMsgCancelProposal.
 - [\#772](https://github.com/cosmos/evm/pull/772) Avoid panic on close if evm mempool not used.
 - [\#774](https://github.com/cosmos/evm/pull/774) Emit proper allowance amount in erc20 event.
--- a/tests/integration/x/erc20/test_ibc_callback.go
+++ b/tests/integration/x/erc20/test_ibc_callback.go
@@ -102,28 +102,28 @@ func (s *KeeperTestSuite) TestOnRecvPacketRegistered() {
 			expCoins:      coins,
 		},
 		{
-			name: "error - invalid sender (no '1')",
+			name: "success - invalid sender (no '1')",
 			malleate: func() {
 				transfer := transfertypes.NewFungibleTokenPacketData(registeredDenom, "100", "evmos", ethsecpAddrCosmos, "")
 				bz := transfertypes.ModuleCdc.MustMarshalJSON(&transfer)
 				packet = channeltypes.NewPacket(bz, 100, transfertypes.PortID, sourceChannel, transfertypes.PortID, cosmosEVMChannel, timeoutHeight, 0)
 			},
 			receiver:      secpAddr,
-			ackSuccess:    false,
-			checkBalances: false,
+			ackSuccess:    true,
+			checkBalances: true,
 			expErc20s:     big.NewInt(0),
 			expCoins:      coins,
 		},
 		{
-			name: "error - invalid sender (bad address)",
+			name: "success - invalid sender (bad address)",
 			malleate: func() {
 				transfer := transfertypes.NewFungibleTokenPacketData(registeredDenom, "100", "badba1sv9m0g7ycejwr3s369km58h5qe7xj77hvcxrms", ethsecpAddrCosmos, "")
 				bz := transfertypes.ModuleCdc.MustMarshalJSON(&transfer)
 				packet = channeltypes.NewPacket(bz, 100, transfertypes.PortID, sourceChannel, transfertypes.PortID, cosmosEVMChannel, timeoutHeight, 0)
 			},
 			receiver:      secpAddr,
-			ackSuccess:    false,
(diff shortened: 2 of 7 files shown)


mempool/blockchain.go
```diff
@@ -44,6 +44,10 @@ type Blockchain struct {
 	previousHeaderHash common.Hash
 	latestCtx          sdk.Context
 	mu                 sync.RWMutex
+	// commitMu guards access to state during ABCI Commit.
+	// Readers acquire RLock via StateAt; the committer acquires Lock
+	// around Commit to avoid concurrent iavl access in tests.
+	commitMu sync.RWMutex
```
aljo242: I think we should call this something like `testingCommitMu` for clarity

vladjdk:
Regarding the `TestBlockchainRaceCondition` test, are we sure that this race condition only occurs during testing, and not during production. If it occurs in production, is it a problem? If so, what should we do to solve it?

almk-dev:
@vladjdk There is actually no data race in the `TestBlockchainRaceCondition` test. It's just a nil pointer error that was introduced because now we call `CacheContext()` so we need to actually set up the test properly with a multistore.  

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x22df70b]

goroutine 43 [running]:
github.com/cosmos/cosmos-sdk/types.Context.CacheContext({{0x0, 0x0}, {0x0, 0x0}, {{0x0, 0x0}, {0x0, 0x0}, 0x1, {0x0, ...}, ...}, ...})
	/home/runner/go/pkg/mod/github.com/cosmos/cosmos-sdk@v0.53.4/types/context.go:357 +0xeb
github.com/cosmos/evm/mempool.(*Blockchain).StateAt(0xc000c70808, {0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef, 0x12, ...})
	/home/runner/work/evm/evm/mempool/blockchain.go:217 +0x538
github.com/cosmos/evm/mempool_test.TestBlockchainRaceCondition.func3()
	/home/runner/work/evm/evm/mempool/blockchain_test.go:99 +0x155
created by github.com/cosmos/evm/mempool_test.TestBlockchainRaceCondition in goroutine 41
	/home/runner/work/evm/evm/mempool/blockchain_test.go:95 +0x1365
FAIL	github.com/cosmos/evm/mempool	0.157s
```

mempool/blockchain.go
almkx: https://github.com/cosmos/evm/pull/732/commits/f4809e553ab96db3f3cf76a2cc2dbbf7fe8fe6b3

tests/integration/mempool/test_mempool_integration.go
```diff
@@ -134,6 +134,8 @@ func (s *IntegrationTestSuite) TestMempoolInsert() {
 
 	for _, tc := range testCases {
 		s.Run(tc.name, func() {
+			// Clean up previous test's resources before resetting
+			s.TearDownTest()
```
aljo242, line 138: do we need these calls?  Feels like we should be calling this at the end of tests?

Makefile
```diff
@@ -156,7 +156,7 @@ test-race: run-tests
 
 test-evmd: ARGS=-timeout=15m
 test-evmd:
-	@cd evmd && go test -race -tags=test -mod=readonly $(ARGS) $(EXTRA_ARGS) $(PACKAGES_EVMD)
+	@cd evmd && go test -count=1 -race -tags=test -mod=readonly $(ARGS) $(EXTRA_ARGS) $(PACKAGES_EVMD)
```
aljo242, line 159: what does this change do?

almkx: > Adds the count=1 flag to Makefile test targets so that flakes are discovered earlier (locally).

This makes it so runs aren't cached, and will re-run it fresh each time. Although the cached runs are nice and fast, it hides flakiness when one successful run is cached and all subsequent runs pass, making it seem more stable than it is. 

I think it's a useful change for local development, and doesn't affect CI runs since those are one-off each time anyways.

tests/integration/mempool/test_mempool_integration.go
almkx: We need these at the beginning, in the case where the _**first**_ test case in a function is affected by goroutines spawned by a different set of tests from a different function. 

In that case, having it at the end only fully handles cleanup for previous cases within the same test function.

aljo242 (approved):

PR #742, opened 2025-10-20
fix(rpc): correct block number conversion in eth_getProof

Cordtus:
Fixes `eth_getProof` RPC endpoint failing with "cannot query with height < 0" error when using special block parameters like "latest", "pending",
  or "earliest".

  **Problem:** Direct type cast `int64(blockNum)` preserves negative values from special block constants (EthLatestBlockNumber = -2,
  EthPendingBlockNumber = -1, EthEarliestBlockNumber = -5), failing query validation

  **Fix:** Use `blockNum.Int64()` method which converts negative values to 0
  
  `rpc/backend/` (see GetCode, GetBalance, GetStorageAt).

  **Review:** Check `rpc/backend/account_info.go:51` - for consistency with other methods.
  
    ---

  ## Author Checklist

  I have...

  - [Y] tackled an existing issue or discussed with a team member
  - [Y] left instructions on how to review the changes
  - [Y] targeted the `main` branch


rpc/backend/account_info.go
```diff
@@ -48,7 +48,7 @@ func (b *Backend) GetProof(address common.Address, storageKeys []string, blockNr
 		return nil, err
 	}
 
-	height := int64(blockNum)
+	height := blockNum.Int64()
```
vladjdk, line 51: Let's refactor the Int64() function to be named AdjustedInt64() to clearly denote that it transforms 0's into 1's. Otherwise, PR looks good to me!

vladjdk (changes requested):

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### BUG FIXES
 
+- [\#742](https://github.com/cosmos/evm/pull/742) Fix eth_getProof block number conversion for special block parameters (latest, pending, earliest).
```
vladjdk: This also doesn't seem correct - would change this

Issue #765, opened 2025-10-27
[Bug] Gas estimation is not aligned with geth


cloudgray:
## Problem

The result of gas estimation is different from the result from go-ethereum.

### Test Result

Test Name | Estimated Gas
-- | --
native transfer | geth: 21000 evmd: 21000
native transfer with state overrides | geth: 21000 evmd: 21000
erc20 deploy | geth: 1899426 evmd: 1883312
erc20 mint with deployed contract | geth: 51564 evmd: 51172
erc20 mint with state overrides | geth: 67571 evmd: 66723
erc20 transfer with state overrides | geth: 52114 evmd: 49140

#### 1. Binary search logic is different between cosmos/evm and go-ethereum.

<details>
<summary>geth</summary>
<pre><code class="language-go">
// Binary search for the smallest gas limit that allows the tx to execute successfully.
for lo+1 < hi {
	if opts.ErrorRatio > 0 {
		// It is a bit pointless to return a perfect estimation, as changing
		// network conditions require the caller to bump it up anyway. Since
		// wallets tend to use 20-25% bump, allowing a small approximation
		// error is fine (as long as it's upwards).
		if float64(hi-lo)/float64(hi) < opts.ErrorRatio {
			break
		}
	}
	mid := lo + (hi-lo)/2
	if mid > lo*2 {
		// Most txs don't need much higher gas limit than their gas used, and most txs don't
		// require near the full block limit of gas, so the selection of where to bisect the
		// range here is skewed to favor the low side.
		mid = lo * 2
	}
	failed, _, err = execute(ctx, call, opts, mid)
	if err != nil {
		// This should not happen under normal conditions since if we make it this far the
		// transaction had run without error at least once before.
		log.Error("Execution error in estimate gas", "err", err)
		return 0, nil, err
	}
	if failed {
		lo = mid
	} else {
		hi = mid
	}
}
</code></pre>
</details>

<details>
<summary>cosmos/evm</summary>
<pre><code>
for lo+1 < hi {
	mid := (hi + lo) / 2
	failed, _, err := executable(mid)
	// If the error is not nil(consensus error), it means the provided message
	// call or transaction will never be accepted no matter how much gas it is
	// assigned. Return the error directly, don't struggle any more.
	if err != nil {
		return 0, err
	}
	if failed {
		lo = mid
	} else {
		hi = mid
	}
}
</code></pre>

</details>

#### 2. `feemarketParams.MinGasMultiplier` is applied when gas estimation.
[...]

PR #766, opened 2025-10-27
fix: align gas estimation logic with go-ethereum v1.16.3

cloudgray:
# Description

Closes: #765 

### 1. Binary search logic is different between cosmos/evm and go-ethereum.

- Apply Geth’s optimization for gas estimation binary search, [allowing a 1.5% margin of error](https://github.com/ethereum/go-ethereum/blob/737ffd1bf0cbee378d0111a5b17ae4724fb2216c/eth/gasestimator/gasestimator.go#L181-L189) between the low and high values.

- Modify the [logic for calculating the mid](https://github.com/ethereum/go-ethereum/blob/737ffd1bf0cbee378d0111a5b17ae4724fb2216c/eth/gasestimator/gasestimator.go#L190-L196) value to match Geth’s implementation.

- Modify default low value as `GasUsed - 1` from execution. Not `ethparams.TxGas (21000) - 1`

### 2. feemarketParams.MinGasMultiplier is applied when gas estimation.

When performing `EstimateGas`, set the internal value passed to `ApplyMessageWithConfig` to true so that `feemarketParams.MinGasMultiplier` is not applied.

### 3. Incorrect Storage Handling from StateDB (especially for the case origin value == changed value)

- Rework `SetState`/`setState` to track current and original values exactly like go-ethereum, clearing dirty flags when reverting to the original value.
- Add `SetStateOverride` so overrides go straight into originStorage, making the slot appear committed.
- Update `StateDB.Finalise` to call `clearJournalAndRefund()` ([mirroring go-ethereum](https://github.com/ethereum/go-ethereum/blob/737ffd1bf0cbee378d0111a5b17ae4724fb2216c/core/state/statedb.go#L980-L984)) so the journal and refund counter are reset after overrides.
- Update the RPC override path to call `SetStateOverride` instead of managing dirty slots manually.
- Update `StateDB.Commit()`: set `stateObject.dirtyStorage` value to `stateObject.originalStorage` and in case of cacheCtxCommit, does not delete `stateObject.dirtyStorage` value
   - It is for handling changes from `stateObject.setState()` that now removes dirty storage when changed value equals to original storage value.

### 4. "EIP-7623: Data-heavy transactions pay the floor gas" is not applied 

- Apply EIP-7623.

---

## Author Checklist

I have...

### DIFF
--- a/x/vm/keeper/grpc_query.go
+++ b/x/vm/keeper/grpc_query.go
@@ -410,7 +410,8 @@ func (k Keeper) EstimateGasInternal(c context.Context, req *types.EthCallRequest
 			// resetting the gasMeter after increasing the sequence to have an accurate gas estimation on EVM extensions transactions
 			tmpCtx = buildTraceCtx(tmpCtx, msg.GasLimit)
 		}
-		// pass false to not commit StateDB
+		// pass "commit" as false to avoid committing StateDB
+		// pass "internal" as true to avoid applying feemarketParams.MinGasMultiplier for gas estimation.
 		rsp, err = k.ApplyMessageWithConfig(tmpCtx, *msg, nil, false, cfg, txConfig, true, overrides)
 		if err != nil {
 			if errors.Is(err, core.ErrIntrinsicGas) || errors.Is(err, core.ErrFloorDataGas) {
--- a/x/vm/statedb/journal.go
+++ b/x/vm/statedb/journal.go
@@ -127,8 +127,8 @@ type (
 	storageChange struct {
 		account   *common.Address
 		key       common.Hash
-		prevalue  common.Hash
-		origvalue common.Hash
+		preValue  common.Hash
+		origValue common.Hash
 	}
 	transientStorageChange struct {
 		account       *common.Address
@@ -250,7 +250,7 @@ func (ch codeChange) Dirtied() *common.Address {
 }
 
 func (ch storageChange) Revert(s *StateDB) {
-	s.getStateObject(*ch.account).setState(ch.key, ch.prevalue, ch.origvalue)
+	s.getStateObject(*ch.account).setState(ch.key, ch.preValue, ch.origValue)
 }
 
 func (ch storageChange) Dirtied() *common.Address {
--- a/x/vm/statedb/state_object.go
+++ b/x/vm/statedb/state_object.go
@@ -256,8 +256,8 @@ func (s *stateObject) SetState(key common.Hash, value common.Hash) common.Hash {
 	s.db.journal.append(storageChange{
 		account:   &s.address,
 		key:       key,
-		prevalue:  prev,
-		origvalue: origin,
+		preValue:  prev,
+		origValue: origin,
 	})
 	s.setState(key, value, origin)
 	return prev


x/vm/keeper/grpc_query.go
```diff
@@ -411,7 +411,7 @@ func (k Keeper) EstimateGasInternal(c context.Context, req *types.EthCallRequest
 			tmpCtx = buildTraceCtx(tmpCtx, msg.GasLimit)
 		}
 		// pass false to not commit StateDB
```
aljo242: comment should also be adjusted - why is it ok to pass true now?

x/vm/statedb/journal.go
```diff
@@ -117,8 +125,10 @@ type (
 		prev    uint64
 	}
 	storageChange struct {
-		account       *common.Address
-		key, prevalue common.Hash
+		account   *common.Address
+		key       common.Hash
+		prevalue  common.Hash
+		origvalue common.Hash
```
aljo242:
```suggestion
		preValue  common.Hash
		origValue common.Hash
```

nit

x/vm/statedb/state_object.go
```diff
@@ -251,13 +251,15 @@ func (s *stateObject) SetState(key common.Hash, value common.Hash) common.Hash {
...
+	origin := s.GetCommittedState(key)
 	// New value is different, update and journal the change
 	s.db.journal.append(storageChange{
-		account:  &s.address,
-		key:      key,
-		prevalue: prev,
+		account:   &s.address,
+		key:       key,
+		prevalue:  prev,
+		origvalue: origin,
```
aljo242:
```suggestion
		preValue:  prev,
		origValue: origin,
```
nit

x/vm/types/utils.go
```diff
@@ -172,8 +172,26 @@
...
+	const errRatio = 0.015
+
 	for lo+1 < hi {
-		mid := (hi + lo) / 2
+		if errRatio > 0 {
+			// It is a bit pointless to return a perfect estimation, as changing
+			// network conditions require the caller to bump it up anyway. Since
+			// wallets tend to use 20-25% bump, allowing a small approximation
+			// error is fine (as long as it's upwards).
+			if float64(hi-lo)/float64(hi) < errRatio {
```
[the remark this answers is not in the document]
aljo242: estimation so this is fine

x/vm/keeper/grpc_query.go
cloudgray: "false" in this comment seems to mean commit = false
What I changed is internal = true

It seems confusing. I'll clarify comment!

cloudgray: applied (https://github.com/cosmos/evm/pull/766/commits/52c35953a0eeb1283c7a31fb7c7dce0ce7fe4490)

x/vm/statedb/journal.go
cloudgray: applied (https://github.com/cosmos/evm/pull/766/commits/52c35953a0eeb1283c7a31fb7c7dce0ce7fe4490)

x/vm/statedb/state_object.go
cloudgray: applied (https://github.com/cosmos/evm/pull/766/commits/52c35953a0eeb1283c7a31fb7c7dce0ce7fe4490)

PR #768, opened 2025-10-27
feat(precompiles)!: added ics02 client router precompile

srdtrk:
# Description

Closes: STACK-1652

---

## Author Checklist

I have...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 ### BUG FIXES
 
 - [\#748](https://github.com/cosmos/evm/pull/748) Fix DynamicFeeChecker in Cosmos ante handler to respect NoBaseFee feemarkets' parameter.
+- [\#690](https://github.com/cosmos/evm/pull/690) Fix Ledger hardware wallet support for coin type 60.
 - [\#766](https://github.com/cosmos/evm/pull/766) Align gas estimation logic with go-ethereum v1.16.3
 - [\#769](https://github.com/cosmos/evm/pull/769) Fix erc20 ibc middleware to not to validate sender address format.
 - [\#756](https://github.com/cosmos/evm/pull/756) Fix error message typo in NewMsgCancelProposal.
--- a/client/keys/add.go
+++ b/client/keys/add.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"flag"
 	"fmt"
 	"io"
 	"os"
@@ -13,7 +14,9 @@ import (
 	"github.com/spf13/cobra"
 
 	cryptohd "github.com/cosmos/evm/crypto/hd"
+	evmkeyring "github.com/cosmos/evm/crypto/keyring"
 	bip39 "github.com/cosmos/go-bip39"
+	ledger "github.com/cosmos/ledger-cosmos-go"
 
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
@@ -22,6 +25,8 @@ import (
 	"github.com/cosmos/cosmos-sdk/crypto/hd"
 	"github.com/cosmos/cosmos-sdk/crypto/keyring"
 	"github.com/cosmos/cosmos-sdk/crypto/keys/multisig"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	cosmosLedger "github.com/cosmos/cosmos-sdk/crypto/ledger"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
@@ -172,7 +177,37 @@ func RunAddCmd(ctx client.Context, cmd *cobra.Command, args []string, inBuf *buf
 	// If we're using ledger, only thing we need is the path and the bech32 prefix.
 	if useLedger {
 		bech32PrefixAccAddr := sdk.GetConfig().GetBech32AccountAddrPrefix()
-
+		needHardware := flag.Lookup("test.v") == nil && os.Getenv("GO_TEST") != "1"
+		if needHardware {
+			switch coinType {
+			case 60:
+				cosmosLedger.SetDiscoverLedger(func() (cosmosLedger.SECP256K1, error) {
+					return evmkeyring.LedgerDerivation()
+				})
+				cosmosLedger.SetCreatePubkey(func(key []byte) cryptotypes.PubKey {
+					return evmkeyring.CreatePubkey(key)
+				})
+				cosmosLedger.SetAppName(evmkeyring.AppName)
+				cosmosLedger.SetDERConversion(false)
+			case 118:
+				cosmosLedger.SetDiscoverLedger(func() (cosmosLedger.SECP256K1, error) {
+					device, err := ledger.FindLedgerCosmosUserApp()
+					if err != nil {
+						return nil, err
+					}
+					return device, nil
+				})
+				cosmosLedger.SetCreatePubkey(func(key []byte) cryptotypes.PubKey {
+					return &secp256k1.PubKey{Key: key}
+				})
+				cosmosLedger.SetAppName(cosmosLedger.AppName)
+				cosmosLedger.SetDERConversion(true)
+			default:
+				return fmt.Errorf(
+					"unsupported coin type %d for Ledger. Supported coin types: 60 (Ethereum app), 118 (Cosmos app)", coinType,
+				)
+			}
+		}
 		// use the provided algo to save the ledger key
 		k, err := kb.SaveLedgerKey(name, algo, bech32PrefixAccAddr, coinType, account, index)
 		if err != nil {
--- a/evmd/cmd/evmd/cmd/root.go
+++ b/evmd/cmd/evmd/cmd/root.go
@@ -2,10 +2,11 @@ package cmd
 
 import (
 	"errors"
-	"github.com/cosmos/evm/utils"
-	"github.com/cosmos/evm/x/vm/types"
 	"io"
 	"os"
+
+	"github.com/cosmos/evm/utils"
+	"github.com/cosmos/evm/x/vm/types"
 
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	"github.com/spf13/cast"
@@ -18,7 +19,7 @@ import (
 	dbm "github.com/cosmos/cosmos-db"
 	cosmosevmcmd "github.com/cosmos/evm/client"
 	evmdebug "github.com/cosmos/evm/client/debug"
-	cosmosevmkeyring "github.com/cosmos/evm/crypto/keyring"
+	"github.com/cosmos/evm/crypto/hd"
 	"github.com/cosmos/evm/evmd"
 	"github.com/cosmos/evm/evmd/config"
 	cosmosevmserver "github.com/cosmos/evm/server"
@@ -81,7 +82,7 @@ func NewRootCmd() *cobra.Command {
 		WithHomeDir(config.MustGetDefaultNodeHome()).
 		WithViper(""). // In simapp, we don't use any prefix for env variables.
 		// Cosmos EVM specific setup
-		WithKeyringOptions(cosmosevmkeyring.Option()).
+		WithKeyringOptions(hd.EthSecp256k1Option()).
 		WithLedgerHasProtobuf(true)
 
 	rootCmd := &cobra.Command{
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/cosmos/go-bip39 v1.0.0
 	github.com/cosmos/gogoproto v1.7.0
 	github.com/cosmos/ibc-go/v10 v10.0.0-beta.0.0.20251027215440-22f0033d0aee
+	github.com/cosmos/ledger-cosmos-go v0.16.0
 	github.com/creachadair/tomledit v0.0.28
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/ethereum/go-ethereum v1.16.5
@@ -123,7 +124,6 @@ require (
 	github.com/cosmos/gogogateway v1.2.0 // indirect
 	github.com/cosmos/iavl v1.2.6 // indirect
 	github.com/cosmos/ics23/go v0.11.0 // indirect
-	github.com/cosmos/ledger-cosmos-go v0.16.0 // indirect
 	github.com/crate-crypto/go-eth-kzg v1.3.0 // indirect
 	github.com/crate-crypto/go-ipa v0.0.0-20240724233137-53bbb0ceb27a // indirect
 	github.com/creachadair/atomicfile v0.3.7 // indirect
--- a/wallets/usbwallet/hub.go
+++ b/wallets/usbwallet/hub.go
@@ -55,7 +55,7 @@ type Hub struct {
 func NewLedgerHub() (*Hub, error) {
 	return newHub(LedgerScheme, 0x2c97, []uint16{
 		// Device definitions taken from
-		// https://github.com/LedgerHQ/ledger-live/blob/38012bc8899e0f07149ea9cfe7e64b2c146bc92b/libs/ledgerjs/packages/devices/src/index.ts
+		// https://github.com/LedgerHQ/ledger-live/blob/595cb73b7e6622dbbcfc11867082ddc886f1bf01/libs/ledgerjs/packages/devices/src/index.ts
 
 		// Original product IDs
 		0x0000, /* Ledger Blue */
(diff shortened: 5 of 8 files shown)


precompiles/ics02/ICS02I.sol
```diff
@@ -0,0 +1,65 @@
...
+address constant ICS02_PRECOMPILE_ADDRESS = 0x0000000000000000000000000000000000000808;
+
+/// @dev The ICS02 contract's instance.
+ICS02I constant ICS02_CONTRACT = ICS02I(ICS02_PRECOMPILE_ADDRESS);
+
+/// @author CosmosLabs
+/// @title ICS02 Client Router Precompile Interface
+/// @dev The interface through which solidity contracts will interact with IBC Light Clients (ICS02)
+/// @custom:address 0x0000000000000000000000000000000000000808
+interface ICS02I {
```
srdtrk, line 16: We don't need misbehavior here since in IBC-Go, `update_client = misbehaviour`

precompiles/ics02/ics02.go
```diff
@@ -0,0 +1,148 @@
...
+	storetypes "cosmossdk.io/store/types"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+var _ vm.PrecompiledContract = (*Precompile)(nil)
+
+const (
+	// TODO: These gas values are placeholders and should be determined through proper benchmarking.
```
srdtrk: I put my reduced estimation. I can actually benchmark solidity IBC Eureka, in which case the tx would be around ~300,000 due to zk proofs there

precompiles/ics02/ics02.go
```diff
@@ -0,0 +1,148 @@
...
+	bankKeeper cmn.BankKeeper,
+	clientKeeper ibcutils.ClientKeeper,
+) *Precompile {
+	return &Precompile{
+		cdc: cdc,
+		Precompile: cmn.Precompile{
+			KvGasConfig:           storetypes.GasConfig{},
+			TransientKVGasConfig:  storetypes.GasConfig{},
+			ContractAddress:       common.HexToAddress(evmtypes.ICS02PrecompileAddress),
+			BalanceHandlerFactory: cmn.NewBalanceHandlerFactory(bankKeeper),
```
srdtrk: I'm not sure I need to set this, since this contract is never expected to adjust balances. Then I can also remove bankKeeper from here

precompiles/ics02/ics02.go
```diff
@@ -0,0 +1,148 @@
...
+
+// IsTransaction checks if the given method name corresponds to a transaction or query.
+func (Precompile) IsTransaction(method *abi.Method) bool {
+	// GetClientStateMethod is the only query method.
+	return GetClientStateMethod != method.Name
+}
+
+// Logger returns a precompile-specific logger.
+func (p Precompile) Logger(ctx sdk.Context) log.Logger {
+	return ctx.Logger().With("evm extension", "clients")
```
srdtrk:
```suggestion
	return ctx.Logger().With("evm extension", "ics02")
```

precompiles/ics02/ics02.go
```diff
@@ -0,0 +1,148 @@
...
+		return GasUpdateClient
+	case VerifyMembershipMethod:
+		return GasVerifyMembership
+	case VerifyNonMembershipMethod:
+		return GasVerifyNonMembership
+	// Read-only transactions
+	case GetClientStateMethod:
+		return GasGetClientState
+	default:
+		return 0
```
srdtrk: Maybe we should panic here as unreachable? But other precompiles don't do this

precompiles/ics02/ics02.go
```diff
@@ -0,0 +1,148 @@
...
+		return p.GetClientState(ctx, contract, stateDB, method, args)
+	default:
+		return nil, fmt.Errorf(cmn.ErrUnknownMethod, method.Name)
+	}
+}
+
+// IsTransaction checks if the given method name corresponds to a transaction or query.
+func (Precompile) IsTransaction(method *abi.Method) bool {
+	// GetClientStateMethod is the only query method.
+	return GetClientStateMethod != method.Name
```
srdtrk: Could also create a switch statement and panic if it is an unrecognized method as unreacheable. But other precompiles don't do switch as well

precompiles/ics02/query.go
```diff
@@ -0,0 +1,45 @@
...
+
+	clientStateAny, err := codectypes.NewAnyWithValue(clientState)
+	if err != nil {
+		return nil, err
+	}
+	if len(clientStateAny.Value) == 0 {
+		return nil, fmt.Errorf("client state not found for client ID %s", clientID)
+	}
+
+	return method.Outputs.Pack(clientStateAny.Value)
```
srdtrk, line 44: I could also just proto marshal the codectypes.Any and return the bytes of that. Instead of returning the any.Value. This would be more similar to what the query_server does in ibc-go

precompiles/ics02/tx.go
```diff
@@ -0,0 +1,150 @@
...
+const (
+	UpdateClientMethod        = "updateClient"
+	VerifyMembershipMethod    = "verifyMembership"
+	VerifyNonMembershipMethod = "verifyNonMembership"
+)
+
+const (
+	UpdateResultSuccess      uint8 = 0
+	UpdateResultMisbehaviour uint8 = 1
+	UpdateResultNoop         uint8 = 2
```
srdtrk: This is because abi encodes enums as uint8

precompiles/ics02/tx.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, err
+	}
+
+	if host.ClientIdentifierValidator(clientID) != nil {
+		return nil, errorsmod.Wrapf(
+			clienttypes.ErrInvalidClient,
+			"invalid client ID: %s",
+			clientID,
+		)
+	}
```
srdtrk, line 52: This check is technically redundant as the keeper won't let you update the client if the clientID doesn't already exist. However, similar checks exist in ics20 precompile, so I added this for consistency.

precompiles/ics02/tx.go
```diff
@@ -0,0 +1,150 @@
...
```
srdtrk, line 89: ditto

precompiles/ics02/tx.go
```diff
@@ -0,0 +1,150 @@
...
```
srdtrk, line 129: ditto

precompiles/ics02/types.go
```diff
@@ -0,0 +1,113 @@
...
+		return "", nil, clienttypes.Height{}, nil, fmt.Errorf("invalid proof bytes: %v", args[1])
+	}
+
+	var proofHeight height
+	heightArg := abi.Arguments{method.Inputs[2]}
+	if err := heightArg.Copy(&proofHeight, []interface{}{args[2]}); err != nil {
+		return "", nil, clienttypes.Height{}, nil, fmt.Errorf("error while unpacking args to TransferInput struct: %s", err)
+	}
+
+	// TODO: make sure path is deserilized like this
```
srdtrk, line 105:
```suggestion
```

x/vm/types/precompiles.go
```diff
@@ -13,6 +13,7 @@ const (
 	BankPrecompileAddress         = "0x0000000000000000000000000000000000000804"
 	GovPrecompileAddress          = "0x0000000000000000000000000000000000000805"
 	SlashingPrecompileAddress     = "0x0000000000000000000000000000000000000806"
+	ICS02PrecompileAddress        = "0x0000000000000000000000000000000000000808"
```
srdtrk: I took the address `0x0000000000000000000000000000000000000808` instead of `0x0000000000000000000000000000000000000807` because when I grepped the repo, I saw that the latter is used somewhere. Not sure exactly where it is used, but I can take the latter if it is free

precompiles/ics02/ics02.go
vladjdk: https://github.com/cosmos/evm/blob/fe790fb8b398cf4bc99dfe5e727d39e5ed28d195/precompiles/common/precompile.go#L51

We could just use the dynamic calculator unless you think there's a better reason to guesstimate here.

precompiles/ics02/ics02.go
vladjdk: You don't need to set this for precompiles that don't set balances. Let's remove.

evmd/tests/ibc/ics02_precompile_client_test.go
```diff
@@ -0,0 +1,439 @@
...
+		calldata  []byte
+		expErr    bool
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+	}{
+		{
+			name: "success: update client",
```
vladjdk, line 161: any failure cases that we can add here?

evmd/tests/ibc/ics02_precompile_client_test.go
```diff
@@ -0,0 +1,439 @@
...
+				s.Require().NoError(err)
+
+				calldata, err = s.chainAPrecompile.Pack(ics02.UpdateClientMethod, clientID, updateBz)
+				s.Require().NoError(err)
+				// ====
+
+				err = s.chainA.App.(*evmd.EVMD).IBCKeeper.ClientKeeper.UpdateClient(s.chainA.GetContext(), clientID, header)
+				s.Require().NoError(err)
+
+				// TODO: right now, precompile always returns UpdateResultSuccess even on noop
```
vladjdk: Could you add an issue for this?

evmd/tests/ibc/ics02_precompile_client_test.go
```diff
@@ -0,0 +1,439 @@
...
+		expErr    bool
+		expResult *big.Int
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+	}{
+		{
+			name: "success: prove membership of clientState",
```
vladjdk, line 386: Let's add a failure case here as well

evmd/tests/ibc/ics02_precompile_client_test.go
```diff
@@ -0,0 +1,439 @@
...
+		expErr    bool
+		expResult *big.Int
+	)
+
+	testCases := []struct {
+		name     string
+		malleate func()
+	}{
+		{
+			name: "success: prove non-membership of clientState",
```
vladjdk, line 697: ditto

precompiles/ics02/ics02.go
vladjdk: I think we could remove this and use the regular gas calc

tests/integration/precompiles/staking/test_staking.go
```diff
@@ -381,7 +381,7 @@ func (s *PrecompileTestSuite) TestRun() {
 				s.Require().NoError(err, "failed to pack input")
 				return input
 			},
-			21295, // use enough gas to avoid out of gas error
+			21559, // use enough gas to avoid out of gas error
```
srdtrk, line 384: I had to do this because this gas needs to match exactly with the amount of gas used during these errors (because of the assertion [here](https://github.com/cosmos/evm/blob/1fe71eabfae673a04c852ae02267e980bf6c59ea/tests/integration/precompiles/staking/test_staking.go?plain=1#L467).

I guess this can be explained if the total number of precompiles impact the gas usage here

tests/integration/precompiles/staking/test_staking.go
```diff
@@ -391,7 +391,7 @@ func (s *PrecompileTestSuite) TestRun() {
 			func(_ keyring.Key) []byte {
 				return []byte("invalid")
 			},
-			21295, // use enough gas to avoid out of gas error
+			21559, // use enough gas to avoid out of gas error
```
srdtrk, line 394: ditto

precompiles/ics02/ics02.go
vladjdk: I'd make a switch statement to reject any messages that aren't in the list of message supported by the precompile. No need to panic, just return false: https://github.com/cosmos/evm/blob/fe790fb8b398cf4bc99dfe5e727d39e5ed28d195/precompiles/distribution/distribution.go#L187-L199

precompiles/ics02/query.go
vladjdk: I would do what you prefer here and how you would most benefit from using this. Either way works for me.

x/vm/types/precompiles.go
vladjdk: Let's use 807, that's an accidental artifact left over from the callbacks address, which we determined has no use so removed it.

precompiles/ics02/ics02.go
srdtrk: How do you do that? Like use [this function](https://github.com/cosmos/evm/blob/81bab01c59b68c655320d95fc09547493c0cbb78/precompiles/common/precompile.go?plain=1#L38)?

precompiles/ics02/ics02.go
srdtrk: done

precompiles/ics02/query.go
srdtrk: I think this way would be slightly faster for the relayer

evmd/tests/ibc/ics02_precompile_client_test.go
srdtrk: I actually checked IBC-Go, and we don't even emit an event in this scenario. I think I should just remove no-op from the abi?

This no-op is only here because it is found in ibc-solidity, not in ibc-go. So I can remove it. Wdyt?

srdtrk: I removed noop for now.

evmd/tests/ibc/ics02_precompile_client_test.go
srdtrk: done

evmd/tests/ibc/ics02_precompile_client_test.go
srdtrk: done

precompiles/ics02/ics02.go
srdtrk: I think I did this, but please verify

evmd/tests/ibc/ics02_precompile_client_test.go
srdtrk: done

docs/migrations/v0.5.0_to_v0.6.0.md
```diff
@@ -0,0 +1,55 @@
+# Cosmos EVM v0.4.0 → v0.5.0 Migration
```
aljo242:
```suggestion
# Cosmos EVM v0.5.0 → v0.6.0 Migration
```

docs/migrations/v0.5.0_to_v0.6.0.md
```diff
@@ -0,0 +1,55 @@
...
+
+- Create a branch: `git switch -c upgrade/evm-v0.6`.
+- Ensure a clean build + tests green pre-upgrade.
+- Snapshot your current params/genesis for comparison later.
+
+---
+
+## 1) Dependency bumps (go.mod)
+
+- Bump `github.com/cosmos/evm` to v0.5.0 and run:
```
aljo242:
```suggestion
- Bump `github.com/cosmos/evm` to v0.6.0 and run:
```

precompiles/ics02/tx.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/ethereum/go-ethereum/core/vm"
+
+	clienttypes "github.com/cosmos/ibc-go/v10/modules/core/02-client/types"
+	commitmenttypesv2 "github.com/cosmos/ibc-go/v10/modules/core/23-commitment/types/v2"
+	host "github.com/cosmos/ibc-go/v10/modules/core/24-host"
+	ibcexported "github.com/cosmos/ibc-go/v10/modules/core/exported"
+
+	errorsmod "cosmossdk.io/errors"
+
+	// codectypes "github.com/cosmos/cosmos-sdk/codec/types"
```
aljo242: remove?

docs/migrations/v0.5.0_to_v0.6.0.md
srdtrk: done

docs/migrations/v0.5.0_to_v0.6.0.md
srdtrk: done

precompiles/ics02/tx.go
srdtrk: done

evmd/tests/ibc/ics02_precompile_client_test.go
vladjdk: let's remove this comment in a future PR

precompiles/ics02/ics02.go
vladjdk: lgtm

vladjdk (approved):

Issue #778, opened 2025-10-29
Investigate FeeHistory performance optimizations


technicallyty:
currently the FeeHistory RPC endpoint will query multiple batches of CometBlocks. This RPC query is to support dynamic fee eip 1559 txs (i think?). we could likely introduce some caching here. it makes multiple RPC calls to comet AND to the SDK's grpc endpoints.

abdulazeem-tk4vr:
I would like to pick this up and get started on it.

abdulazeem-tk4vr:
Ready for review. 
@technicallyty

PR #781, opened 2025-10-30
EVM-373 : Investigate FeeHistory performance optimizations

abdulazeem-tk4vr:
# Description

Optimize eth_feeHistory performance and reduce redundant RPC calls.

Related to :
https://github.com/cosmos/evm/issues/778
https://linear.app/cosmoslabs/issue/EVM-373/investigate-feehistory-performance-optimizations

Key changes:
- Remove duplicate BlockResults fetch per block by replacing GetBlockByNumber with RPCBlockFromCometBlock after fetching Comet BlockResults.
- Add lightweight height-keyed caches:
  - Comet block cache (height → ResultBlock)
  - Comet block results cache (height → ResultBlockResults)
  - FeeMarket params cache (height → Params)
  - Consensus gas limit cache 
- Add simple benchmark scripts to time eth_feeHistory locally.

Most critical files to review:
- evm/rpc/backend/chain_info.go
- evm/rpc/backend/comet.go
- evm/rpc/backend/backend.go
- evm/rpc/backend/utils.go

Flows that originate from FeeHistory and need caching :
- cometBlockCache (ResultBlock)
Flow: FeeHistory → CometBlockByNumber → cache read/fill.
- cometBlockResultsCache (ResultBlockResults)
Flow: FeeHistory → CometBlockResultByNumber → cache read/fill.
- feeParamsCache (FeeMarket params)
Flow: FeeHistory → ProcessBlock → getFeeMarketParamsAtHeight → cache read/fill → CalcBaseFee.
- consensusGasLimitCache (consensus gas limit)
Flow: FeeHistory → RPCBlockFromCometBlock → EthBlockFromCometBlock → BlockMaxGasAtHeight → cache read/fill → header.GasLimit → ProcessBlock uses header.

Notes:
- No behavior change to results; only fewer RPCs and faster repeated queries.
- Debug cache logs were temporarily added for validation and then removed.
- Local testing using performance scripts, makefile unit tests, and result comparison looks good.

Closes: #778

---

## Author Checklist

I have...

- [ ] Left instructions on how to review the changes
  - Verify eth_feeHistory responses (fields/lengths/values) for several fixed heights.
  - Confirm single BlockResults fetch per block in the FeeHistory loop:
    - rpc/backend/chain_info.go: Backend.FeeHistory now uses RPCBlockFromCometBlock(cometBlock, cometBlockResult, true) instead of GetBlockByNumber.
  - Caching changes to validate:
    - FeeMarket params cache:
      - rpc/backend/backend.go: Backend.getFeeMarketParamsAtHeight(height int64) + height-keyed cache on Backend.
    - Comet block/results caches:
      - rpc/backend/comet.go (or rpc/backend/blocks.go, per repo layout): Backend.CometBlockByNumber and Backend.CometBlockResultByNumber read/write height-keyed caches.
    - Consensus gas-limit cache (new):
      - rpc/backend/backend.go: Backend: BlockMaxGasAtHeight (ctx, height) + cached map.
    - Concurrency/bounds:
      - Caches are map+sync.RWMutex on Backend; soft bound ≈ JSONRPC.FeeHistoryCap × 2.
  - Performance check:
    - Run eth_feeHistory twice against the same fixed newest block (not “latest”); the second call should be faster (cache hits).


rpc/backend/comet.go
```diff
@@ -49,11 +66,29 @@ func (b *Backend) CometBlockResultByNumber(height *int64) (*cmtrpctypes.ResultBl
...
+		b.cacheMu.RUnlock()
+	}
 	res, err := b.RPCClient.BlockResults(b.Ctx, height)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch block result from CometBFT %d: %w", *height, err)
 	}
-
+	if height != nil {
+		b.cacheMu.Lock()
+		if cap := int(b.Cfg.JSONRPC.FeeHistoryCap) * 2; cap > 0 && len(b.cometBlockResultsCache) >= cap {
```
technicallyty, line 83: why do we multiply the cap by 2 here?

abdulazeem-tk4vr: tl;dr it gives better hit rates for overlapping requests where max no. of blocks are called for a reasonable cap size.

Well I noticed that FeeHistoryCap limits a single eth_feeHistory call to a default of 100 blocks. So in a possible worst case scenario where a user is indexing the network for details, consecutive requests often overlap because "latest" moves forward. 

FeeHistoryCap = 100
Cache size = 200 entries

Request 1: blocks 1000-1099
Cache stores: blocks 1000-1099 (100 entries)

Request 2 (few seconds later): blocks 1050-1149  
Cache hits: blocks 1050-1099 (50 hits)
Cache stores: blocks 1100-1149 (50 new entries)
Total cached: 150 entries (well within the 200 limit)

abdulazeem-tk4vr: If a user keeps making a request for 100 entries in a random range, we will lose the previous context. Then this isnt that helpful. 

I was just thinking that if there were requests x and y in completely different ranges, y would override the cache with just `FeeHistoryCap`. But if again somehow, x was requested because y was a mistake, then there is a warm cache with `FeeHistoryCap *2`. 

this is more like a safety check, so that we provide some leeway for older entries to be there. I think either way is good enough to get the job done.

abdulazeem-tk4vr: @technicallyty please do let me know what you think about this

technicallyty: i will try to carve out time to look at this this week

abdulazeem-tk4vr: Thanks, I appreciate it. If it makes your job any easier, the existing RPC unit tests are working with the added features. I dont think it breaks anything.

abdulazeem-tk4vr: gentle reminder!

Issue #807, opened 2025-11-06
Problem: un-intuitive semantics with werc20


yihuang:
We reviewed the werc20 precompile in MantraChain. While it has merits, its drawbacks may outweigh the benefits because of the un-intuitive non-standard semantic of it:

* The semantic is not the same as weth contract, because of how balance change, which is risky if developers try to use it as drop-in replacement for weth contract.
* The semantic is not even like any other erc20 tokens, because the balance can automatically changed when native value transfer happens..

Basically, the werc20 token balance can change without any erc20 `transfer` calls, or the native balance can change without any native value transfer calls, these behaviors might be surprising for developers that come from normal EVM compatible chains.

[This section in moonbeem doc](https://docs.moonbeam.network/learn/core-concepts/security/#setting-a-value) describe a concrete security issue related to werc20 behavior.

## Proposal

Since we'll refactoring dynamic erc20 dynamic precompiles, we can remove werc20 along the way, to simplify the system overall.

PR #808, opened 2025-11-06
feat: remove werc20 precompile

yihuang:
# Description

Closes: #807

---

## Author Checklist

I have...


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### DEPENDENCIES
 
+### API-BREAKING
+
+- [\#808](https://github.com/cosmos/evm/pull/808) Remove werc20 precompile.
```
aljo242, line 9: PR lgtm - is there any documentation that we need to add for folks who will be updating their chains?

yihuang: Hmm, by default, if a chain has previously enabled werc20 and upgrade, that precompile will be silently gone, and that address will become like an EOA. If we want to prevent user mistakenly interact with that address, we can deploy a bytecode to that address which always revert for any input.

```solidity
// SPDX-License-Identifier: LGPL-3.0-only

pragma solidity >=0.8.0;

contract Revert {
    fallback() external {
        revert("WERC20 placeholder");
    }
}
```

PR #812, opened 2025-11-07
fix: evm tx index and log indexes are not patched

mmsqe:
# Description

Under BlockSTM, ctx.TxIndex() is unreliable, to patch transactionIndex and log indexes for JSON-RPC

* add `PatchTxResponses` as post-execution pass that assigns sequential txIndex/logIndex and emits the ethereum_tx hash event
* remove unreliable index source from ante/evm/11_emit_event.go and remove unused ResetTransientGasUsed
* add RPC fallback to extract `FindEthTxIndexByHash`, reused by `GetTransactionByHash` and `ReceiptsFromCometBlock` for legacy blocks missing AttributeKeyTxIndex

Closes: #XXXX

---

## Author Checklist

I have...


rpc/backend/comet_to_eth.go
```diff
@@ -273,6 +273,14 @@ func (b *Backend) ReceiptsFromCometBlock(
 			return nil, fmt.Errorf("failed to convert tx result to eth receipt: %w", err)
 		}
 
+		if txResult.EthTxIndex == -1 {
```
aljo242, line 308: can we write a brief comment on why we are doing this?

x/vm/keeper/config.go
```diff
@@ -38,8 +38,8 @@ func (k *Keeper) EVMConfig(ctx sdk.Context, proposerAddress sdk.ConsAddress) (*s
 // TxConfig loads `TxConfig` from current transient storage
 func (k *Keeper) TxConfig(ctx sdk.Context, txHash common.Hash) statedb.TxConfig {
 	return statedb.NewTxConfig(
-		txHash,              // TxHash
-		uint(ctx.TxIndex()), //#nosec G115 // TxIndex
+		txHash, // TxHash
+		0,
```
aljo242: why is this Ok to do? can we document?

rpc/backend/comet_to_eth.go
mmsqe: do you mean add back [this](https://github.com/cosmos/evm/blob/v0.3.2/rpc/backend/tx_info.go#L211) comment?

aljo242: yes

canu0205 (approved):
hey @vladjdk this PR fixes post-blockstm-enabled evm JSON-RPC log index issue #1086. and it looks good to me! @mmsqe thanks for ur work 👍 

will close the #1087 since this only mutates rpc backend log index while current pr actually writes correct log index and tx index.

cc. @cloudgray

x/vm/types/response.go
```diff
@@ -0,0 +1,93 @@
...
+// Note: txIndex starts at 0 and is incremented for each Ethereum transaction found.
+// This ensures proper indexing when multiple EVM transactions are included in a block.
+func PatchTxResponses(input []*abci.ExecTxResult) []*abci.ExecTxResult {
+	var (
+		txIndex  uint64
+		logIndex uint64
+	)
+	for _, res := range input {
+		// assume no error result in msg handler
+		if res.Code != 0 {
```
canu0205, line 24: - `EmitTxHashEvent` used to emit the minimal `ethereum_tx(ethereumTxHash, txIndex)` event from ante, which made expected-failure evm txs still queryable through the fallback comet event index.
- after removing it, `PatchTxResponses` becomes the only source of that event, but it only runs for `ExecTxResult.Code == 0`. as a result, txs that exceed block gas or fail during stateDB commit no longer emit `ethereumTxHash` / `txIndex` in events, so the fallback `GetTxByEthHash` / `GetTxByTxIndex` path cannot find them.
- this also creates an indexing mismatch for later txs: `PatchTxResponses` skips expected-failure txs when assigning `txIndex`, while `EthMsgsFromCometBlock` and the kv indexer still count them as valid ethereum txs in block order. that means a later successful tx can be patched with `txIndex = 0` while the indexer / rpc path treats it as `txIndex = 1`.

canu0205: https://github.com/mmsqe/evm/pull/1#issue-4194055670 addressing bug in here

mmsqe: thanks to point out, will try to reproduce that later

mmsqe: Sorry that I reverted the change after comparing existing behavior, since RPC intentionally exclude non-zero code txs, which are not affect EVM tx/receipt indexing. And block out-of-gas is handled at end of block execution, so prior valid tx indexes are unchanged.

canu0205: - so u mean we usually running evm kv indexer for evm chains and this evm kv indexer does indexes expected failure txs. ([ref](https://github.com/mmsqe/evm/blob/1651ab925357e09893dd3229417b58b234faa906/indexer/kv_indexer.go#L62-L64))
- `PatchTxResponses` affects comet indexer, but comet indexer is the fallback behavior of evm json rpc backend ([ref](https://github.com/cosmos/evm/blob/f5d4de06c32880a16002246f49b526f5961343c8/rpc/backend/tx_info.go#L310-L316))
- therefore, we don't need to deal with expected failure txs in the `PatchTxResponse`

is my understanding correct? @mmsqe

mmsqe: yes, kv indexer is primary path and already indexes expected-failure txs. PatchTxResponses only feeds CometBFT indexer fallback, which doesn't need to handle them. We also need [remove](https://github.com/cosmos/evm/pull/1107/changes) StateDBCommitError as expected failure, since it cause invalid message index errors in block RPCs and was never needed given kv indexer handles it correctly.

canu0205: btw ru guys using evm kv indexer? I think some team might use the comet indexer by default since app.toml sets `enable-indexer = false` ([ref](https://github.com/cosmos/evm/blob/5a7f5358d5d607c407781304219aae34db027d8d/server/config/config.go#L357)). in that case, behavior of the chain running evm indexer and comet indexer diverges, and I think it's not an ideal design.

mmsqe: we enable indexer, should be the same after block gas meter get [disabled](https://github.com/cosmos/evm/blob/main/evmd/app.go#L267)

vladjdk:
@greptile review

x/vm/types/response.go
```diff
@@ -0,0 +1,93 @@
...
+			continue
+		}
+
+		var txMsgData sdk.TxMsgData
+		if err := proto.Unmarshal(res.Data, &txMsgData); err != nil {
+			panic(err)
+		}
```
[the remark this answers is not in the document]
vladjdk: @mmsqe can you take a look at this?

vladjdk:
@greptile review

vladjdk:
Can we add a PR description here, and possibly attach to an issue?

vladjdk:
Closing for #1132

Issue #839, opened 2025-11-17
independent execution failure of precisebank test


cloudgray:
There are some test cases that when selected to run independently, they'll fail, for example the following cmd will fail in main branch:

```
$ cd evmd; go test -v -tags=test ./... -run TestPreciseBankKeeper
```

It's quite confusing when debugging test failures.

ref: https://github.com/cosmos/evm/issues/830

Issue #830, opened 2025-11-13
interdependencies between tests


yihuang:
There are some test cases that when selected to run independently, they'll fail, for example the following cmd will fail in main branch:

```
$ cd evmd; go test -v -tags=test ./... -run TestPreciseBankGenesis
```

It's quite confusing when debugging test failures.

PR #842, opened 2025-11-19
fix(test): avoid mutation of global int256 variable

mmsqe:
# Description

to avoid fail test like https://github.com/cosmos/evm/issues/830 https://github.com/cosmos/evm/issues/839

---

## Author Checklist

I have...

### DIFF
--- a/tests/integration/x/precisebank/test_genesis.go
+++ b/tests/integration/x/precisebank/test_genesis.go
@@ -104,17 +104,7 @@ func (s *GenesisTestSuite) TestInitGenesis() {
 		{
 			"invalid - module balance insufficient",
 			func() {
-				// Ensure module account has 0 balance
-				moduleAddr := s.network.App.GetAccountKeeper().GetModuleAddress(types.ModuleName)
-				existingBalance := s.network.App.GetBankKeeper().GetBalance(s.network.GetContext(), moduleAddr, types.IntegerCoinDenom())
-				if existingBalance.IsPositive() {
-					err := s.network.App.GetBankKeeper().BurnCoins(
-						s.network.GetContext(),
-						types.ModuleName,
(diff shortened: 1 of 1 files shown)


cloudgray:
@mmsqe 
This is quite a shocking discovery.

I can’t believe that `common.U2560` had a non-zero value assigned to it and was causing interference between different test cases…

Thanks to this, we can remove the temporary workarounds in the other PRs (#837, #840) and finally address the root cause!

mmsqe:
> @mmsqe This is quite a shocking discovery.
> 
> I can’t believe that `common.U2560` had a non-zero value assigned to it and was causing interference between different test cases…
> 
> Thanks to this, we can remove the temporary workarounds in the other PRs (#837, #840) and finally address the root cause!

Not sure why temporary, we need both fixes to ensure test isolation and avoid global mutation right?

cloudgray:
> Not sure why temporary, we need both fixes to ensure test isolation and avoid global mutation right?

@mmsqe 
- Even taking test isolation into consideration, this situation seems highly abnormal and exceptional.
- Under normal circumstances, when the app is initialized, the module account balance should be zero, and the very idea of handling a case where it is not zero feels incorrect.
- Moreover, the balance-adjusting workaround is not applied to all accounts — it only applies to the precisebank module account. Since this workaround was introduced solely to deal with the underlying issue of the global variable being overwritten, I believe it is best to roll it back once that root cause is resolved.
- I also think this is preferable because future developers who encounter this balance-adjusting helper function without understanding its context will have a hard time grasping why it exists in the first place.

cloudgray (approved):
LGTM!

FYI, I'll close PR #839 .
Because this pr solved problem clearly so we don't need PR #839  anymore.

> good to merge?
@aljo242 Yes!

PR #856, opened 2025-11-24
feat(mempool): Run AnteHandler on promote/demote executables

mattac21:
# Description

Drop transactions from `pending` and `queue` pools based on `anteHandler` success. 

During `promoteExecutables` and `demoteUnexecutables`, we now executes the app's `anteHandler` sequence to see if a transaction is valid or not. If it is not valid, we remove it from the mempool entirely and demote any subsequent nonce transactions for this account (if the transaction was removed from the pending pool). This essentially replicates an app side `recheckTxs` functionality.

Notes:
* This does not implement ensuring that this execution has completed before `Mempool.Select` runs.
* We may be able to fully remove the logic in `Mempool.Remove` that executes the `anteHandler` again to see if a tx should be remove or not, now that we have the `anteHandler` executing in the promote/demote goroutines. However if the above needs fixing (i.e. `Mempool.Select` can run before these goroutines complete/run), then there would be be a race condition were txs that should have been removed are selected to be in a proposal again. Thus, this behavior will be removed in a later PR when this is verified.
* Unit tests are failing I due to unrelated things in the PR, they seem to just fail for the base `feat/krakato` branch, some combination of OE + virtual bank breaks things.

Closes: #XXXX

---

## Author Checklist

I have...

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -6,22 +6,26 @@ import (
 	"fmt"
 	"sync"
 
-	"cosmossdk.io/log"
-	"cosmossdk.io/math"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/holiman/uint256"
+
 	cmtproto "github.com/cometbft/cometbft/proto/tendermint/types"
 	cmttypes "github.com/cometbft/cometbft/types"
-	"github.com/cosmos/cosmos-sdk/client"
-	sdk "github.com/cosmos/cosmos-sdk/types"
-	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
-	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
+
 	"github.com/cosmos/evm/mempool/miner"
 	"github.com/cosmos/evm/mempool/txpool"
 	"github.com/cosmos/evm/mempool/txpool/legacypool"
 	"github.com/cosmos/evm/rpc/stream"
 	"github.com/cosmos/evm/utils"
 	evmtypes "github.com/cosmos/evm/x/vm/types"
-	ethtypes "github.com/ethereum/go-ethereum/core/types"
-	"github.com/holiman/uint256"
+
+	"cosmossdk.io/log"
+	"cosmossdk.io/math"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
 )
 
 var _ sdkmempool.ExtMempool = &ExperimentalEVMMempool{}


mempool/mempool.go
```diff
@@ -361,6 +379,9 @@ func (m *ExperimentalEVMMempool) shouldRemoveFromEVMPool(tx sdk.Tx) bool {
 		return false // Cannot validate, keep transaction
 	}
 
+	// TODO: We may be able to just fully remove this and never remove from the
+	// evm pool when comet tells us to. Relying only on the anteHandler check
+	// in promote/demote executables to remove txs.
```
mattac21, line 369: @vladjdk for thoughts here

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1402,12 +1404,31 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
+		// check is slower, we would like it to happen on the fewest txs as
+		// possible.
+		//
+		// TODO: We can BlockSTM this by collecting all txs into a list and
+		// then running the ante handlers on them in one parallel batch.
+		// However this requires a small refactor of BlockSTM to remove its
+		// dependency on DeliverTx and allow it to use any fn in parallel.
+		var recheckDrops []*types.Transaction
+		if pool.RecheckTxFn != nil {
+			recheckDrops, _ := list.Filter(func(tx *types.Transaction) bool { return pool.RecheckTxFn(tx) != nil })
```
mattac21: ignoring the invalidated txs because the queued pool is not `strict` and no invalids will be returned.

mempool/mempool.go
```diff
@@ -134,6 +134,40 @@ func NewExperimentalEVMMempool(
...
+
+		ctx, err := bc.GetLatestContext()
+		if err != nil {
+			return fmt.Errorf("failed to get latest context: %w", err)
+		}
+
+		_, err = config.AnteHandler(ctx, cosmosTx, false)
+		if errors.Is(err, ErrNonceGap) ||
+			errors.Is(err, sdkerrors.ErrInvalidSequence) ||
+			errors.Is(err, sdkerrors.ErrOutOfGas) ||
```
mattac21: This is the same set of errors that we ignore in [`shouldRemoveFromTxPool`](https://github.com/cosmos/evm/blob/f1f4c2aee76243f0ffe0dd444e05b9cb2ef9898b/mempool/mempool.go#L366). 

For example we should not remove txs with nonce gaps, the evm mempool handles that and nonce gaps are allowed (same for invalid sequence, I dont 100% understand why we ignore this one since it feels cosmos specific and these are evm txs, but it pops up in the same situations).

OutOfGas is ignored since I think this is a quirk of the context that is passed into the ante handler. We sometimes see the block gas as 0.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1900,3 +1949,14 @@ func (pool *LegacyPool) Clear() {
 func (pool *LegacyPool) HasPendingAuth(addr common.Address) bool {
 	return pool.all.hasAuth(addr)
 }
+
+func (pool *LegacyPool) CurrentCacheContext() sdk.Context {
```
mattac21: Pulling the latest context from the pools `currentState` `vm.StateDB`. I was previously getting this from the pools `blockchain` instance, however there is actually a mismatch between the two states which I didnt expect.

When using the `blockchain` context, I would sometimes see `nonce too low` errors during `demoteUnexecutables` (so txs in the pending pool), before they were actually committed according to this node (nonce 1 in the mempool, account nonce would go to 2 before we try and remove nonce 1 tx from pending pool via Remove). I think this is a race condition between when the individual contexts are updated in relation to removing txs after the are committed.

The pools `currentState` always has the correct nonce though from the mempools point of view, so we simply opt to use that (via a cast since the evm interface doesnt expose a method to fetch the cosmos context, obviously).

mempool/mempool.go
```diff
@@ -134,6 +134,29 @@ func NewExperimentalEVMMempool(
...
+		txBuilder := txConfig.NewTxBuilder()
+		cosmosTx, err := msg.BuildTx(txBuilder, evmtypes.GetEVMCoinDenom())
+		if err != nil {
+			return fmt.Errorf("failed to build cosmos tx from evm tx: %w", err)
+		}
+
+		_, err = config.AnteHandler(ctx, cosmosTx, false)
+		if errors.Is(err, ErrNonceGap) ||
+			errors.Is(err, sdkerrors.ErrInvalidSequence) ||
+			errors.Is(err, sdkerrors.ErrOutOfGas) {
```
mattac21: This is the same set of errors that we ignore in shouldRemoveFromTxPool.

For example we should not remove txs with nonce gaps, the evm mempool handles that and nonce gaps are allowed (same for invalid sequence, I dont 100% understand why we ignore this one since it feels cosmos specific and these are evm txs, but including it here for completeness).

OutOfGas is ignored since it seems to be a quirk of the mempools chain context, it never has the maxblockgas set.

tests/systemtests/mempool/test_replacement.go
```diff
@@ -261,17 +261,17 @@ func RunMixedTxsReplacementEVMAndCosmos(t *testing.T, base *suite.BaseTestSuite)
...
-				func(s *TestSuite, ctx *TestContext) {
-					_, err := s.SendCosmosTx(t, s.Node(0), "acc0", 0, s.BaseFeeMultiplier(10), nil)
-					require.NoError(t, err, "failed to send tx")
-					tx2, err := s.SendEthTx(t, s.Node(1), "acc0", 0, s.BaseFeeMultiplier(20), s.BaseFeeMultiplier(20))
-					require.NoError(t, err, "failed to send tx")
-
-					ctx.SetExpPendingTxs(tx2)
-				},
-			},
+			// name: "single pending tx (low prio cosmos tx first) %s",
```
mattac21, line 264: this test was causing non determinism, asked vlad and these tests have been problematic in the past, the strictly evm tx tests seem fine, but this cosmos + evm was causing issues. I debugged for quite a while but figured it was likely best to move on and come back to this later if we see issues, since we dont actually care about the cosmos tx use case at the moment.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1900,3 +1949,18 @@ func (pool *LegacyPool) Clear() {
...
+// context.
+func (pool *LegacyPool) CurrentCacheContext() sdk.Context {
+	db, ok := pool.currentState.(*statedb.StateDB)
+	if !ok {
+		// should never happen, not panicing for testing
+		return sdk.Context{}
+	}
+
+	currentContext := db.GetContext()
+	cacheCtx, _ := currentContext.CacheContext()
```
mattac21: would like confirmation that getting a cache context is the right thing to do here. I was basically just copying what is happening [here](https://github.com/cosmos/evm/blob/feat/pending-pool-recheck/mempool/blockchain.go#L219), creating a context at the latest mempool state for ante handlers to run on but wont interfere with anything else.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1402,12 +1406,33 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
+		// dependency on DeliverTx and allow it to use any fn in parallel.
+		var recheckDrops []*types.Transaction
+		if pool.RecheckTxFn != nil {
+			recheckDrops, _ = list.Filter(func(tx *types.Transaction) bool {
+				return pool.RecheckTxFn(pool.CurrentCacheContext(), tx) != nil
+			})
+			for _, tx := range recheckDrops {
+				pool.all.Remove(tx.Hash())
+			}
+			log.Trace("Removed queued transactions that failed recheck", "count", len(recheckDrops))
```
Eric-Warehime, line 1444: We should also add a metric here around this value.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1590,22 +1615,46 @@ func (pool *LegacyPool) demoteUnexecutables() {
...
+		var recheckInvalids []*types.Transaction
+		var recheckDrops []*types.Transaction
+		if pool.RecheckTxFn != nil {
+			recheckDrops, recheckInvalids = list.Filter(func(tx *types.Transaction) bool {
+				return pool.RecheckTxFn(pool.CurrentCacheContext(), tx) != nil
+			})
+			for _, tx := range recheckDrops {
+				hash := tx.Hash()
+				pool.all.Remove(hash)
+				log.Trace("Removed pending transaction that failed recheck", "hash", hash)
```
Eric-Warehime, line 1661: I guess maybe a new metric here as well.

mempool/txpool/legacypool/legacypool.go
Eric-Warehime: It seems fine to get the query context this way. The only thing I think we might want to change is keeping it in memory and re-using it for each transaction until the blockchain's notify new block resets it.

As is, we're requesting a new query context for every transaction we're validating in both pending and queued. Whereas checktx currently operates serially on a single context which is reset in Commit.

mempool/txpool/legacypool/list.go
```diff
@@ -362,19 +362,23 @@ func (l *list) Forward(threshold uint64) types.Transactions {
 // a point in calculating all the costs or if the balance covers all. If the threshold
 // is lower than the costgas cap, the caps will be reset to a new high after removing
 // the newly invalidated transactions.
-func (l *list) Filter(costLimit *uint256.Int, gasLimit uint64) (types.Transactions, types.Transactions) {
+func (l *list) CostFilter(costLimit *uint256.Int, gasLimit uint64) (types.Transactions, types.Transactions) {
```
swift1337: nit: update fn comment `Filter ... → CostFilter ...`

mempool/txpool/legacypool/list.go
```diff
@@ -362,19 +362,23 @@ func (l *list) Forward(threshold uint64) types.Transactions {
...
-	// Filter out all the transactions above the account's funds
-	removed := l.txs.Filter(func(tx *types.Transaction) bool {
+	return l.Filter(func(tx *types.Transaction) bool {
 		return tx.Gas() > gasLimit || tx.Cost().Cmp(costLimit.ToBig()) > 0
 	})
+}
 
+// Filter filters txs in the list by a filter function. If the filter rn
+// returns false for a tx, it is removed.
+func (l *list) Filter(filterFn func(tx *types.Transaction) bool) ([]*types.Transaction, []*types.Transaction) {
```
swift1337:
```suggestion
func (l *list) Filter(filterFn func(tx *types.Transaction) bool) (removed []*types.Transaction, invalids []*types.Transaction) {
```

mempool/txpool/legacypool/list.go
swift1337: also, why does return type differ from `[]*types.Transaction`?

swift1337:
```suggestion
func (l *list) CostFilter(costLimit *uint256.Int, gasLimit uint64) (removed types.Transactions, invalids types.Transactions) {
```

mempool/txpool/legacypool/legacypool.go
```diff
@@ -256,6 +258,8 @@ type LegacyPool struct {
 	changesSinceReorg int // A counter for how many drops we've performed in-between reorg.
 
 	BroadcastTxFn func(txs []*types.Transaction) error
+
+	RecheckTxFn func(ctx sdk.Context, t *types.Transaction) error // Called on every tx during promoteExecutables and demoteExecutables, removes based on failure
```
swift1337: nit: move comment on top

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1402,12 +1406,33 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 		}
 		log.Trace("Removed old queued transactions", "count", len(forwards))
 		// Drop all transactions that are too costly (low balance or out of gas)
-		drops, _ := list.Filter(pool.currentState.GetBalance(addr), gasLimit)
-		for _, tx := range drops {
+		costDrops, _ := list.CostFilter(pool.currentState.GetBalance(addr), gasLimit)
+		for _, tx := range costDrops {
```
swift1337, line 1418: Q: Is the order guaranteed for iterating dropped txs? Do we care about determinism in the mempool? Same Q for `range recheckDrops`

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1429,8 +1454,8 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 		}
 		queuedRateLimitMeter.Mark(int64(len(caps)))
 		// Mark all the items dropped as removed
-		pool.priced.Removed(len(forwards) + len(drops) + len(caps))
-		queuedGauge.Dec(int64(len(forwards) + len(drops) + len(caps)))
+		pool.priced.Removed(len(forwards) + len(costDrops) + len(recheckDrops) + len(caps))
```
swift1337: nit: let's create smth like

`totalRemoved := len(forwards) + len(costDrops) + len(recheckDrops) + len(caps)`

mempool/txpool/legacypool/legacypool.go
swift1337: > re-using it for each transaction

I agree. I'd do anything that works for PoC, but later we can sub to new block events and perform context updates async

https://github.com/cosmos/evm/blob/f1f4c2aee76243f0ffe0dd444e05b9cb2ef9898b/mempool/mempool.go#L398

mempool/mempool.go
swift1337: nit: consider moving to smth like `isErrIgnoreAnte(err)`

mempool/txpool/legacypool/legacypool.go
mattac21: added metrics here https://github.com/cosmos/evm/pull/856/commits/31eb4efb1980f0de25e6220e9870d6606d03b6c4

mempool/txpool/legacypool/legacypool.go
mattac21: added metrics here https://github.com/cosmos/evm/pull/856/commits/31eb4efb1980f0de25e6220e9870d6606d03b6c4

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1900,3 +1949,18 @@ func (pool *LegacyPool) Clear() {
...
 }
+
+// CurrentCacheContext gets a cached version of the current context in the
+// pools stateDB. Note that this may momentarily differ from the BlockChain's
+// context.
+func (pool *LegacyPool) CurrentCacheContext() sdk.Context {
+	db, ok := pool.currentState.(*statedb.StateDB)
+	if !ok {
+		// should never happen, not panicing for testing
+		return sdk.Context{}
```
swift1337: I don't have much context, but this looks a bit brittle

swift1337 (commented):
Left minor comments. Looking forward to integrate this w/ comet!

mempool/txpool/legacypool/legacypool.go
mattac21: agree on top is the right way to do things but the convention of the file is having the comments to the right

mempool/txpool/legacypool/legacypool.go
mattac21: good callouts updated here https://github.com/cosmos/evm/pull/856/commits/2d1ce09edc0ad4deb024fc9778b75c37a6462511

mattac21: the reorg loop should actually always have the latest blocks context in the `pool.currentstate` var. when a new block arrives the reorg loop is called and `pool.currentstate` is [updated](https://github.com/cosmos/evm/blob/feat/pending-pool-recheck/mempool/txpool/legacypool/legacypool.go#L1373) before promote/demote is called [here](https://github.com/cosmos/evm/blob/feat/pending-pool-recheck/mempool/txpool/legacypool/legacypool.go#L1297)

mempool/txpool/legacypool/list.go
mattac21: dont think there is a good reason, `types.Transactions` is just a wrapper around `[]*types.Transaction`, Ill updated to use the wrapper that seems to be the convention

mempool/mempool.go
mattac21: good callout, also moved creating the rechecktxfn into its own helper function, similar to the broadcastevmtransaction function.

mempool/txpool/legacypool/legacypool.go
mattac21: yea I dont love this either, I really wanted to panic here, but couldnt figure out a good way to do this without having to setup a whole valid state db in a test and didnt want to do that lol.

mattac21: the only place the legacypool's state is modified is [here](https://github.com/cosmos/evm/blob/feat/pending-pool-recheck/mempool/blockchain.go#L200) by the cosmos/evm's version of a `blockchain`, which always returns the an instance of the type we are expecting (expect on state 0, which I just added an extra guard for)

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1402,12 +1411,37 @@
...
+		// check is slower, we would like it to happen on the fewest txs as
+		// possible.
+		//
+		// TODO: We can BlockSTM this by collecting all txs into a list and
+		// then running the ante handlers on them in one parallel batch.
+		// However this requires a small refactor of BlockSTM to remove its
+		// dependency on DeliverTx and allow it to use any fn in parallel.
+		var recheckDrops []*types.Transaction
+		if pool.RecheckTxFn != nil {
+			recheckStart := time.Now()
```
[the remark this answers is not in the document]
mattac21: only for a metric

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1590,22 +1624,51 @@
...
+		// possible.
+		//
+		// TODO: We can BlockSTM this by collecting all txs into a list and
+		// then running the ante handlers on them in one parallel batch.
+		// However this requires a small refactor of BlockSTM to remove its
+		// dependency on DeliverTx and allow it to use any fn in parallel.
+		var recheckInvalids []*types.Transaction
+		var recheckDrops []*types.Transaction
+		if pool.RecheckTxFn != nil {
+			recheckStart := time.Now()
```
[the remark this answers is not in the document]
mattac21: only for a metric

mempool/txpool/legacypool/legacypool.go
mattac21: good question, I believe the order should be guaranteed here. `costDrops` and `recheckDrops` are both lists not maps. After giving it a bit of thought I think even if these were maps I dont think it would matter, since we are just removing from the `all` lookup here, we only care that at the end of the loop all of the txs are properly deleted, but it shouldn't have any side effects based on the order of deletion from what I can tell (although it is in the same order, im just thinking if this were a map).

mempool/mempool.go
vladjdk: This is likely something I forgot to remove after creating the new error for nonce gaps, but we should test before getting rid of it

mempool/mempool.go
vladjdk: Ideally we handle this in the mempool routines, yes. As long as we're skimming through the list of transactions and checking them all frequently enough, we should be fine to bounce all Remove requests from Comet

tests/systemtests/mempool/test_replacement.go
vladjdk: What's most likely happening is these transactions get committed before we're able to check if they're in pending, so the test is failing.

A todo here would be to change this to pass on tx commited OR in pending list.

edit: made issue: https://github.com/cosmos/evm/issues/864

mempool/mempool.go
```diff
@@ -505,3 +510,37 @@ func broadcastEVMTransactions(clientCtx client.Context, txConfig client.TxConfig
...
+
+		_, err = anteHandler(ctx, cosmosTx, false)
+		return dropAnteNonceErr(err)
+	}
+}
+
+// dropAnteNonceErr returns nil if err is considered an error that should be
+// ignored from the anteHandlers in the context of the recheckTxFn. If the
+// error should not be ignored, it is returned unmodified.
+func dropAnteNonceErr(err error) error {
```
```suggestion
func tolerateAnteErr(err error) error {
```

swift1337 (approved):

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1900,3 +1967,23 @@ func (pool *LegacyPool) Clear() {
 func (pool *LegacyPool) HasPendingAuth(addr common.Address) bool {
 	return pool.all.hasAuth(addr)
 }
+
+// CurrentCacheContext gets a cached version of the current context in the
+// pools stateDB. Note that this may momentarily differ from the BlockChain's
+// context.
+func (pool *LegacyPool) CurrentCacheContext() sdk.Context {
```
vladjdk: I don't really love introducing SDK types into the geth code as this makes it harder to maintain. 

Is there a way where we can cache the current query context in mempool.go and use that inside the callback code instead of passing it in as a parameter from legacypool?

mempool/txpool/legacypool/legacypool.go
vladjdk: Would be nice to have a NewDefaultStateDB() function for this purpose

mempool/txpool/legacypool/legacypool.go
vladjdk: We determined this to be because the antehandlers actually update state, so unless you reset the context between antehandler checks, you'll start getting `ErrNonceLow` errors each time you run them through. What would happen if you used the blockchain context and cached it every time you needed to run it through the `RecheckTx` instead of using it as is? Or is the blockchain context already changed by the time you use it for the first time in `demoteUnexecutables`?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1402,12 +1413,37 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 		}
 		log.Trace("Removed old queued transactions", "count", len(forwards))
 		// Drop all transactions that are too costly (low balance or out of gas)
-		drops, _ := list.Filter(pool.currentState.GetBalance(addr), gasLimit)
-		for _, tx := range drops {
+		costDrops, _ := list.CostFilter(pool.currentState.GetBalance(addr), gasLimit)
```
vladjdk, line 1417: I wonder if the separate cost filter is still needed since we technically have one in the antehandler via a combination of https://github.com/cosmos/evm/blob/befde4fafa15c469a181708043224c95dea66c52/ante/evm/07_can_transfer.go, https://github.com/cosmos/evm/blob/61eedfcf340e8140b82e9e02478b086b81f63eda/ante/evm/08_gas_consume.go, and https://github.com/cosmos/evm/blob/491aa78f515250590ab235a865233418f387c288/ante/evm/fee_checker.go

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1590,22 +1628,51 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			log.Trace("Removed old pending transaction", "hash", hash)
 		}
 		// Drop all transactions that are too costly (low balance or out of gas), and queue any invalids back for later
-		drops, invalids := list.Filter(pool.currentState.GetBalance(addr), gasLimit)
+		drops, costInvalids := list.CostFilter(pool.currentState.GetBalance(addr), gasLimit)
```
vladjdk, line 1632: Ditto, could we just remove this and use the recheck?

Would also be worth noting that it seems we do a lot of duplicate checks, even in the antehandler, so we should track removing those in a separate issue.

mempool/mempool.go
```diff
@@ -505,3 +510,37 @@ func broadcastEVMTransactions(clientCtx client.Context, txConfig client.TxConfig
...
+	if errors.Is(err, ErrNonceGap) ||
+		errors.Is(err, sdkerrors.ErrInvalidSequence) ||
```
vladjdk: I truly believe we can get rid of this after changing the err name in https://github.com/cosmos/evm/pull/591/commits/0d945fb4760e5d325aace30c03bcf40e74fcbfa1

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+	if pool.queue[from].Len() != 3 {
+		t.Errorf("queued transaction count mismatch: have %d, want %d", pool.queue[from].Len(), 3)
+	}
+	if pool.all.Count() != 3 {
+		t.Errorf("total transaction count mismatch: have %d, want %d", pool.all.Count(), 3)
+	}
+	pool.mu.RUnlock()
+
+	// Set up RecheckTxFn to fail tx1
+	pool.RecheckTxFn = func(_ sdk.Context, tx *types.Transaction) error {
```
vladjdk: lol nice benefit of using callbacks

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+	// Reset will always try and promote executables from queued to go to
+	// pending. This will call pool.RecheckTxFn and block until the Reset is
+	// done.
+	pool.Reset(nil, nil)
+
+	// Verify tx1 was removed from all pools (failed recheck) and tx2 stayed in
+	// queued, and tx0 got promoted to pending
+	pool.mu.RLock()
+	if pool.pending[from] == nil || pool.pending[from].Len() != 1 || pool.pending[from].txs.Get(tx0.Nonce()) == nil {
+		t.Errorf("pending pool should have only tx1 included")
```
vladjdk, line 2727:
```suggestion
		t.Errorf("pending pool should have only tx0 included")
```

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+		if tx.Nonce() == 1 {
+			return errors.New("recheck failed for tx1")
+		}
+		return nil
+	}
+
+	// Reset will always try and promote executables from queued to go to
+	// pending. This will call pool.RecheckTxFn and block until the Reset is
+	// done.
+	pool.Reset(nil, nil)
```
vladjdk, line 2721: The way we know that this is running thru promote and not demote is because we're not adding to the pool, but instead using enqueue?

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+	tx12, tx22 := transaction(2, 100000, key1), transaction(2, 100000, key2)
+
+	from1, _ := deriveSender(tx10)
+	from2, _ := deriveSender(tx20)
+	testAddBalance(pool, from1, big.NewInt(100000000000000))
+	testAddBalance(pool, from2, big.NewInt(100000000000000))
+
+	// Add all transactions to the pool, they should all go to queued, then
+	// this will internally call requestPromoteExecutables
+	pool.addRemoteSync(tx10)
```
vladjdk, line 2771: ditto, same question

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+	// Trigger demoteUnexecutables via Reset
+	pool.Reset(nil, nil)
+
+	// Verify tx10 and tx22 were removed from all pools (failed recheck).
+	//
+	// Since the pending pool is strict, all txs after the removed txs will
+	// also be invalidated and dropped back down to the queued pool
+	pool.mu.RLock()
+	if pool.all.Get(tx10.Hash()) != nil {
+		t.Errorf("tx1 should be removed from all pools after failing recheck")
```
vladjdk:
```suggestion
		t.Errorf("tx10 should be removed from all pools after failing recheck")
```

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+	}
+	if pool.all.Get(tx22.Hash()) != nil {
+		t.Errorf("tx1 should be removed from all pools after failing recheck")
```
vladjdk:
```suggestion
		t.Errorf("tx22 should be removed from all pools after failing recheck")
```

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -2668,6 +2670,178 @@ func TestRemoveTxTruncatePoolRace(t *testing.T) {
...
+		t.Errorf("from2 should have no queued transactions")
+	}
+
+	dropped := pendingRecheckDropMeter.Snapshot().Count()
+	if dropped != 2 {
+		t.Error("2 pending recheck drops should have been recorded by meter, got", dropped)
+	}
+	invaliated := pendingRecheckInvalidateMeter.Snapshot().Count()
+	if invaliated != 2 {
+		t.Error("2 pending recheck invalidate should have been recorded by meter, got", invaliated)
```
vladjdk, line 2847: so invalidated here would be tx11 and tx12?

mempool/mempool.go
mattac21: ty ill create a ticket to do this

mempool/txpool/legacypool/legacypool.go
mattac21: I think using the blockchain context and creating the cache like im doing here would also work, do you have a preference on which way to do it?

mempool/mempool.go
mattac21: you're talking about `ErrInvalidSequence`? I've never seen it pop up in system test, so i'll remove it and try it out

mempool/txpool/legacypool/legacypool.go
mattac21: yeah I realzied they are validating the same thing as well, but my thought was ante handlers are more expensive to execute, so I wanted to leave these in to remove as many txs as possible before running ante handlers. Maybe a premature optimization since I actually have no idea how this would affect things in practice (if at all, maybe the extra iteration is more expensive), but that was the thought.

mempool/txpool/legacypool/legacypool_test.go
mattac21: this will technically run a demote as well first, but since the pending pool is empty that wont do anything, and then promote will run and promote things since they are in queued (since we used enqueue)

mempool/txpool/legacypool/legacypool_test.go
mattac21: all txns will go into queued at first, but with this it will also request for promote executables to be run after every insert in queued. we haven't setup the recheckfn to do anything yet, so they will all successfully promote into pending. then we set the recheckfn to fail and call pool.Reset to trigger the reorg loop to call demote + promote

mempool/txpool/legacypool/legacypool_test.go
mattac21: exactly yeah, ill add a comment

mempool/txpool/legacypool/legacypool_test.go
mattac21: copy paste failed me

mempool/txpool/legacypool/legacypool.go
mattac21: yea agree let me try and play around with it

mempool/mempool.go
mattac21: removed

mempool/txpool/legacypool/legacypool.go
mattac21: done here, also changed back to using the blockchain context and caching it, seems to work well https://github.com/cosmos/evm/pull/856/commits/41aef5384a562c66ee6cdb0cc02fb24b00c7ea70

mempool/txpool/legacypool/legacypool.go
mattac21: same answer as above

mempool/mempool.go
vladjdk: yeah

mempool/txpool/legacypool/legacypool.go
vladjdk: Made an issue to check this out in the future: #861

mempool/txpool/legacypool/legacypool.go
vladjdk: Hmm, this seems to now cache it once per transaction, undoing [Eric's suggestion](https://github.com/cosmos/evm/pull/856#discussion_r2562755047). Is there a way we can keep this outside of legacypool but still cache it once per loop?

mattac21: I have created a monster lol https://github.com/cosmos/evm/pull/856/commits/429eaffa0027a0a1446d8a61e16b1207f20e0620

mattac21: dont love this but also dont hate it, lmk what you think. it feels ok to me for a poc, but probably want some refactoring later im sure there's a better way to do this. I played around with caching the ctx on the evm mempool and the legacy pool would pass a flag into the recheck tx to tell it to update the cache, but that felt ugly where it would pass the flag to reset only on the first invocation of the loop. also played around with the legacy pool having another callback attached it to signal to the evm mempool to update the cache but that felt like the legacy pool knowing a weird impl detail of the recheckfn which i didnt love.

vladjdk: Yeah it's kind of ugly but I can't think of a better way to do it right now, so lgtm!

mempool/txpool/legacypool/legacypool.go
vladjdk: We fixed this in the other comments!

PR #863, opened 2025-11-26
feat(rpc): add OpenTelemetry tracing

technicallyty:
# Description

adds otel tracers to the RPC package.

i also had to update tests because of the context changes wrt mocks. We were using non-typesafe mocks, so i updated those to use typesafe ones. now updating function signatures will also cause mocks to fail. 

---

## Author Checklist

I have...

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,17 +3,17 @@ src="repo_header.svg"
 alt="Cosmos EVM - A plug-and-play solution that adds EVM compatibility and customizability to your chain"
 />
 
+## What is Cosmos EVM?
+
+Cosmos EVM is a plug-and-play solution that adds EVM compatibility and customizability to your Cosmos SDK chain. Cosmos EVM is used by Ondo, Mezo, Mantra, XRP sidechain, Telegram Application Chain (TAC), Stable, and others. Cosmos EVM equips Cosmos chains with complete Ethereum capabilities: Solidity smart contracts, Ethereum JSON-RPC, native support for the EVM wallet/token/user experience, and access to the entire Ethereum developer ecosystem. Its precompiles and extensions allow developers to leverage modules like [IBC](https://github.com/cosmos/ibc-go) with EVM and get native ERC-20 support for tokens on Cosmos. 
+
+Cosmos EVM is customizable for your business use case, chain architecture, and performance needs.
+
 **Please note**: This repo is undergoing changes while the code is being audited and tested. For the time being we will
 be making v0.x releases. Some breaking changes might occur. Cosmos Labs will only mark the Cosmos EVM repository as stable with a v1
 release after the audit, key stability features and benchmarking are completed.
 
 **Visit the official documentation for Cosmos EVM**: [evm.cosmos.network](https://evm.cosmos.network/)
-
-## What is Cosmos EVM?
-
-Cosmos EVM is a plug-and-play solution that adds EVM compatibility and customizability to your Cosmos SDK chain. Cosmos EVM equips Cosmos chains with complete Ethereum capabilities: Solidity smart contracts, Ethereum JSON-RPC, native support for the EVM wallet/token/user experience, and access to the entire Ethereum developer ecosystem. Its precompiles and extensions allow developers to leverage modules like [IBC](https://github.com/cosmos/ibc-go) with EVM and get native ERC-20 support for tokens on Cosmos. 
-
-Cosmos EVM is customizable for your business use case, chain architecture, and performance needs.
 
 
 ## Integration


rpc/backend/tracer.go
```diff
@@ -0,0 +1,5 @@
+package backend
+
+import "go.opentelemetry.io/otel"
+
+var tracer = otel.Tracer("evm/rpc/backend")
```
swift1337: let's drop this file and move this to `backend.go`

rpc/backend/account_info.go
```diff
@@ -23,17 +26,22 @@ import (
 )
 
 // GetCode returns the contract code at the given address and block number.
-func (b *Backend) GetCode(address common.Address, blockNrOrHash rpctypes.BlockNumberOrHash) (hexutil.Bytes, error) {
-	blockNum, err := b.BlockNumberFromComet(blockNrOrHash)
+func (b *Backend) GetCode(ctx context.Context, address common.Address, blockNrOrHash rpctypes.BlockNumberOrHash) (bz hexutil.Bytes, err error) {
+	ctx, span := tracer.Start(ctx, "GetCode", trace.WithAttributes(attribute.String("address", address.String()), attribute.String("blockNorHash", unwrapBlockNOrHash(blockNrOrHash))))
```
swift1337, line 31: I see a lot of repeated code that can be simplified with a small wrapper. Consider the following:

## Creating span for existing ctx:
```go
// wrapper would be located here:
// import https://github.com/cosmos/evm/trace

// func Start(ctx context.Context, name string, attr... Attribute)
//   trace.Str supports string or Stringable
//   trace.Block automatically parses block or hash
//   ... other cosmos/evm specific attrs

ctx, span := trace.Start(ctx, "GetCode", trace.Str("address", address), trace.Block("block", blockNrOrHash))

// regular span end
defer span.End()

//  captures err if not nil
defer span.EndErr(err)

File: rpc/namespaces/ethereum/debug/api.go [hunk 3]
```diff
@@ -29,6 +33,8 @@ import (
 	"github.com/cosmos/cosmos-sdk/server"
 )
 
+var tracer = otel.Tracer("evm/rpc/namespaces/ethereum/debug")
```
swift1337, line 37: same approach:
```go
// import github.com/cosmos/evm/trace
var tracer = trace.New("evm/rpc/namespaces/ethereum/debug")
```

rpc/backend/account_info.go
swift1337: ## Creating span for new ctx:

```go
ctx, span := trace.StartNew("GetCode", attrs...)
defer span.End()
```

swift1337:
Curious, where do we configure tracing sampling rate here?

rpc/backend/account_info.go
technicallyty: i like the idea of `span.EndErr`, doing all the cleanup in one go, i could add this. for the other bits of the wrapper, im indifferent. having to do context.Background for a new span is pretty rare (geth API design weirdness). limiting the wrapper to only attributes could be a footgun later down the line if more `StartSpanOptions` are added that we may want to utilize. currently there are 2 (WithAttributes, WithTimestamp). if you feel pretty strongly though, i'm open to discussing more

aljo242: i like the End() and EndErr()

rpc/backend/account_info.go
```diff
@@ -70,8 +81,8 @@ func (b *Backend) GetProof(address common.Address, storageKeys []string, blockNr
 		height = int64(bn) //#nosec G115 -- checked for int overflow already
 	}
 
-	ctx := rpctypes.ContextWithHeight(height)
-	clientCtx := b.ClientCtx.WithHeight(height)
+	ctx = rpctypes.ContextWithHeight(height, ctx)
```
aljo242: this argument ordering is weird

rpc/types/block.go
```diff
@@ -51,12 +51,19 @@ func NewBlockNumber(n *big.Int) BlockNumber {
 // ContextWithHeight wraps a context with the a gRPC block height header. If the provided height is
 // 0, it will return an empty context and the gRPC query will use the latest block height for querying.
 // Note that all metadata is processed and removed by the CometBFT layer, so it won't be accessible at gRPC server level.
-func ContextWithHeight(height int64) context.Context {
+func ContextWithHeight(height int64, ctxs ...context.Context) context.Context {
```
aljo242: i think it is ok to break here instead of this

rpc/backend/account_info.go
aljo242: might be good to make a trace wrapper in the sdk?

rpc/types/block.go
technicallyty: [make the context height thing cleaner](https://github.com/cosmos/evm/pull/863/commits/aa145cae20a05116c54e514cd876672755c236a1)

rpc/backend/account_info.go
technicallyty: well span.End is already a thing, and adding `EndErr` would require me to create a wrapper type around the span.

i have already made a change to make ending a span with errors cleaner: [add span EndSpanErr helper function](https://github.com/cosmos/evm/pull/863/commits/ced7b3765d5b65559f36753c5d14cfd5f0824c8b)

technicallyty:
TODO: point SDK dep to main

aljo242 (approved):

PR #869, opened 2025-12-01
fix(erc20): ibc callbacks logic reordering

reecepbcups:
# Description

BUG FIX: moved the parsing of the recipient to be after the check if it is the native token transfer for no-op in x/erc20

If the token is in the `bondDenom, err := k.stakingKeeper.BondDenom`, then we can just pass to the next middleware (using packet forward middleware under the hood in my repo). 

---

## Author Checklist

I have...


reecepbcups:
handled

PR #872, opened 2025-12-02
feat(mempool): InsertTx and ReapTxs handlers

mattac21:
# Description

Adds handlers for `ReapTxs` and `InsertTx`. 

This slightly differs from the original design as it does not have a separate queue to process txs received from `InsertTx` before validating them via a goroutine and adding to the mempool. This queue was serving the exact same purpose as the `queue` mempool in the `legacypool`, so I opted to use that as the holding area for invalid txs (which is what it already is). A tx is considered validated once it has been promoted from the `queue` pool to the `pending` pool. When this happens a tx will now be added to a `reapList` (via the `OnPromoteTx` callback on the `legacypool`) that manages the next set of txs to be returned from the next call of `ReapTxs`.

An edge case here is that it is possible for a tx to be promoted multiple times from the `queue` pool to the `pending` pool. Thus we have to be careful and include a `reapGuard` in the `ExperimentalEVMMempool` to protect against having the same tx added to the `reapList` after already being returned. 

A tx is only removed from the `reapList` if it is returned via a call to `ReapTxs` (in this case we do _not_ remove from the `reapGuard` since it may still be in the mempool and be promoted again), or it is dropped from the mempool entirely (in this case we do remove from the `reapGuard` since the tx is no longer in the mempool and it will not be promoted again). 

I opted for the simplest implementation possible for the `reapList`. I expect for there to be some performance issues with it at scale and there are very obvious optimizations we can make (a doubly linked list instead of a slice is probably a very simple optimization to speed up the common `Insert` case, but slow down the less common `Drop` case. If we change to a linked list we can eliminate most of the locking between `Reap` and `Insert`. Caching encoded tx bytes so we dont do it on Reap, etc), but I simply wanted to keep it as minimal as possible before over optimizing.

For cosmos txs, we take a much simpler approach, we call CheckTx (anteHandlers) in the hot path before inserting on the check state (so this will break if we run checktx to eventually get to our insert call as well because of the double check). Then add to the reap list after successful insertion into the cosmos mempool, and remove from the reap list when `Remove` is called for cosmos txs. There is a slight issue here with the priority nonce mempool since txs can be dropped without `Remove` being called if they are replaced by a higher priority nonce tx, but we are choosing to ignore that for now.

Closes: #XXXX

---

## Author Checklist

I have...

### DIFF
--- a/mempool/reap_list_test.go
+++ b/mempool/reap_list_test.go
@@ -30,7 +30,7 @@ type mockEncoder struct {
 	failCosmosHashes map[string]bool
 }
 
-func (m *mockEncoder) EncodeEVMTx(tx *types.Transaction) ([]byte, error) {
+func (m *mockEncoder) EVMTx(tx *types.Transaction) ([]byte, error) {
 	if m.failEVMNonces != nil && m.failEVMNonces[tx.Nonce()] {
 		return nil, errors.New("encoding failed")
 	}
@@ -45,7 +45,7 @@ func (m *mockEncoder) EncodeEVMTx(tx *types.Transaction) ([]byte, error) {
 	return make([]byte, 100+(tx.Nonce()*10)), nil
 }
 
-func (m *mockEncoder) EncodeCosmosTx(tx sdk.Tx) ([]byte, error) {
+func (m *mockEncoder) CosmosTx(tx sdk.Tx) ([]byte, error) {
 	// Create a deterministic byte representation for testing
 	// Use the tx id to ensure uniqueness
 	mockTx, ok := tx.(*mockCosmosTx)
@@ -352,13 +352,13 @@ func TestReapList_EncodingFailure(t *testing.T) {
 // nonceEncoder embeds nonce in bytes for order verification testing
 type nonceEncoder struct{}
 
-func (e *nonceEncoder) EncodeEVMTx(tx *types.Transaction) ([]byte, error) {
+func (e *nonceEncoder) EVMTx(tx *types.Transaction) ([]byte, error) {
 	buf := make([]byte, 8)
 	binary.LittleEndian.PutUint64(buf, tx.Nonce())
 	return buf, nil
 }
 
-func (e *nonceEncoder) EncodeCosmosTx(tx sdk.Tx) ([]byte, error) {
+func (e *nonceEncoder) CosmosTx(tx sdk.Tx) ([]byte, error) {
 	return []byte("cosmos-tx"), nil
 }
 
@@ -520,11 +520,11 @@ func TestReapList_FirstTransactionExceedsLimit(t *testing.T) {
 // alwaysFailEncoder always returns an error for encoding
 type alwaysFailEncoder struct{}
 
-func (e *alwaysFailEncoder) EncodeEVMTx(tx *types.Transaction) ([]byte, error) {
+func (e *alwaysFailEncoder) EVMTx(tx *types.Transaction) ([]byte, error) {
 	return nil, errors.New("encoding always fails")
 }
 
-func (e *alwaysFailEncoder) EncodeCosmosTx(tx sdk.Tx) ([]byte, error) {
+func (e *alwaysFailEncoder) CosmosTx(tx sdk.Tx) ([]byte, error) {
 	return nil, errors.New("encoding always fails")
 }
 
@@ -630,14 +630,14 @@ type selectiveCosmosFailEncoder struct {
 	failID int
 }
 
-func (e *selectiveCosmosFailEncoder) EncodeEVMTx(tx *types.Transaction) ([]byte, error) {
+func (e *selectiveCosmosFailEncoder) EVMTx(tx *types.Transaction) ([]byte, error) {
 	hash := tx.Hash().Bytes()
 	result := make([]byte, 100)
 	copy(result, hash)
 	return result, nil
 }
 
-func (e *selectiveCosmosFailEncoder) EncodeCosmosTx(tx sdk.Tx) ([]byte, error) {
+func (e *selectiveCosmosFailEncoder) CosmosTx(tx sdk.Tx) ([]byte, error) {
 	mockTx, ok := tx.(*mockCosmosTx)
 	if ok && mockTx.id == e.failID {
 		return nil, errors.New("encoding failed for specific tx id")


evmd/mempool.go
```diff
@@ -70,3 +75,42 @@ func (app *EVMD) createMempoolConfig(appOpts servertypes.AppOptions, logger log.
...
+
+func (app *EVMD) NewInsertTxHandler(evmMempool *evmmempool.ExperimentalEVMMempool) sdk.InsertTxHandler {
+	return func(req *abci.RequestInsertTx) (*abci.ResponseInsertTx, error) {
+		txBytes := req.GetTx()
+
+		tx, err := app.TxDecode(txBytes)
+		if err != nil {
+			// TODO: is this the right response here for completely invalid txs
+			// that we cannot process?
+			return &abci.ResponseInsertTx{Code: CodeTypeNoRetry}, fmt.Errorf("decoding tx: %w", err)
```
mattac21: @swift1337 question for you here

tests/integration/mempool/test_mempool_integration.go
```diff
@@ -1014,7 +1014,7 @@ func (s *IntegrationTestSuite) TestEVMTransactionComprehensive() {
 			wantError: false,
 			verifyFunc: func() {
 				mpool := s.network.App.GetMempool()
-				s.Require().Equal(1, mpool.CountTx())
+				s.Require().Equal(0, mpool.CountTx())
```
mattac21, line 1017: I think this broke with the previous anteHandler pr but didnt notice. This tx is now below the base fee so running the anteHandlers drops it from the pool when we try and promote from queued to pending, this is expected.

evmd/mempool.go
swift1337: The logic here should actually be reversed. Comet expects 3 cases:
- `code == 0`. OK
- `0 < code < 32000`. Non-retriable failures (eg tx signature mismatch)
- `code >= 32000`. Retriable failure (eg the mempool is full)

https://github.com/cometbft/cometbft/blob/73384e926a1bd2752985d0e2cc022a27773f6dca/mempool/app_mempool.go#L308

evmd/mempool.go
```diff
@@ -70,3 +75,42 @@ func (app *EVMD) createMempoolConfig(appOpts servertypes.AppOptions, logger log.
 		MinTip:           server.GetMinTip(appOpts, logger),
 	}, nil
 }
+
+const (
+	CodeTypeNoRetry = abci.CodeTypeRetry + 1
+)
+
+func (app *EVMD) NewInsertTxHandler(evmMempool *evmmempool.ExperimentalEVMMempool) sdk.InsertTxHandler {
```
swift1337, line 91: Should we fully delegate InsertTx to EVM mempool and simply move this to  `ExperimentalEVMMempool{}.InsertTx()` ?

Same for `ExperimentalEVMMempool{}.ReapTxs()`

We could have `./mempool/abci.go`:

```go
package mempool

// abci.go

func (m *ExperimentalEVMMempool) InsertTx(...)  {}

func (m *ExperimentalEVMMempool) ReapTxs(...)  {}
```

evmd/mempool.go
swift1337:
```suggestion
			return nil, fmt.Errorf("unable to decode tx: %w", err)
```

mempool/reap_list.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+)
+
+type ReapList struct {
+	txs      ethtypes.Transactions
+	txLookup map[common.Hash]int
+	txsLock  sync.RWMutex
+	encodeTx func(tx *ethtypes.Transaction) ([]byte, error)
```
swift1337:
```suggestion
	cometEncoder func(tx *ethtypes.Transaction) ([]byte, error)
```

mempool/mempool.go
```diff
@@ -282,6 +305,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
 	return nil
 }
 
+var (
+	ErrInvalidTx    = errors.New("tx is invalid")
```
swift1337: Move to `errors.go` as this package uses it

mempool/mempool.go
```diff
@@ -282,6 +305,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
...
+	ErrAlreadyKnown = errors.New("tx already known")
+)
+
+// InsertEVMTxAynsc inserts a tx to the EVM mempool asynchronously. No
+// validation will be run on the tx at insert time besides checking to see if
+// the tx is a valid EVM tx. If it is not an EVM tx, ErrTxIsNotEVM will be returned.
+func (m *ExperimentalEVMMempool) InsertEVMTxAsync(tx sdk.Tx) error {
+	ethMsg, err := m.getEVMMessage(tx)
+	if err != nil {
+		return fmt.Errorf("%w, tx is not an evm tx: %w", ErrInvalidTx, err)
```
swift1337: no need to wrap err, getEVMMessage already has `ErrNotEVMTransaction`

mempool/reap_list.go
```diff
@@ -0,0 +1,118 @@
...
+		if tx == nil {
+			// txs may have "holes" (nil) due to txs being invalidated and
+			// dropped while they are waiting in the reap list
+			nextStart = idx + 1
+			continue
+		}
+
+		txBytes, err := rl.encodeTx(tx)
+		if err != nil {
+			nextStart = idx + 1
```
swift1337: I think here we need to drop the tx from ReapList

mempool/reap_list.go
```diff
@@ -0,0 +1,118 @@
...
+			continue
+		}
+
+		txSize := uint64(len(txBytes))
+		txGas := tx.Gas()
```
swift1337: Q: hm, this is eth gas, does cosmos-evm treat it equally to comet-bft's gas?

mempool/mempool.go
```diff
@@ -282,6 +305,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
...
+// markTxToBeReaped adds a transaction to the reap list when it's promoted to pending status.
+// This is called by the legacy pool when transactions become executable.
+func (m *ExperimentalEVMMempool) markTxToBeReaped(tx *ethtypes.Transaction) {
+	if _, loaded := m.reapGuard.LoadOrStore(tx.Hash(), nil); loaded {
+		return
+	}
+	m.reapList.Push(tx)
+}
+
+func (m *ExperimentalEVMMempool) encodeEVMTx(tx *ethtypes.Transaction) ([]byte, error) {
```
```suggestion
func (m *ExperimentalEVMMempool) encodeToCosmos(tx *ethtypes.Transaction) ([]byte, error) {
```

mempool/mempool.go
```diff
@@ -282,6 +305,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
...
+	// they are fully removed from the mempool and the mempool will no longer
+	// try and add them to the reapList.
+
+	return txs, nil
+}
+
+// markTxToBeReaped adds a transaction to the reap list when it's promoted to pending status.
+// This is called by the legacy pool when transactions become executable.
+func (m *ExperimentalEVMMempool) markTxToBeReaped(tx *ethtypes.Transaction) {
+	if _, loaded := m.reapGuard.LoadOrStore(tx.Hash(), nil); loaded {
```
swift1337: Can we move reapGuard to ReapList? It already has a mutex and a map of tx_hash (we can move to RWMutex)

mempool/mempool.go
```diff
@@ -294,6 +317,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
...
+	ethMsg, err := m.getEVMMessage(tx)
+	if err != nil {
+		return fmt.Errorf("%w, tx is not an evm tx: %w", ErrInvalidTx, err)
+	}
+
+	hash := ethMsg.Hash()
+	m.logger.Debug("inserting EVM transaction", "tx_hash", hash)
+
+	errs := m.txPool.Add([]*ethtypes.Transaction{ethMsg.AsTransaction()}, false)
+	if len(errs) > 1 {
```
swift1337: consider this:

```go
// create helper func   
err := compactErrors(errs ...error) error
   
switch {
 case err == nil:
   return nil
 case errors.Is(A):
    // ...
 case errors.Is(B):
    // ...
default:
  return ErrInvalidTx
}
```

mempool/mempool.go
```diff
@@ -294,6 +317,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
...
+	if len(errs) > 0 && errs[0] != nil {
+		// got an error adding, multiple packages error types can be returned
+		// here, convert to a simplified error set
+		err := errs[0]
+		m.logger.Error("failed to insert EVM transaction", "error", err, "tx_hash", hash)
+
+		switch {
+		case errors.Is(err, txpool.ErrAlreadyKnown):
+			return ErrAlreadyKnown
+		case errors.Is(err, legacypool.ErrTxPoolOverflow) || errors.Is(err, txpool.ErrUnderpriced) || errors.Is(err, legacypool.ErrFutureReplacePending):
```
swift1337: Why not return these errors directly? My understanding is that we need to use regular SDK errors with proper namespaces and codes, so comet can log errors, but retry them later if needed

mempool/txpool/legacypool/legacypool.go
```diff
@@ -805,11 +812,16 @@ func (pool *LegacyPool) add(tx *types.Transaction) (replaced bool, err error) {
...
+			pool.TxRemoved(old)
 			pendingReplaceMeter.Mark(1)
 		}
 		pool.all.Add(tx)
 		pool.priced.Put(tx)
 		pool.queueTxEvent(tx)
+
+		// tx went straight to the pending queue and bypassed the queue, mark
+		// it as promoted
+		pool.TxPromoted(tx)
```
swift1337: should be unexported

mempool/txpool/legacypool/legacypool.go
```diff
@@ -906,13 +918,15 @@ func (pool *LegacyPool) promoteTx(addr common.Address, hash common.Hash, tx *typ
 		// An older transaction was better, discard this
 		pool.all.Remove(hash)
 		pool.priced.Removed(1)
+		pool.TxRemoved(tx)
```
swift1337: should be unexported

mempool/txpool/legacypool/legacypool.go
```diff
@@ -2014,3 +2049,15 @@ func (pool *LegacyPool) SubscribeToNextReorg() (chan struct{}, error) {
 func (pool *LegacyPool) HasPendingAuth(addr common.Address) bool {
 	return pool.all.hasAuth(addr)
 }
+
+func (pool *LegacyPool) TxPromoted(tx *types.Transaction) {
```
swift1337: Also, maybe `markPromoted` is more suitable? same for `markRemoved`

mempool/mempool.go
```diff
@@ -294,6 +317,120 @@ func (m *ExperimentalEVMMempool) InsertInvalidNonce(txBytes []byte) error {
...
+var (
+	ErrInvalidTx    = errors.New("tx is invalid")
+	ErrMempoolFull  = errors.New("mempool is full")
+	ErrAlreadyKnown = errors.New("tx already known")
+)
+
+// InsertEVMTxAynsc inserts a tx to the EVM mempool asynchronously. No
+// validation will be run on the tx at insert time besides checking to see if
+// the tx is a valid EVM tx. If it is not an EVM tx, ErrTxIsNotEVM will be returned.
+func (m *ExperimentalEVMMempool) InsertEVMTxAsync(tx sdk.Tx) error {
```
swift1337: Hm, we need to think about this method more. (I'd also propose `SubmitTransaction` naming)

We need to support non-EVM transactions here as well (w/ fallback to cosmos mempool) because with app-side mempol, all rpc/gossiped transactions are routed via `abci.InsertTx` instead of `abci.CheckTx`

evmd/mempool.go
mattac21: I personally think it makes sense to have them in evmd/mempool.go. For the `InserTx` handler we need a reference to `evmd` in order to decode the tx bytes to a `sdk.Tx`, so I dont think it makes sense to put that logic into the mempool. For `ReapTxs` there isn't really any logic besides taking the values out of the `abci.RequestReapTxs` struct, which imo also doesnt belong in the mempool.

mempool/reap_list.go
mattac21: imo `encodeTx` (or something non comet specific) makes sense here since the reap list does not have to do with comet in theory. If the caller of `NewReapList` wants to pass in an encoder that is called `cometEncoder` to them, then that makes sense, but the `reapList` `encoder` field should not be called `comet` since it could be for anything.

mempool/reap_list.go
mattac21: We are dropping this tx by incrementing `nextStart` to `idx + 1`. At the end of this function we are going to reslice the list to `rl.txs[nextStart:]` (or to an empty list), so by setting `nextStart` to be higher than the index of this bad tx, we know it will be removed.

mempool/reap_list.go
mattac21: mm great question, im not sure, let me dig into it.

mattac21: I think they are the same. This is the gas limit of the evm tx, and in [this conversion function](https://github.com/cosmos/evm/blob/ma/mempool-insert/x/vm/types/msg.go#L344) to build an sdk txfrom an evm tx, we set the `GasLimit` of the `sdk.Tx` to be the same as the `tx.Gas` value.

mempool/reap_list.go
mattac21: this isn't super clear tho so ill add a comment

mempool/mempool.go
mattac21: yeah we can, I had originally implemented dropping in two different ways and the guard had to be on the mempool, but now it can just be combined with the txlookup and that can serve both use cases.

mempool/mempool.go
mattac21: I'm not sure what you mean here? why does this need to use an sdk type error? This is just being returned to the `InsertTxHandler` in `evmd/mempool` which translates it to one of the abci codes you defined.

mattac21: my thought on the error handling here was I realized there was like 5 different very specific error types that can be returned from here, from multiple different packages. I didnt want callers (the `InsertTxHandler` in `evmd/mempool.go`) to have to know about all of those details and how to handle those specific errors, so I combined them into a more simple set of errors.

mempool/txpool/legacypool/legacypool.go
mattac21: agree yeah I didnt love the name

mempool/mempool.go
mattac21: yeah good callout, that was an oversight, ill update to handle cosmos txs too.

mattac21: mm this is actually a very large refactor, since we dont have the idea of the queued pool for cosmos txs, if they make it into the cosmos mempool without having checktx run on them, they will be committed to a block without verification, so we need a separate goroutine to checktx them.

Eric-Warehime: Yeah, agree we will definitely need to support cosmos txs. Let's move it to a followup task though.

mempool/mempool.go
Eric-Warehime: maybe [errors.Join](https://pkg.go.dev/errors#Join)?

mempool/mempool.go
Eric-Warehime: Would we not also call this method from the EVM RPC? Maybe the specific error would matter more there since it wouldn't need to be translated to an ABCI response?

swift1337: >Would we not also call this method from the EVM RPC? 

Yeah, evm rpc would call `abci.InsertTx`

mempool/mempool.go
mattac21: oh well... its been done lol

mattac21: but not in a very smart/efficient way like I was describing above, basically just runs check tx before cosmos txs being inserted as well, but cosmos txs should be able to be submitted now. in a follow up task we can make cosmos txs have the same gains that we get for evm txs with the queue.

mempool/mempool.go
mattac21: we want to be sure we are only getting one error since we are only submitting one tx so join doesnt really work here. I am just going to remove this and move the logic to handle the error into the insert tx handler

mempool/mempool.go
```diff
@@ -226,40 +252,120 @@ func (m *ExperimentalEVMMempool) GetTxPool() *txpool.TxPool {
...
+	// fails
+	msCache := ms.CacheMultiStore()
+
+	// set the branched multistore as the multistore that the context will use.
+	// so writes happening via this context will use the branched multistore.
+	ctx = ctx.WithMultiStore(msCache)
+
+	// execute the anteHandlers on our new context, and get a context that has
+	// the anteHandler updates written to it.
+	if _, err := m.anteHandler(ctx, tx, false); err != nil {
```
mattac21, line 352: system tests are already broken but this is going to break system tests as well if there is a check tx running in the rpc call path, then we use the same check tx state here, we are going to get nonce errors since the same tx is being checked twice.

mempool/mempool.go
```diff
@@ -358,6 +474,7 @@ func (m *ExperimentalEVMMempool) Remove(tx sdk.Tx) error {
 	if err != nil {
 		m.logger.Error("failed to remove Cosmos transaction", "error", err)
 	} else {
+		m.reapList.DropCosmosTx(tx)
```
mattac21, line 480: There is a sight issue here with the priority nonce mempool and reap list interaction that I am choosing to ignore and made a followup ticket to address. But the priority nonce mempool removes txs from it when it gets one with a higher priority. There is no way to hook into that process and drop from the reap list when those txs are removed. I am just ignoring this for now since this shouldn't affect our testing, but is something to note.

mempool/mempool.go
```diff
@@ -226,40 +252,120 @@ func (m *ExperimentalEVMMempool) GetTxPool() *txpool.TxPool {
...
+	// is completely separate from the context used to verify EVM txs. If we
+	// share a context here, then we would to block this on the verifications
+	// running inside of the legacypool, which may take awhile... but sharing a
+	// context between the two seems more correct. Note that the legacypool
+	// also uses two separate contexts right now (one for queued verification
+	// and one for pending. which one would we share with? should we combine
+	// the two? we cant directly a context between the two, since we may verify
+	// the same tx twice at the same height which would error for a valid tx,
+	// maybe we need some cache to ensure this doesnt happen, then they could
+	// be shared and we skip double verifications.)
```
mattac21, line 329: this is a larger note I am thinking about, how we manage the state in all of the new checks that we are introducing. Does it ever make sense to share the state, should they all be separate, etc. For this check I am choosing to use the actual check state since these txs are essentially being 'streamed' in from an rpc, so are able to maintain the state of prev txs only if we use and write to the check state. For the legacypool ante handlers we are starting from a fresh block state every run and running all txs in the individual pools against that fresh state.

I think there are some optimizations we can make where they can share a single state and we then dont have to run each tx multiple times potentially, but then we are going to introduce some locking between all of the different checks which may further block proposals. Just a thought.

mempool/reap_list.go
```diff
@@ -0,0 +1,203 @@
...
+
+type txWithHash struct {
+	bytes []byte
+	hash  string
+	gas   uint64
+}
+
+type ReapList struct {
+	// txs is a list of transactions and their respective hashes
+	// NOTE: this currently has unbound size
```
mattac21, line 28: this can be handled in a future pr if we want to change, but we are just relying on the fact that we will be continuously deleting from this list on tx removals/drops, and will not insert into it if the mempool becomes full and we stop accepting txs. so this is implicitly tied to the size of the mempool(s).

mempool/reap_list.go
```diff
@@ -0,0 +1,203 @@
...
+	txs []*txWithHash
```
mattac21, line 29: reap list now handles evm and cosmos txs

mempool/encoding.go
```diff
@@ -0,0 +1,46 @@
...
+type TxEncoder struct {
+	txConfig client.TxConfig
+}
+
+func NewTxEncoder(txConfig client.TxConfig) *TxEncoder {
+	return &TxEncoder{txConfig: txConfig}
+}
+
+// EncodeEVMTx encodes an evm tx to its sdk representation as bytes.
+func (e *TxEncoder) EncodeEVMTx(tx *ethtypes.Transaction) ([]byte, error) {
```
swift1337:
```suggestion
func (e *TxEncoder) EVMTx(tx *ethtypes.Transaction) ([]byte, error) {
```

mempool/encoding.go
```diff
@@ -0,0 +1,46 @@
...
+	txBytes, err := e.txConfig.TxEncoder()(txBuilder.GetTx())
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode transaction: %w", err)
+	}
+
+	return txBytes, nil
+}
+
+// EncodeCosmosTx encodes a cosmos tx to bytes.
+func (e *TxEncoder) EncodeCosmosTx(tx sdk.Tx) ([]byte, error) {
```
swift1337:
```suggestion
func (e *TxEncoder) CosmosTx(tx sdk.Tx) ([]byte, error) {
```

mempool/encoding.go
```diff
@@ -0,0 +1,46 @@
...
+	msg.FromEthereumTx(tx)
+
+	// Build cosmos tx
+	txBuilder := e.txConfig.NewTxBuilder()
+	if err := txBuilder.SetMsgs(msg); err != nil {
+		return nil, fmt.Errorf("failed to set msg in tx builder: %w", err)
+	}
+
+	// Encode to bytes
+	txBytes, err := e.txConfig.TxEncoder()(txBuilder.GetTx())
```
swift1337:
```suggestion
	return e.CosmosTx(txBuilder.GetTx())
```

mempool/mempool.go
```diff
@@ -200,8 +204,30 @@ func NewExperimentalEVMMempool(
 		anteHandler:        config.AnteHandler,
 		operateExclusively: config.OperateExclusively,
 	}
+	evmMempool.reapList = NewReapList(txEncoder)
 
 	legacyPool.RecheckTxFnFactory = recheckTxFactory(txConfig, config.AnteHandler)
+
+	// Once we have validated that the tx is valid (and can be promoted, set it
+	// to be reaped)
+	legacyPool.OnTxPromoted = func(tx *ethtypes.Transaction) {
```
swift1337, line 216: not a blocker for testing, but I'd like to avoid public props that can be set outside of struct's constructor. It makes mental model more complex because the reader must keep multiple init places in mind

mempool/reap_list.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// PushCosmosTx enqueues a cosmos tx into the reap list.
+func (rl *ReapList) PushCosmosTx(tx sdk.Tx) error {
+	txBytes, err := rl.txEncoder.EncodeCosmosTx(tx)
+	if err != nil {
+		return fmt.Errorf("encoding evm tx to bytes: %w", err)
+	}
+
+	var gas uint64
```
swift1337, line 149: we should return err if gas==0

mempool/reap_list.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		rl.txIndex[tx.hash] = i
+	}
+
+	return result
+}
+
+// PushEVMTx enqueues an EVM tx into the reap list.
+func (rl *ReapList) PushEVMTx(tx *ethtypes.Transaction) error {
+	txBytes, err := rl.txEncoder.EncodeEVMTx(tx)
```
swift1337: nit: this will trigger multiple encodings for duplicate insertions. should be fine for PoC

mempool/mempool.go
swift1337: Yeah, let's put a ticket in Linear. After benchmarking, we can compare a diff agaings `main` and write down what should be productized, improved, refactored, etc...

mempool/reap_list.go
mattac21: good catch ill change

mempool/mempool.go
mattac21: yea agree, this was just how the broadcast function was done which was already here so I was copying the pattern. but fully agree this should change if we want to productionize

PR #873, opened 2025-12-03
feat(rpc): use abci.InsertTx for SendRawTransaction

swift1337:
This PR alters EVM RPC to use `abci.InsertTx` instead of broadcasting tx from app to cometbft. It also contains minor rpc/evmBackend refactoring

Current implementation mocks **InsertTx** with **CheckTx** (will be swapped with mempool call in the following PRs)

Closes: STACK-1846

---

## Author Checklist

I have...

### DIFF
--- a/rpc/backend/call_tx.go
+++ b/rpc/backend/call_tx.go
@@ -24,6 +24,7 @@ import (
 
 	errorsmod "cosmossdk.io/errors"
 
+	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/flags"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
@@ -153,10 +154,10 @@ func (b *Backend) SendRawTransaction(data hexutil.Bytes) (common.Hash, error) {
 	// publish tx directly to app-side mempool, avoiding broadcasting to consensus layer.
 	if b.UseAppMempool {
 		if err := b.mempoolInsertTx(txHash, txBytes); err != nil {
-			return common.Hash{}, err
+			return b.handleSendTxError(tx, ethSigner, err)
 		}
 
-		return tx.Hash(), nil
+		return txHash, nil
 	}
 
 	syncCtx := b.ClientCtx.WithBroadcastMode(flags.BroadcastSync)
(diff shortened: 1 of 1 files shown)


rpc/apis.go
```diff
@@ -150,43 +152,44 @@ func init() {
...
+func BuildRPCs(
+	selectedAPIs []string,
+	ctx *server.Context,
 	clientCtx client.Context,
 	stream *stream.RPCStream,
-	allowUnprotectedTxs bool,
-	indexer servertypes.EVMTxIndexer,
-	selectedAPIs []string,
-	mempool *evmmempool.ExperimentalEVMMempool,
+	backendProvider func() backend.BackendI,
```
mattac21: I like the refactor to accept a `backend.BackendI` instead of building it in the constructor, but why use a `backendProvider` function here instead of directly passing the `backend.BackendI` as an arg? the only impl I see is returning `evmBackend` in `StartJSONRPCServer`.

rpc/backend/call_tx.go
```diff
@@ -146,9 +147,19 @@ func (b *Backend) SendRawTransaction(data hexutil.Bytes) (common.Hash, error) {
 		return common.Hash{}, fmt.Errorf("failed to encode transaction: %w", err)
 	}
 
-	txHash := ethereumTx.AsTransaction().Hash()
+	txHash := tx.Hash()
+
+	// publish tx directly to app-side mempool, avoiding broadcasting to consensus layer.
+	if b.UseAppMempool {
+		if err := b.mempoolInsertTx(txHash, txBytes); err != nil {
```
mattac21, line 156: not sure what your plan was exactly to merge this, but do you want to wait for the actual inserttx handler to get merged and then modify this to use the real impl? or what was your thought?

rpc/apis.go
swift1337: Yeah, I thought it would be convenient if different RPCs used different backends, but no — all of them use the same. Good call

rpc/backend/call_tx.go
swift1337: I think it's safe to merge because it will simply call `CheckTx`

https://github.com/cosmos/evm/pull/873/files#diff-8a13a79f302fcfae6a776342d9ba125cb8beccc6c8a8a92bd1adcb1c84561d68R57

PR #880, opened 2025-12-08
feat(mempool): add timeout for reorg

swift1337:
# Description

- [x] Unify `mempool.Select` and `mempool.SelectBy` calls
- [x]  Share re-check function across the whole reorg loop
- [x] Implement "max timeout" for total recheck tx for the reog loop in evm mempool
- [x] Add basic test case 

## Timeout logic

LegacyPool.WaitForReorgHeight now also has a timeout parameter (hard-coded to `300ms` in `mempool.Select`)

```go
func (pool *LegacyPool) WaitForReorgHeight(ctx context.Context, height int64, timeout time.Duration)
```

Inside this function, we call `reorgContext` that either sets `ctx.timeout = timeout` or lowers existing timeout by **100ms** ~ `ctx.timeout -= 100ms`. 

After ctx is timed out, is calls an atomic bool `reorgCancelRequested` that is checked for every account during `promoteExecutables` or `demoteUnexecutables`

**Note**: due to how EVM mempool works, it's challenging to properly propagate `ctx` across all goroutines as the mempool follows the actor model with lots of channels. This leads to a semi-forced context timeout, where we might still have **some** operations running after the timeout (ie, hard timeout if not guaranteed)

Closes: STACK-1927

---

## Author Checklist

I have...

### DIFF
--- a/mempool/txpool/legacypool/legacypool_test.go
+++ b/mempool/txpool/legacypool/legacypool_test.go
@@ -2888,7 +2888,7 @@ func TestPromoteExecutablesRecheckTx(t *testing.T) {
 func TestPromoteExecutablesCancelReorg(t *testing.T) {
 	// ARRANGE
 	// Given a pool with txs
-	pool, _, _ := setupPool()
+	pool, rechecker, _ := setupPool()
 	defer pool.Close()
 
 	const txsCount = 10
@@ -2915,6 +2915,17 @@ func TestPromoteExecutablesCancelReorg(t *testing.T) {
 		recheckDuration = 120 * time.Millisecond
 		maxReorgWait    = 2 * recheckDuration
 	)
+
+	recheck := func(ctx sdk.Context, tx *types.Transaction) (sdk.Context, error) {
+		txHash := tx.Hash().Hex()
+
+		t.Logf("start: recheck: tx %s", txHash)
+		time.Sleep(recheckDuration)
+		t.Logf("end: recheck: tx %s", txHash)
+		return ctx, nil
+	}
+
+	rechecker.SetRecheck(recheck)
 
 	// ACT
 	go func() {


mempool/txpool/legacypool/legacypool.go
```diff
@@ -1319,6 +1324,10 @@ func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest,
 		// the flatten operation can be avoided.
 		promoteAddrs = dirtyAccounts.flatten()
 	}
+
+	// function to perform tx rechecking on cosmos-sdk side
+	txReChecker := pool.buildRecheckFn()
```
mattac21: with the way the current recheck function is written, we can't use a single instance of it for both `promoteExecutables` and `demoteUnexecutables`. Say we run `anteHandlers` on a queued tx in `promoteExecutables` and `reset != nil` so `demoteUnexecutables` is going to run after promote. If this tx passes recheck, it will have updated its accounts nonce in the context. This _same_ context is then going to be used by `demoteUnexecutables` to recheck the same transaction. It will then fail with an `ErrNonceLow` (tx nonce lower than accounts nonce) and be removed from the mempool. I think this is why the system tests are failing where they are, when ever I accidentally do this, that is typically where it fails.

I dont think this is necessarily something you 100% need to fix though since I am reworking how a lot of the recheck context works anyways in my pr, so I dont think its worth it for you to spend time here when I'm about to change it, so this is just a callout. (could also be why solidity tests are failing, idk how exactly those work tho)

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1680,6 +1696,11 @@ func (pool *LegacyPool) demoteUnexecutables() {
 		}
 		pendingNofundsMeter.Mark(int64(len(drops)))
 
+		// if reorg cancellation requested, return early
+		if pool.reorgCancelRequested.Load() {
+			return
+		}
```
mattac21: if we timeout during `demoteUnexecutables`, then we may have txs in the pending pool that have not been rechecked. We need to be sure that we dont include these in a proposal (so not returned from `SelectBy`) until they are properly rechecked. Possibly I am missing this but it doesnt look like there is anything preventing this? It looks like we would return here on a timeout, and then `SelectBy` would still potentially return all pending txs.

mempool/txpool/legacypool/legacypool.go
swift1337: done

mempool/txpool/legacypool/list.go
```diff
@@ -276,6 +276,8 @@ type list struct {
 	costcap   *uint256.Int // Price of the highest costing transaction (reset only if exceeds balance)
 	gascap    uint64       // Gas limit of the highest spending transaction (reset only if exceeds block limit)
 	totalcost *uint256.Int // Total cost of all transactions in the list
+
+	lastCheckedHeight int64 // the height at which the list was last checked
```
swift1337, line 317: No mutex required, as this is an internal structure and its access is already under a lock

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1737,6 +1766,15 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			if _, ok := pool.queue[addr]; !ok {
 				pool.reserver.Release(addr)
 			}
+		} else {
+			// set the last checked height for the list
+			// this is used to return only checked accounts in Pending method
+			list.setLastCheckedHeight(height)
```
swift1337, line 1838: @mattac21 @Eric-Warehime , so this is the place where we'd set the last checked height. 

The drawback of this approach: _now,_ to be included in `pool.Pending() → mempool.SelectBy()` results, tx should go through `demoteUnexecutables`, which is not the case for same-block-inclusion (imagine an addition to an empty mempool) eg:

 `[time_0: block(1), t1: add_tx(), t2: propose_block(2)]`

Do you see any easy ways to handle this case?

PR #881, opened 2025-12-08
feat(mempool): Promote executables runs on context set by demote unexecutables

mattac21:
# Description

Updates `promoteExecutables` to run on a context that has been updated by the `recheckFn` of all of the rechecks done in `demoteUnexecutables` (if available). 

This is necessary if a pending tx's `recheckFn` execution will make updates to the context that may cause a queued tx to be invalid during `promoteExecutables`. If then we do not run rechecks in `promoteExecutables` on top of the state that has had `demoteExecutables` rechecks run on it, may promote a tx into the pending queue that is invalid, and only demote it the next time we see a block, allowing for a time period where we can create an invalid proposal.

To do this I did a bit of a refactor to how context is managed across the reorg loop. Previously we had a single large callback function that manages it all and captures the context to be used as a closure, relying on callers to know when exactly they should call the factory function. Now we have a `rechecker` struct that manages persisting a shared context across reorg loop invocations, relying on callers to get the context when necessary, use it for invocations of `recheck`, and write their updates back to it. This does introduce some sdk code into the legacypool for the `Rechecker` interface.

Closes: STACK-1951

---

## Author Checklist

I have...

### DIFF
--- a/mempool/txpool/legacypool/legacypool.go
+++ b/mempool/txpool/legacypool/legacypool.go
@@ -2088,6 +2088,8 @@ func (pool *LegacyPool) markTxRemoved(tx *types.Transaction) {
 // ignored from recheck, i.e. we do not want to drop txs from the mempool if we
 // have received specific errors from recheck.
 func tolerateRecheckErr(err error) error {
+	// TODO: Fix import cycle if we try and properly match on
+	// errors.Is(mempool.ErrNonceLow)
 	if err != nil && strings.Contains(err.Error(), "tx nonce is higher than account nonce") {
 		return nil
 	}


mempool/txpool/legacypool/rechecker.go
```diff
@@ -0,0 +1,104 @@
...
+// time you recheck a pending transaction via Pending. Note this is not thread
+// safe with itself or Pending.
+func (r *rechecker) NewQueuedChecker() func(tx *ethtypes.Transaction) error {
+	// Create a cache ctx so that all queued txs that are checked using the
+	// returned function act on the same context. These invocations of the
+	// recheckFn will only persist context updates back to the main queueCtx if
+	// they are invoked successfully (since they will now be promoted to
+	// pending, we should run all future queued rechecks on a context that has
+	// been updated by this tx).
+	ctx, _ := r.queueCtx.CacheContext()
```
Eric-Warehime: We need to branch the gas meter to avoid race conditions. See https://github.com/cosmos/evm/pull/882

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1437,6 +1467,12 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 	// Track the promoted transactions to broadcast them at once
 	var promoted []*types.Transaction
 
+	// create a checker instance to ensure all queued txs are valid
+	var checker func(tx *types.Transaction) error
```
Eric-Warehime: This variable doesn't seem necessary...we can just construct this where it's used can't we?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1437,6 +1467,12 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
+	if pool.rechecker != nil {
```
Eric-Warehime: I think we should just make these interfaces and put no-op implementations in place. The nil checks are annoying and confusing to the readability imo.

mempool/txpool/legacypool/rechecker.go
```diff
@@ -0,0 +1,104 @@
...
+		r.queueCtx = newCtx
+	}
+	return err
+}
+
+// NewQueuedChecker returns a function that can be used to recheck queued
+// transactions. Note that you should get a new QueuedChecker function every
+// time you recheck a pending transaction via Pending. Note this is not thread
+// safe with itself or Pending.
+func (r *rechecker) NewQueuedChecker() func(tx *ethtypes.Transaction) error {
```
Eric-Warehime: This is very similar to the constructor pattern where we need re-create a new function in order to have it wrapped by the correct context. 

I think I actually prefer something like the sdk's state manager where we can get/set/reset under locks.

mempool/txpool/legacypool/legacypool.go
mattac21: technically we could but then we would be running every queued txs recheck without the previous queued txs context updates. it is placed outside of the loop so that the context is captured here, then each queued tx recheck writes to the same context.

mempool/txpool/legacypool/rechecker.go
mattac21: down to change this, I felt like this much smaller factory fn was a bit better than the single mega one. I was trying to avoid the promote / demote functions having to know about how to manage state/context for their recheck calls. But I guess having to know to create a QueuedChecker outside of the loop is essentially the same.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -300,6 +312,7 @@ func New(config Config, chain BlockChain) *LegacyPool {
 		queue:               make(map[common.Address]*list),
 		beats:               make(map[common.Address]time.Time),
 		all:                 newLookup(),
+		rechecker:           newNopRechecker(),
```
mattac21, line 324: eliminates a nil check on the rechecker, if the rechecker is not provided via options, we use the noop impl

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1465,26 +1504,26 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
-			recheckStart := time.Now()
-			recheckFn := pool.RecheckTxFnFactory(pool.chain)
-			recheckDrops, _ = list.Filter(func(tx *types.Transaction) bool {
-				return recheckFn(tx) != nil
-			})
-			for _, tx := range recheckDrops {
-				pool.all.Remove(tx.Hash())
-				pool.markTxRemoved(tx)
+		recheckStart := time.Now()
+		recheckDrops, _ := list.FilterSorted(func(tx *types.Transaction) bool {
```
mattac21, line 1489: implemented a new `FilterSorted` that will ensure we do rechecks in ascending nonce order for accounts. I did this because I started seeing errors during queued and pending rechecks where we had 2 txs in the pool and it would check the higher nonce first, previously we just iterated over the raw list of txs, which didnt have a guaranteed ordering.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1465,26 +1504,26 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
+			newCtx, err := pool.rechecker.Recheck(ctx, tx)
+			if !newCtx.IsZero() {
+				ctx = newCtx
+			}
+			if err == nil && reset == nil {
```
mattac21, line 1494: during promote when resetting (new block, demote will also run), we can't write back context updates to the context that demote will also use, since we would increment the nonce here, and then recheck again in demote and fail with a nonce too low error. however we do still want the ctx that we use for the next pending recheck to have these changes, so we are still updating `ctx` to `newCtx` even if we are resetting.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -2063,3 +2102,13 @@ func (pool *LegacyPool) markTxRemoved(tx *types.Transaction) {
 		pool.OnTxRemoved(tx)
 	}
 }
+
+// tolerateRecheckErr returns nil if err is an error string that should be
+// ignored from recheck, i.e. we do not want to drop txs from the mempool if we
+// have received specific errors from recheck.
+func tolerateRecheckErr(err error) error {
+	if err != nil && strings.Contains(err.Error(), "tx nonce is higher than account nonce") {
```
mattac21, line 2093: this is rough, but importing the actual error here is a import cycle. probably a better way to handle this, but I felt this was fine for the poc

mempool/txpool/legacypool/legacypool_test.go
```diff
@@ -3087,3 +3081,24 @@ func BenchmarkMultiAccountBatchInsert(b *testing.B) {
 		pool.addRemotesSync([]*types.Transaction{tx})
 	}
 }
+
+type MockRechecker struct {
```
mattac21, line 3085: This is implemented in legacypool_test and mempool_test which also feels very bad, but putting this in a legacypool/mocks package and importing here is an import cycle, putting it in mempool/mocks and importing here is an import cycle. We should also find a better way to do this (I think putting it in legacypool/mocks and then change this package to `_test` would fix), but again I felt this was fine for poc.

mempool/txpool/legacypool/legacypool_test.go
mattac21, line 1: I spent awhile trying to add tests to this but basically ended up failing to write any meaningful kinds of tests. The tests would either have a ton of mocks where these tests weren't meaningful, or no mocks but then its basically impossible to invoke the issue since I dont think its actually possible to have a pending tx cause a queued tx to fail... I think we can come back if necessary, but it seems like this part of the code is so in flux right now it didnt feel worth it to spend time on more testing here. LMK others feel different and I can spend more time on this

go.mod
swift1337, line 1: Should `evmd/go.mod` be updated as well?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -309,9 +322,22 @@ func New(config Config, chain BlockChain) *LegacyPool {
...
 	pool.latestReorgHeight.Store(0)
 
+	for _, opt := range opts {
+		opt(pool)
+	}
+
 	return pool
 }
 
+type Option func(pool *LegacyPool)
```
swift1337: nit: move Options above the constructor for better readability

mempool/txpool/legacypool/legacypool.go
```diff
@@ -141,6 +143,22 @@ type BlockChain interface {
 	StateAt(root common.Hash) (vm.StateDB, error)
 }
 
+// Rechecker defines the minimal set of methods needed to recheck transactions
+// and manage the context that the transactions are rechecked against.
+type Rechecker interface {
+	// GetContext gets the current context that transactions should be rechecked against,
+	// and a way to update that context with changes.
+	GetContext() (sdk.Context, func())
```
swift1337: Comment was a bit confusing imo

```suggestion
	GetContext() (context sdk.Context, write func())
```

mempool/rechecker.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return sdk.Context{}, fmt.Errorf("converting evm tx %s to cosmos tx: %w", tx.Hash(), err)
+	}
+
+	return r.anteHandler(ctx, cosmosTx, false)
+}
+
+// Update updates the base context for rechecks based on the latest chain
+// state.
+func (r *Rechecker) Update(chain legacypool.BlockChain, header *ethtypes.Header) {
```
swift1337, line 69: Please add a note whether it's safe to call this concurrently, same for other methods

mempool/txpool/legacypool/legacypool.go
swift1337: Yeah, I wanted to flag this as well. I thought we could move this inside the rechecker itself, but then we'd return `err=nil` and call `write()`, which is incorrect 🤔.

OK, let's keep this, but add TODO

mempool/txpool/legacypool/list.go
```diff
@@ -132,6 +148,26 @@ func (m *SortedMap) reheap() {
 	m.cacheMu.Unlock()
 }
 
+// filterSorted is the same as filter, but iteration over the transactions goes
+// from lowest to highest nonce.
+func (m *SortedMap) filterSorted(filter func(*types.Transaction) bool) types.Transactions {
+	var removed types.Transactions
+
+	sorted := m.flatten()
```
swift1337, line 157:
```suggestion
  // flatten sorts txs by nonce asc.
	sorted := m.flatten()
```

mempool/txpool/legacypool/legacypool.go
mattac21: yea exactly, I actually had that at one point as well and was super confused for a bit why things weren't working lol

go.mod
mattac21: this update came from generating mocks, so I think this is fine as is

mempool/rechecker.go
mattac21: I added that to the `Rechecker` struct comment, but I can copy paste to methods as well

Issue #889, opened 2025-12-11
Experimental EVM mempool not gossiping nonce + n transactions


jhelison:
# Issue description

Recently, we have found an issue where the transactions with nonce + n were not being gossiped.
This is the behaviour that we have faced:

**If sent to a Validator node with enough voting power**

Within a validator, the message is kept on the local mempool and all TXs are added to a block:
1. TX A with nonce + 1 is created, lands in queue
2. TX B with nonce + 0 is created, lands in runnable
3. TX B is processed and lands in a block
4. TX A is processed and lands in a block

**Non validator node**

On non-validator nodes, only the runnable TX is gossiped to the validator and added to a block:
1. TX A with nonce + 1 is created, lands in queue
2. TX B with nonce + 0 is created, lands in runnable
3. TX B is processed and lands in a block
4. TX A is now stuck and can't be re-sent unless overpriced under the same nonce

# How to replicate

This issue can be replicated by:
- Have a node with the Experimental EVM Mempool enabled
- A node available with P2P connections to a validator, transactions must be sent to the public node

The following must be applied:
1. Send an EVM TX with nonce + 1
2. Send an EVM TX with nonce + 0 (expected nonce)

The following will happen:
- Only the TX with nonce + 0 will be executed
- TX with nonce + 1 will be stuck on the local mempool and will not be gossiped
- Wallet will be stuck unless an overpriced TX is sent to recover the Nonce + 1 TX

The following would be the expected behaviour:
- Nonce + 0 should be executed
- Nonce + 1 should be executed (after internal mempool reordering)

jhelison:
This is a proposal to fix the issue [#890](https://github.com/cosmos/evm/pull/890)

PR #890, opened 2025-12-11
fix: experimental EVM mempool not gossiping nonce + n transactions

jhelison:
# Description

This is related to issue https://github.com/cosmos/evm/issues/889 and has multiple fixes related to the Experimental mempool not gossiping `nonce+n` transactions.
More details about the problem can be found in the GitHub issue. Below you will find problems faced, solutions, and explanations.

The explanation follows the order in which they were faced during development.

## Empty clientCtx

An empty client context was being used on the function [broadcastEVMTransactions](https://github.com/cosmos/evm/blob/d97c9e5370b224b5d48e03592bb4074a42cee289/mempool/mempool.go#L483):
- The experimentation mempool object was being created at a moment when the `clientCtx` was not available
- A function called [app.SetClientCtx](https://github.com/cosmos/evm/blob/d97c9e5370b224b5d48e03592bb4074a42cee289/evmd/app.go#L1066) exists, but when executed, only the clientCtx on the App was replaced
- This then was making `BroadcastTxSync` fail
  - Logs should be thrown for the [error here](https://github.com/cosmos/evm/blob/fe790fb8b398cf4bc99dfe5e727d39e5ed28d195/mempool/txpool/legacypool/legacypool.go#L1336), but due to a logger mismatch, nodes don't show the error log

**Fix**

Fix for the issue can be found at commit https://github.com/cosmos/evm/commit/5b0b89acf1dafef9070a82b8020e8ecfca138a79:
- It enforces a always up to date fetch of the clientCtx from the app
- This can be fixed in other ways too

## Bad encoding on transactions

The function [broadcastEVMTransactions](https://github.com/cosmos/evm/blob/d97c9e5370b224b5d48e03592bb4074a42cee289/mempool/mempool.go#L483) had a weird encoding from TXs:
- Checking the TXs bytes, there were differences between the moment they were broadcast and then re-broadcast on `broadcastEVMTransactions`

**Fix**

This was fixed on commit https://github.com/cosmos/evm/commit/882a856497804b90359b46074a9a3ab4e46017dc:
- There, we enforce a new encoding style that replicates [SendRawTransaction](https://github.com/cosmos/evm/blob/main/rpc/backend/call_tx.go#L136-L152)
- One detail is the line [508](https://github.com/cosmos/evm/blob/882a856497804b90359b46074a9a3ab4e46017dc/mempool/mempool.go#L508) that makes the loop to continue happening on failed TX with errors related to mempool

## Badlock on the nonce reordering

This one was painful to solve, but a deadlock was found on TX processing from [BroadcastTxSync](https://github.com/cosmos/evm/blob/d97c9e5370b224b5d48e03592bb4074a42cee289/mempool/mempool.go#L498). But to sum it up, the following happens:
1. [SendRawTransaction](https://github.com/cosmos/evm/blob/main/rpc/backend/call_tx.go#L136-L152) calls [CheckTxAsync](https://github.com/cometbft/cometbft/blob/v0.38.18/abci/client/local_client.go#L48-L49), it locks the function under a mutex
2. Add to the mempool is executed in a way that creates a [channel called done](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L983-L987)
3. This channel only closes when [runReorg](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L1252) is completed
4. But [runReorg](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L1252) calls [BroadcastTxFn](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L1335) if a TX is promoted
5. [BroadcastTxFn](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L1335) then calls [CheckTxAsync](https://github.com/cometbft/cometbft/blob/v0.38.18/abci/client/local_client.go#L48-L49) and lands on the same locked mutex
  - [runReorg](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L1252) never completes, the [channel done](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L983-L987) is never closed, and the first [CheckTxAsync](https://github.com/cometbft/cometbft/blob/v0.38.18/abci/client/local_client.go#L48-L49) never removes its locks, since it depends on the second call to [CheckTxAsync](https://github.com/cometbft/cometbft/blob/v0.38.18/abci/client/local_client.go#L48-L49)

**Fix**

A fix can be found on commit https://github.com/cosmos/evm/commit/f7d966972f00c260c346105902fa80c879742048:
- It seems [Add is supposed to be called with sync as false as described on the function header](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L939-L940)
- The fix just turns off the sync on `Insert`. This makes the [channel done](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L983-L987) to not wait, then completing the first and second calls of [CheckTxAsync](https://github.com/cometbft/cometbft/blob/v0.38.18/abci/client/local_client.go#L48-L49)

## Tx already in mempool

With the fixes above, nodes were running, but no gossiping was happening:
- Let's say that the TX `A` with nonce + 1 exists on the mempool, then TX `B` with nonce + 0 is added
  - This then makes the TX `A` to be promoted, then going on [broadcastEVMTransactions](https://github.com/cosmos/evm/blob/d97c9e5370b224b5d48e03592bb4074a42cee289/mempool/mempool.go#L483)
  - At this stage, both TXs are on the mempool, and TX `B` is gossiped
  - But when the app tries to gossip `B` it fails
- The failure is caused by [ErrAlreadyKnown](https://github.com/cosmos/evm/blob/99abc7f3a47af1289ddd90c279e49f885ee06278/mempool/txpool/legacypool/legacypool.go#L684) that then breaks the flow just before being added to [cometbft mempool](https://github.com/cometbft/cometbft/blob/a5726c1be6d49761e6b3933fd01edbaacce9bb56/mempool/clist_mempool.go#L413)  (and gossiped)

**Fix**

The fix can be found at commit https://github.com/cosmos/evm/commit/d4868524b065f9e760677c5751e7728a982478b6:
- At the `NewCheckTxHandler`, we can just check if the error is `ErrAlreadyKnown`. If so, we can move with the remaining flow
  - This doesn't cause the TX to be added to mempool locally again, but does enforce the gossip
  - If the same TX with nonce + n is sent twice, the system still fails with `underpriced`, and the TX can still be overpriced to be replaced (this is the expected behaviour)

## Others

With these changes, I was able again to send non-ordered nonces to the mempool.
- Locally testing with 50 unordered TXs were done, all were sent on the same block without any issues

---

## Author Checklist

I have...

### DIFF
--- a/evmd/mempool.go
+++ b/evmd/mempool.go
@@ -8,7 +8,6 @@ import (
 	"cosmossdk.io/log"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
-	"github.com/cosmos/cosmos-sdk/client"
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
 
@@ -40,7 +39,6 @@ func (app *EVMD) configureEVMMempool(appOpts servertypes.AppOptions, logger log.
 		app.EVMKeeper,
 		app.FeeMarketKeeper,
 		app.txConfig,
-		func() client.Context { return app.clientCtx }, // Enforces a fresh client context
 		mempoolConfig,
 		cosmosPoolMaxTx,
 	)
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -33,9 +33,6 @@ var _ sdkmempool.ExtMempool = &ExperimentalEVMMempool{}
 // This should be used only in tests to ensure deterministic behavior
 var AllowUnsafeSyncInsert = false
 
-// ClientCtxProvider defines a function type that provides a fresh client context from the app
-type ClientCtxProvider func() client.Context
-
 const (
 	// SubscriberName is the name of the event bus subscriber for the EVM mempool
 	SubscriberName = "evm"
@@ -60,6 +57,7 @@ type (
 		/** Utils **/
 		logger        log.Logger
 		txConfig      client.TxConfig
+		clientCtx     client.Context
 		blockchain    *Blockchain
 		blockGasLimit uint64 // Block gas limit from consensus parameters
 		minTip        *uint256.Int
@@ -96,7 +94,6 @@ func NewExperimentalEVMMempool(
 	vmKeeper VMKeeperI,
 	feeMarketKeeper FeeMarketKeeperI,
 	txConfig client.TxConfig,
-	getClientCtx ClientCtxProvider,
 	config *EVMMempoolConfig,
 	cosmosPoolMaxTx int,
 ) *ExperimentalEVMMempool {
@@ -128,19 +125,6 @@ func NewExperimentalEVMMempool(
 	}
 
 	legacyPool := legacypool.New(legacyConfig, blockchain)
-
-	// Set up broadcast function using clientCtx
-	if config.BroadCastTxFn != nil {
-		legacyPool.BroadcastTxFn = config.BroadCastTxFn
-	} else {
-		// Create default broadcast function using clientCtx.
-		// The EVM mempool will broadcast transactions when it promotes them
-		// from queued into pending, noting their readiness to be executed.
-		legacyPool.BroadcastTxFn = func(txs []*ethtypes.Transaction) error {
-			logger.Debug("broadcasting EVM transactions", "tx_count", len(txs))
-			return broadcastEVMTransactions(getClientCtx(), txs)
-		}
-	}
 
 	txPool, err := txpool.New(uint64(0), blockchain, []txpool.SubPool{legacyPool})
 	if err != nil {
@@ -187,6 +171,7 @@ func NewExperimentalEVMMempool(
 	cosmosPoolConfig.MaxTx = cosmosPoolMaxTx
 	cosmosPool = sdkmempool.NewPriorityMempool(*cosmosPoolConfig)
 
+	// Create the evmMempool
 	evmMempool := &ExperimentalEVMMempool{
 		vmKeeper:      vmKeeper,
 		txPool:        txPool,
@@ -198,6 +183,16 @@ func NewExperimentalEVMMempool(
 		blockGasLimit: config.BlockGasLimit,
 		minTip:        config.MinTip,
 		anteHandler:   config.AnteHandler,
+	}
+
+	// Set up broadcast function
+	if config.BroadCastTxFn != nil {
+		legacyPool.BroadcastTxFn = config.BroadCastTxFn
+	} else {
+		// Create default broadcast function using clientCtx.
+		// The EVM mempool will broadcast transactions when it promotes them
+		// from queued into pending, noting their readiness to be executed.
+		legacyPool.BroadcastTxFn = evmMempool.defaultBroadcastTxFn
 	}
 
 	vmKeeper.SetEvmMempool(evmMempool)
@@ -215,6 +210,11 @@ func (m *ExperimentalEVMMempool) GetBlockchain() *Blockchain {
 // This provides direct access to the EVM-specific transaction management functionality.
 func (m *ExperimentalEVMMempool) GetTxPool() *txpool.TxPool {
 	return m.txPool
+}
+
+// SetClientCtx sets the client context provider for broadcasting transactions
+func (m *ExperimentalEVMMempool) SetClientCtx(clientCtx client.Context) {
+	m.clientCtx = clientCtx
 }
 
 // Insert adds a transaction to the appropriate mempool (EVM or Cosmos).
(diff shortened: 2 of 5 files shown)


aljo242:
@jhelison thanks for opening - looks like we have some CI failing :)

mempool/mempool.go
```diff
@@ -227,7 +230,7 @@ func (m *ExperimentalEVMMempool) Insert(goCtx context.Context, tx sdk.Tx) error
 		hash := ethMsg.Hash()
 		m.logger.Debug("inserting EVM transaction", "tx_hash", hash)
 		ethTxs := []*ethtypes.Transaction{ethMsg.AsTransaction()}
-		errs := m.txPool.Add(ethTxs, true)
+		errs := m.txPool.Add(ethTxs, false)
```
vladjdk: This makes sense

mempool/check_tx.go
```diff
@@ -25,6 +26,13 @@ func NewCheckTxHandler(mempool *ExperimentalEVMMempool) types.CheckTxHandler {
 					return sdkerrors.ResponseCheckTxWithEvents(err, gInfo.GasWanted, gInfo.GasUsed, anteEvents, false), nil
 				}
 			}
+			// If its already known, this can mean the the tx was promoted from nonce gap to valid
+			// and by allowing ErrAlreadyKnown to be silent, we allow re-gossiping of such txs
+			// this also covers the case of re-submission of the same tx enforcing overpricing for replacement
+			if errors.Is(err, txpool.ErrAlreadyKnown) {
```
vladjdk, line 33: I would still like to handle this on the RPC side to prevent users from getting null responses when sending duplicate transactions via some client like Forge. Can we check whether a same-hash transaction exists on the RPC before sending broadcasting it to throw the same error?

evmd/mempool.go
```diff
@@ -39,7 +40,7 @@ func (app *EVMD) configureEVMMempool(appOpts servertypes.AppOptions, logger log.
 		app.EVMKeeper,
 		app.FeeMarketKeeper,
 		app.txConfig,
-		app.clientCtx,
+		func() client.Context { return app.clientCtx }, // Enforces a fresh client context
```
vladjdk: Could we instead do the following:
- Remove the setter for the app itself
- Add a SetClientContext(...) in the mempool
- Set the client context for the mempool [here](https://github.com/cosmos/evm/blob/a689881309d1a54f040f67a6d16587de367ae9bf/server/start.go#L459) (and handle nil mempools)

vladjdk (changes requested):
Hey Jhelison, thanks so much for the work on hardening the mempool! Overall, I think your discovery and fixes make a lot of sense. I left just a couple comments about your implementation details.

I think if we can make the context setting cleaner from the get-go in `start.go` and handle already existing transaction errors on the RPC side, we should be good to go. 

There seem to also be errors in the unit tests (probably from a nil context in the broadcast.

evmd/mempool.go
jhelison: You can find this implemented here: https://github.com/cosmos/evm/pull/890/commits/26677d690cc4927d5280abe4ba7c7c74bfb43ebf

Let me know about the solution there. Its weird that `startStandAlone` had the initialization for the App to set the client. Do you know why?

mempool/check_tx.go
jhelison: During my investigation, duplicated transactions are still returning AlreadyKnown. This is caused by the broadcasting process not reaching this check_tx:
- The TX is [found in cache](https://github.com/cometbft/cometbft/blob/a5726c1be6d49761e6b3933fd01edbaacce9bb56/mempool/clist_mempool.go#L257-L270) before reaching the check tx. And the error on duplicated messages is still the same
- So the UX is unchanged for duplicates

But yeah, I do agree that this check is weird here. Moving it also has some challenges:
- We would need to make the check somewhere else that also affects the [clist_mempool](https://github.com/cometbft/cometbft/blob/a5726c1be6d49761e6b3933fd01edbaacce9bb56/mempool/clist_mempool.go#L413). Since we need to bypass it to enforce a new broadcast for an already existing message (or promoted in our case).

Open to feedback — let me know if you have any suggestions or see a better approach for this case

vladjdk: I'm fine to merge this in the meantime and add a systemtest in a future PR—added a task for a test here: #895.

jhelison:
Glad to hear!
Apologies for not adding further unit and integration tests and refining things even more. Things have been rushed on my side these last few days

PR #891, opened 2025-12-11
feat(mempool): Record metrics for tx latency in mempool

mattac21:
# Description

Adds a `txTracker` to evm mempool that will record the duration between events happening as txs move through the evm mempool. The main timings now being tracked are:
1.  `chainInclusionLatency`: How long it takes for a transaction to go from initially being tracked to being included on chain.
2. `queuedInclusionLatency`: How long it takes for a transaction to go from initially being tracked to being included in queued.
3. `pendingInclusionLatency`: How long it takes for a transaction to go from initially being tracked to being included in pending.
4. `queuedDuration`: How long a transaction is in the queued pool for before exiting. Only recorded on exit (if a tx stays in the pool forever, this will not be recorded).
5. `pendingDuration`: How long a transaction is in the pending pool for before exiting. Only recorded on exit (if a tx stays in the pool forever, this will not be recorded).

Note this only tracks txs that are ingested from an RPC on this node. If it was received via gossip, the tx will not be added to the trackers map, then all other functions will error (which will simply be ignored).

This uses the sdk's new otel telemetry, so it is required to set `OTEL_EXPERIMENTAL_CONFIG_FILE` in order to initialize the prometheus server. 

Closes: STACK-1980

---

## Author Checklist

I have...

### DIFF
--- a/mempool/txpool/legacypool/legacypool.go
+++ b/mempool/txpool/legacypool/legacypool.go
@@ -1589,6 +1589,7 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address, reset *txp
 		for _, tx := range caps {
 			hash := tx.Hash()
 			pool.all.Remove(hash)
+			pool.markTxRemoved(tx, Queue)
 			queueRemovalMetric(RemovalReasonCapExceeded).Mark(1)
 			log.Trace("Removed cap-exceeding queued transaction", "hash", hash)
 		}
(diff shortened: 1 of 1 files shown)


server/start.go
```diff
@@ -605,10 +605,12 @@ func openTraceWriter(traceWriterFile string) (w io.Writer, err error) {
 	)
 }
 
-func startTelemetry(cfg cosmosevmserverconfig.Config) (*telemetry.Metrics, error) {
+func startTelemetry(cfg cosmosevmserverconfig.Config) (*telemetry.Metrics, error) { //nolint:staticcheck
+	//nolint:staticcheck
```
mattac21: new sdk version deprecates all existing telemetry infra that would be a larger lift to change right now, I just did //nolint to all deprecations

swift1337 (approved):
Improved linter supressions a bit

Issue #895, opened 2025-12-12
Add systemtest for mempool duplicates


vladjdk:
ref: [https://github.com/cosmos/evm/pull/890#discussion_r2612488167](https://github.com/cosmos/evm/pull/890#discussion_r2612488167)

Verify that the behaviour returned from duplicates is the same as expected on a geth node

PR #897, opened 2025-12-12
fix(mempool): duplicate tx check and gossipping systemtests

vladjdk:
closes: #895, #896, #898 

In #890, we fixed transaction rebroadcasting by ignoring the "already found" error in the mempool. Completely ignoring the error will still fail out transaction duplicates when sent from the client, but the RPC returns an erroneous Comet mempool cache error instead: `"failed to send eth legacy tx: failed to broadcast transaction: : tx already in mempool [cosmossdk.io/errors@v1.0.2/errors.go:77]"`. This PR fixes that and emits the correct "already" found error by checking for txpool existence on the JSON-RPC side. Since rebroadcasting is done via a Cosmos transaction, this means that we can have the best of both worlds—correct erroring on the JSON-RPC side, as well as proper transaction gossipping of promoted queued transactions.

This PR also fixes the systemtests that check for broadcasting of queued transactions. For each broadcasted transaction, we ensure that every node in the systemtest receives each pending transaction through *gossipping*. Previous tests were passing because transactions were being propagated via committed blocks. This is incorrect, and hid our rebroadcasting errors found in #889. This is fixed by enforcing that each test case completes within the same block.

We also test the rebroadcasting function by first submitting a queued transaction that gets promoted only after a nonce gap is filled. The test ensures that the previously queued transactions becomes available to every other node via gossip after being promoted.

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -426,7 +426,7 @@ func (m *ExperimentalEVMMempool) HasEventBus() bool {
 }
 
 // Has returns true if the transaction with the given hash is already in the mempool.
-// This checks both the legacy pool and the main tx pool for EVM transactions.
+// This checks tx pool for EVM transactions, which iterates through all pools (currently only legacypool)
 func (m *ExperimentalEVMMempool) Has(hash common.Hash) bool {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()


tests/systemtests/mempool/test_exceptions.go
```diff
@@ -10,90 +10,6 @@ import (
 	"github.com/test-go/testify/require"
 )
 
-func RunTxRebroadcasting(t *testing.T, base *suite.BaseTestSuite) {
```
vladjdk, line 13: Removed this test. It doesn't actually test for rebroadcasting, and instead it waits for transactions to be committed.

mempool/mempool.go
```diff
@@ -424,6 +425,16 @@ func (m *ExperimentalEVMMempool) HasEventBus() bool {
 	return m.eventBus != nil
 }
 
+// Has returns true if the transaction with the given hash is already in the mempool.
+// This checks both the legacy pool and the main tx pool for EVM transactions.
+func (m *ExperimentalEVMMempool) Has(hash common.Hash) bool {
```
vladjdk, line 430: This fn allows us to get tx existence on the rpc

PR #909, opened 2025-12-16
feat(mempool): incremental recheck

mattac21:
# Description

This adds the ability for proposals to be created out of txs in the pending pool while the pool is resetting via a new block. Previously, this resetting would hold the lock on the pool for the entire duration of the reset, which would need to call recheck on every tx in the pending pool. This may take multiple seconds, which will block a call to `Pending` and therefore block `PrepareProposal`. 

This PR addresses this by pushing txs into a `txCollector` as they are validated via `demoteUnexecutables`. Creating a 'snapshot' of the pending txs that have been validated for a height at any given time. This snapshot no longer shares the same global legacypool lock, thus unblocking `PrepareProposal`'s call to `Pending`. 

This PR also adds functionality where the reorg loop will be cancelled if it is still processing a reset (new block seen), and a new block arrives. This is to prevent the case where `demoteUnexecutables` takes such a long time to process txs that processing the previous height, always takes longer than the timeout the caller is willing to wait in `Pending`, thus preventing any txs from ever making it into a proposal.

This PR also addresses an issue existing in `feat/krakatoa` where the chain would essentially slow down heavily once tx load increased. This was due to tx gossip directly calling `legacypool.Add` which needed to acquire a lock shared with the reorg loop, which may take 5-10s at peak load in order to be released. This would cause a backup of other critical p2p messages, typically resulting in high prevote times. To fix this, a `insertQueue` has been added in order to decouple `legacypool.Add` from the hot path of p2p message processing.

Closes: STACK-2008

---

## Author Checklist

I have...

### DIFF
--- a/mempool/inserter_test.go
+++ b/mempool/inserter_test.go
@@ -126,11 +126,8 @@ func TestInsertQueue_SlowAddition(t *testing.T) {
 	pool := newMockTxPool()
 
 	// Make Add slow to allow queue to back up
-	var addCalled sync.WaitGroup
-	addCalled.Add(1)
 	pool.setAddFn(func(txs []*ethtypes.Transaction, sync bool) []error {
-		addCalled.Done()
-		time.Sleep(200 * time.Millisecond)
+		time.Sleep(10 * time.Second)
 		return []error{nil}
 	})
 
@@ -142,8 +139,7 @@ func TestInsertQueue_SlowAddition(t *testing.T) {
 	tx1 := ethtypes.NewTransaction(1, [20]byte{0x01}, nil, 21000, nil, nil)
 	iq.Push(tx1)
 
-	// Wait for Add to be called
-	addCalled.Wait()
+	time.Sleep(100 * time.Millisecond)
 
 	// Push a bunch of transactions and verify that we did not have to wait for
 	// the 200 ms to add the first tx.


mempool/txpool/legacypool/pending_builder.go
```diff
@@ -0,0 +1,216 @@
...
+
+	resetAt time.Time
+
+	done chan struct{}
+}
+
+// newPendingBuilder creates a new instance of a pendingBuilder.
+func newPendingBuilder(minTip *big.Int, priceBump uint64) *pendingBuilder {
+	pb := &pendingBuilder{
+		in:                     make(chan validatedTxs),
```
swift1337: I don't think we need this channel as Add* calls are still blocking

mattac21: good point youre right

mempool/mempool.go
```diff
@@ -599,22 +593,26 @@ func (m *ExperimentalEVMMempool) getEVMMessage(tx sdk.Tx) (*evmtypes.MsgEthereum
...
+	sdkctx := sdk.UnwrapSDKContext(ctx)
+	committedHeight := sdkctx.BlockHeight() - 1
+	baseFee := m.vmKeeper.GetBaseFee(sdkctx)
 	var baseFeeUint *uint256.Int
 	if baseFee != nil {
 		baseFeeUint = uint256.MustFromBig(baseFee)
 	}
 
-	evmPendingTxs := m.txPool.Pending(txpool.PendingFilter{
+	ctx, cancel := context.WithTimeout(ctx, time.Millisecond*100)
```
swift1337: Q: Can we fetch consensus params from the context and replace `ms` with something that correlates with the configured propose timeout?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1398,7 +1386,7 @@ func (pool *LegacyPool) scheduleReorgLoop() {
 }
 
 // runReorg runs reset and promoteExecutables on behalf of scheduleReorgLoop.
-func (pool *LegacyPool) runReorg(done chan struct{}, reset *txpoolResetRequest, dirtyAccounts *accountSet, events map[common.Address]*SortedMap) {
+func (pool *LegacyPool) runReorg(done chan struct{}, cancelReset chan struct{}, reset *txpoolResetRequest, dirtyAccounts *accountSet, events map[common.Address]*SortedMap) {
```
swift1337: Should we replace these args with `reorgInput{}`?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1512,17 +1505,32 @@ func (pool *LegacyPool) resetInternalState(newHead *types.Header, reinject types
...
 
 	// Inject any transactions discarded due to reorgs
 	log.Debug("Reinjecting stale transactions", "count", len(reinject))
 	core.SenderCacher().Recover(pool.signer, reinject)
 	pool.addTxsLocked(reinject)
 }
 
+// isResetCancelled returns true if the pool is resetting and it has been
+// signaled to cancel the reset.
+func (pool *LegacyPool) isReorgCancelled(reset *txpoolResetRequest, cancelled chan struct{}) bool {
```
swift1337: This can be moved to a standalone function as it doesn't use the `pool`

mempool/txpool/legacypool/tx_collector.go
```diff
@@ -0,0 +1,289 @@
...
+			pending[addr] = lazies
+		}
+	}
+
+	txsCollected.Mark(int64(numSelected))
+	return pending
+}
+
+// add adds txs to the current set.
+func (t *txs) Add(addr common.Address, txs types.Transactions) {
```
swift1337, line 291: It's possible to write a duplicate here. How does the caller ensure that no duplicates are written?

mempool/txpool/legacypool/legacypool.go
mattac21: yeah there is a lot i can consolidate to a struct

mempool/mempool.go
mattac21: yes good idea, still trying to figure out what a good value for this is

mempool/txpool/legacypool/tx_collector.go
mattac21: hm you're right. I think a scenario like:
* new block arrives
* new tx arrives that replaces a tx in the pending queue (this will get added to txs struct via markTxPromoted)
* reorg loop runs with reset != nil
* tx that just replaced the tx in pending is validated again during demote unexecutables and added to txs structure again
* ahhh

I can add in protection to make sure no duplicates are added.

mempool/mempool.go
mattac21: looking at implementing this now and this isn't possible actually without a larger comet + sdk change to pass that value up. The timeouts are contained in the consensus config, not consensus params that the app has access to. I'll expose this timeout as an app configuration option so that we can modify it in tests.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -518,6 +527,24 @@ func (pool *LegacyPool) Reset(oldHead, newHead *types.Header) {
...
+// requestCancelReset sends a request for a current running pool reset (if any)
+// to be cancelled.
+func (pool *LegacyPool) requestCancelReset() {
+	select {
+	case pool.reqCancelResetCh <- struct{}{}:
+		return
+	case <-pool.reorgShutdownCh:
+		return
+	}
+}
```
Eric-Warehime, line 540: Can't this just be one function?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -911,7 +889,8 @@ func (pool *LegacyPool) add(tx *types.Transaction) (replaced bool, err error) {
 
 		// tx went straight to the pending queue and bypassed the queue, mark
 		// it as promoted
-		pool.markTxPromoted(tx)
+		// TODO: we need to run recheck here! this is a way to bypass recheck
```
Eric-Warehime: Can you explain this more--I don't understand the failure case you're calling out here.

mempool/inserter.go
```diff
@@ -0,0 +1,126 @@
...
+		return
+	}
+
+	iq.lock.Lock()
+	iq.queue.PushBack(tx)
+	iq.lock.Unlock()
+
+	// signal that there are txs available
+	select {
+	case iq.signal <- struct{}{}:
```
Eric-Warehime, line 64: Won't this block inserts on the loop consuming this message?

mempool/txpool/legacypool/legacypool.go
mattac21: yeah, I was just following the pattern of how the other ways of sending messages to the reorg loop are implemented already

mempool/inserter.go
mattac21: no since there is a default case right below this. If this tries to push onto the `signal` channel and it is going to block it will go to the default case and which does nothing and the function will return. So the logic is, if there is no pending signal in the `signal` chan, push a new one. However if there already is a signal in the `signal` chan (telling the `loop` that there are txs available), do nothing.

mempool/txpool/legacypool/legacypool.go
mattac21: yeah (and i'll add this to the todo comment too). When we call `markTxPromoted`, internally that adds the tx to the `validPendingTxs` list. Since the typical path to call this is when a tx goes from `queued` -> `pending`, and when that happens we know that recheck has been run on the tx. However this is the only case where we send a tx straight into `pending`, without ever being in `queued` first and being promoted via `promoteExecutables` (which runs recheck). So the way to bypass recheck here is a user sends a valid tx that passes recheck and goes from `queued` -> `pending`. That tx stays in pending for some reason, then they send another tx at the same nonce with a higher price. This logic will then trigger and we will replace that tx in pending with this new tx, **without** first validating the tx via anteHanlders, so theoretically this replacement tx could be invalid and cause the proposal to fail.

mattac21: There are likely some pretty straightforward ways to get around this, maybe in this specific case we just dont add to the valid txs set, so that this tx will still get gossiped, but it wont be possible for the tx to be included in the next block. Then it will only be added to the valid txs set once it passes recheck in demoteUnexecutabels against the next blocks state.

mattac21: actually I do think that would work, ill just implement that.

mattac21: done here https://github.com/cosmos/evm/pull/909/commits/1b125147d553b2aad35ee1840b288b4bc800a5ce

mempool/txpool/legacypool/legacypool.go
mattac21: done here https://github.com/cosmos/evm/pull/909/commits/0d01f26f80b4c4de7caa762377d2f828ce4f516b

PR #915, opened 2025-12-19
feat(mempool): remove redundant locks

swift1337:
Removes mutex calls from high-level mempool as underlying mempools are already thread-safe.

Local tests showed that `insertEVM` took up to ~10ms for the lock, and SelectBy up to ~80ms, which is a lot for consensus progression.

**Note**: `txpool.Add()` is still slow because it also has a lock inside geth's part of the codebase and also perform some operations under this lock. Local bench showed that it might take up to ~10ms for this operation, 
https://github.com/cosmos/evm/blob/48bde581a1764ff24b3c8d8bff74d41a4c30be50/mempool/txpool/legacypool/legacypool.go#L1095

Overall, `abci.InsertTx()` takes ~30ms 

**UPD**: added a couple of minor perf & styling improvements

Closes STACK-2019

### DIFF
--- a/x/vm/types/config.go
+++ b/x/vm/types/config.go
@@ -72,6 +72,8 @@ func SetChainConfig(cc *ChainConfig) error {
 	}
 
 	chainConfig = config
+
+	gethChainConfig = nil
 	gethChainConfig = config.EthereumConfig(nil)
 
 	return nil


mattac21:
system tests are failing in a different spot than they are on main krakatoa branch

PR #932, opened 2026-01-13
feat: parallel gassends

Eric-Warehime:
# Description

Removes the mint/burn in the statedb in favor of just directly calling the SetBalance function on the bank keeper.
This avoids having to invoke Mint/Burn for gas sends which allows us to parallelize native gas send txs by avoiding hitting the module account with minting/burning across every txn.

Requires: https://github.com/cosmos/cosmos-sdk/pull/25771

### DIFF
--- a/evmd/go.mod
+++ b/evmd/go.mod
@@ -387,4 +387,4 @@ replace (
 
 replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260206152034-6d38fc76f4f7
 
-replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.54.0-rc.1.0.20260206164408-ca81dc06295a
+replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.54.0-rc.1.0.20260206171802-11b322099d1a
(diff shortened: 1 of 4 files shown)


evmd/go.mod
```diff
@@ -388,5 +388,3 @@ replace (
 )
 
 replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260107213545-b26b7e46bb3e
-
-replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.54.0-rc.1.0.20251212034845-02e84c8b3abf
```
swift1337: Let's keep this "as is" for _krakatoa_. Creates fewer conflicts in mod files and allow to edit `replace` section
```go
go mod edit -replace github.com/cosmos/cosmos-sdk=github.com/cosmos/cosmos-sdk@$commit
```

go.mod
```diff
@@ -382,6 +382,4 @@ replace (
 
 replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260107213545-b26b7e46bb3e
 
-replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.54.0-rc.1.0.20251212034845-02e84c8b3abf
```
swift1337: same here

PR #945, opened 2026-01-23
refactor: replace Split in loops with more efficient SplitSeq

slightsharp:
# Description

strings.SplitSeq (introduced in Go 1.23)  returns a lazy sequence (strings.Seq), allowing gopher to iterate over tokens one by one without creating an intermediate slice.

It significantly reduces memory allocations and can improve performance for long strings.

More info: https://github.com/golang/go/issues/61901

---

## Author Checklist

I have...

### DIFF
--- a/.github/workflows/bsr-push.yml
+++ b/.github/workflows/bsr-push.yml
@@ -1,9 +1,16 @@
 name: Push to Buf Schema Registry
-# This workflow runs when a new version tag is pushed to the repository.
-# It then pushes the Protobuf files corresponding to that tag on to the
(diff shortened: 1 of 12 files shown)


ethereum/eip712/types.go
```diff
@@ -342,16 +342,16 @@ func typesAreEqual(types1 []apitypes.Type, types2 []apitypes.Type) bool {
 func sanitizeTypedef(str string) string {
```
aljo242, line 342: Is this function tested?

vladjdk: It's not tested - @slightsharp mind adding some tests for this before making changes?

Issue #952, opened 2026-01-26
shouldUseEVM could be slightly optimized


technicallyty:
shoudlUseEVM function [(mempool/iterator.go:142) ](<https://github.com/cosmos/evm/blob/9c3e1fbc8cc57672376135bef30a5e9ad717937d/mempool/iterator.go#L142>)calls:<br>

```
	nextEVMTx, evmFee := i.getNextEVMTx()
	nextCosmosTx, cosmosFee := i.getNextCosmosTx()
```

however, one of it's callers, `getPreferredTransaction`, already has these variables loaded. another one of it's callers: `advanceCurrentIterator` doesn't. 

i think we could make a small perf optimization by allowing `getPreferredTransaction`to pass in its values.

PR #955, opened 2026-01-27
perf(mempool): avoid redundant tx lookups in shouldUseEVM

mmsqe:
# Description

Closes: https://github.com/cosmos/evm/issues/952

---

## Author Checklist

I have...


mempool/iterator.go
```diff
@@ -125,17 +125,22 @@ func (i *EVMMempoolIterator) Tx() sdk.Tx {
...
-	nextEVMTx, evmFee := i.getNextEVMTx()
-	nextCosmosTx, cosmosFee := i.getNextCosmosTx()
+func (i *EVMMempoolIterator) shouldUseEVM(
+	nextEVMTx *txpool.LazyTransaction, evmFee *uint256.Int,
+	nextCosmosTx sdk.Tx, cosmosFee *uint256.Int,
+) bool {
+	if nextEVMTx == nil || evmFee == nil || nextCosmosTx == nil || cosmosFee == nil {
+		nextEVMTx, evmFee = i.getNextEVMTx()
+		nextCosmosTx, cosmosFee = i.getNextCosmosTx()
+	}
```
technicallyty: the nil checks below (L146 & L150) suggest that some of these may be nil anyways, so its possible this change could make things worse in some cases? maybe shouldUseEVM should just always expect that it has the values supplied.

mempool/iterator.go
```diff
@@ -223,7 +236,7 @@ func (i *EVMMempoolIterator) getPreferredTransaction(nextEVMTx *txpool.LazyTrans
 
 // advanceCurrentIterator advances the appropriate iterator based on which transaction was used
 func (i *EVMMempoolIterator) advanceCurrentIterator() {
-	useEVM := i.shouldUseEVM()
+	useEVM := i.shouldUseEVM(nil, nil, nil, nil)
```
technicallyty: wrt above comment, could just fetch here and pass in.

mempool/iterator.go
mmsqe: Just simplified callers with tuple return now.

mempool/iterator.go
```diff
@@ -192,57 +167,92 @@ func (i *EVMMempoolIterator) getNextCosmosTx() (sdk.Tx, *uint256.Int) {
...
+				return evmTx, txSourceEVM
 			}
 		}
 		// Fall back to Cosmos if EVM is not available or conversion fails
 		i.logger.Debug("EVM transaction conversion failed, falling back to Cosmos transaction")
-		return nextCosmosTx
+		if nextCosmosTx != nil {
+			return nextCosmosTx, txSourceCosmos
+		}
+		return nil, txSourceEVM
```
technicallyty, line 252: seems confusing that nil tx returns sourceEvm, on line 185 we treated it as sourceUnknown. seems like a special case. needs docs or updated code to be more clear

mempool/iterator.go
```diff
@@ -20,14 +20,29 @@ import (
 
 var _ mempool.Iterator = &EVMMempoolIterator{}
 
+type evmTxIterator interface {
+	Peek() (*txpool.LazyTransaction, *uint256.Int)
+	Shift()
+	Empty() bool
+}
```
technicallyty, line 34: this needs docs

technicallyty (commented):
overall lgtm

PR #960, opened 2026-01-29
chore: update local_node.sh to set mempool type to "app"

technicallyty:
# Description

running ./local_node was not properly setting config.toml value for mempool

---

## Author Checklist

I have...

### DIFF
--- a/local_node.sh
+++ b/local_node.sh
@@ -338,3 +338,13 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
     echo "pending mode is on, please wait for the first block committed."
   fi
 fi
+
+# Start the node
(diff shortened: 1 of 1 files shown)


local_node.sh
```diff
@@ -337,13 +338,3 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
     echo "pending mode is on, please wait for the first block committed."
   fi
 fi
-
-# Start the node
-evmd start "$TRACE" \
```
mattac21, line 342: Do we no longer want to start or was this by accident

mattac21 (approved):

local_node.sh
technicallyty: oops

PR #966, opened 2026-01-30
feat(mempool): mempool reserver system

technicallyty:
# Description

### **What does this do?**

This PR takes go-ethereum's existing `ReservationTracker` and brings it to the top level mempool so it can be used in both the Cosmos mempool and the legacypool. 

The ReservationTracker ensures an account can only have txs in one pool at a time.

### **Why?**

It lets each pool recheck without having to check on the other pool's state. We can send a tx to the legacy pool without the legacy pool needing to check the Cosmos pool and V/V. A pool invalidating a tx doesn't need to worry about implications of txs in other pools.

Closes: STACK-2092

---

## Author Checklist

I have...

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -9,7 +9,6 @@ import (
 	"testing"
 	"time"
 
-	mempool2 "github.com/cosmos/cosmos-sdk/types/mempool"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -37,6 +36,7 @@ import (
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
 	"github.com/cosmos/cosmos-sdk/testutil"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	mempooltypes "github.com/cosmos/cosmos-sdk/types/mempool"
 	signingtypes "github.com/cosmos/cosmos-sdk/types/tx/signing"
 	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 )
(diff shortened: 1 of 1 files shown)


mempool/mempool.go
```diff
@@ -517,15 +549,20 @@ func (m *ExperimentalEVMMempool) removeCosmosTx(ctx context.Context, tx sdk.Tx,
...
+	m.cosmosReserver.Release(evmAddrs...) //nolint:errcheck // ignoring is fine here.
+
 	return nil
 }
 
 // shouldRemoveFromEVMPool determines whether an EVM transaction should be manually removed.
 func (m *ExperimentalEVMMempool) shouldRemoveFromEVMPool(hash common.Hash, reason sdkmempool.RemoveReason) bool {
 	if reason.Error == nil {
-		return false
+		return true
```
mattac21: why the change? previously we are not doing this since this means that every tx in finalize block will call mempool.remove which needs to acquire the legacypool lock which is expensive and slows down finalize block a ton

technicallyty: because of this: https://github.com/cosmos/evm/blob/886bb3756fc679676149d60154efed11bb92f31d/mempool/mempool.go#L484-L490

i was attempting to remove a tx using the mempool interface in a test here: https://github.com/cosmos/evm/blob/182aef4746c3bec16221ec06f1a605e12b131855/mempool/mempool_test.go#L70

but it doesn't give it an error, so before, that method could never actually remove eth txs from the mempool

mattac21: mmm ok I see, that is done on purpose actually but it is very non obvious lol

we have it setup like this for eth txs since in finalize block if a tx executes successfully we will try and remove it from the mempool via `mempool.Remove`. This actually isn't good for eth txs since the legacypool will then demote all future pending txs that are coming after this tx that was just removed. This causes a bunch of thrash back and forth between the pending and queued pools for evm txs. So instead we just dont remove on successful execution (`mempool.Remove`), and let the pool internally figure out what should be removed based on what txs now have too low of a nonce, and then it will not demote all txs that come after this one.

mattac21: I guess that will still throw off your test tho, maybe you could just call m.RemoveWithReason and give a reason that will force it to be removed?

PR #967, opened 2026-01-30
fix: Fix err ack in ibc v2 middleware

Eric-Warehime:
# Description

Fixes the err ack return value in the ibc v2 middleware for erc20s.

Previously we were swallowing the return value of the keeper invocation. After this change we properly return an error ack if the app's `OnRecvPacket` call passes, but the middleware keeper's `OnRecvPacket` call fails.

Adde test covers this case.

---

## Author Checklist

I have...


evmd/tests/ibc/v2_ibc_middleware_test.go
```diff
@@ -942,3 +943,101 @@ func (suite *MiddlewareV2TestSuite) TestOnTimeoutPacketNativeErc20() {
 		})
 	}
 }
+
+func (suite *MiddlewareV2TestSuite) TestOnRecvPacketNativeERC20Rollback() {
```
vladjdk, line 947: Would be nice to have a comment on what you did exactly to make it fail at the ERC20 step

x/erc20/v2/ibc_middleware.go
```diff
@@ -86,8 +86,11 @@ func (im IBCMiddleware) OnRecvPacket(
 			Status: channeltypesv2.PacketStatus_Failure,
 		}
 	}
-	im.keeper.OnRecvPacket(ctx, packet, ack)
-	return recvResult
+	modifiedAck := im.keeper.OnRecvPacket(ctx, packet, ack)
+	if !modifiedAck.Success() {
+		return channeltypesv2.RecvPacketResult{Status: channeltypesv2.PacketStatus_Failure}
+	}
+	return channeltypesv2.RecvPacketResult{Status: recvResult.Status, Acknowledgement: modifiedAck.Acknowledgement()}
```
srdtrk, line 96: Based on my understanding, `im.keeper.OnRecvPacket` never actually modifies the ack (unless it returns an error acknowledgement). Therefore, after `modifiedAck.Success()`. It might probably be safer to check that modifiedAck matches original ack exactly.

srdtrk (approved):
Lgtm. Left a hardening suggestion, I'll leave it up to you to implement it or not. I can verify that this fixes the original issue

x/erc20/v2/ibc_middleware.go
Eric-Warehime: Implemented.

PR #972, opened 2026-02-02
fix(backend): Krakatoa EVM `SendRawTransaction` compatibility

mattac21:
# Description

With the addition of the insert queue this has broken EVM compatibility that the EVM RPC `SendRawTransaction` should return an error when adding to the mempool fails. 

This PR fixes that behavior by adding `sub: chan error` param that callers can pass into the insert queue in order to subscribe to the result of adding the tx into the mempool later. EVM RPC's like `SendRawTransaction` will now wait on this result, instead of immediately returning when pushing onto the queue. This does however raise a new issue. If we have unbounded insert queue size, we may have thousands of network connections waiting for txs to make it through the insert queue when the insert queue is large and slow to precess txs.

To get around this we add a new configuration parameter `insert-queue-size` to control the maximum allowed amount of txs in the insert queue at once. If the amount is >= this parameter, txs are rejected at RPC time (or p2p gossip time). Callers should use this as a retryable error code.

To txs being inserted over p2p, we still use the fully asynchronous `InsertAsync` function inside of the `InsertTxHandler` that CometBFT will call. This will now only return a retry-able error code if the tx was unable to be inserted onto the insert queue. Any other error it will be treated as non retry-able.

Closes: #STACK-2095

---

## Author Checklist

I have...

### DIFF
--- a/mempool/inserter.go
+++ b/mempool/inserter.go
@@ -107,7 +107,7 @@ func (iq *insertQueue) loop() {
 		// if nothing is available, wait for new txs to become available before
 		// checking again
 		if numTxsAvailable == 0 {
-			if iq.hasNewTxs() {
+			if iq.waitForNewTxs() {
 				continue
 			}
 			return
@@ -121,6 +121,7 @@ func (iq *insertQueue) loop() {
 		iq.lock.Lock()
 		for item := range iq.queue.IterPopFront() {
 			if item.tx == nil {
+				close(item.sub)
 				continue
 			}
 
@@ -137,21 +138,20 @@ func (iq *insertQueue) loop() {
 			close(subscriptions[i])
 		}
 
-		if numTxsAvailable-len(toInsert) > 0 {
-			// there are still txs available, try and insert immediately
-			// again, unless cancelled
-			select {
-			case <-iq.done:
-				return
-			default:
-				continue
-			}
+		// check if we have been told to cancel, if not, check for more txs to
+		// insert
+		select {
+		case <-iq.done:
+			return
+		default:
+			continue
 		}
 	}
 }
 
-func (iq *insertQueue) hasNewTxs() bool {
-	// no txs available, block until signaled or done
+// waitForNewTxs blocks and waits for new txs to become available and returns
+// true if that happens, or false if we have cancelled before then
+func (iq *insertQueue) waitForNewTxs() bool {
 	select {
 	case <-iq.done:
 		return false
--- a/mempool/inserter_test.go
+++ b/mempool/inserter_test.go
@@ -169,19 +169,24 @@ func TestInsertQueue_RejectsWhenFull(t *testing.T) {
 	iq := newInsertQueue(pool, 5, logger)
 	defer iq.Close()
 
-	// Fill the queue to capacity
-	// Note: The first tx will be immediately popped and start processing (where it blocks),
-	// so we need to push maxSize + 1 transactions to actually fill the queue
-	for nonce := uint64(0); nonce <= 5; nonce++ {
-		tx := ethtypes.NewTransaction(nonce, [20]byte{byte(nonce + 1)}, nil, 21000, nil, nil)
-		_ = iq.Push(tx)
-	}
+	// This first tx will be immediately popped and start processing (where it
+	// blocks)
+	nonce := uint64(0)
+	tx := ethtypes.NewTransaction(nonce, [20]byte{byte(nonce + 1)}, nil, 21000, nil, nil)
+	_ = iq.Push(tx)
+	nonce++
 
 	// wait for first tx to be popped and addFn to be called and blocking
 	<-added
 
+	// Fill the queue to capacity
+	for ; nonce <= 5; nonce++ {
+		tx := ethtypes.NewTransaction(nonce, [20]byte{byte(nonce + 1)}, nil, 21000, nil, nil)
+		_ = iq.Push(tx)
+	}
+
 	// Try to push one more transaction with error channel, queue is now at max capacity
-	tx := ethtypes.NewTransaction(100, [20]byte{0x64}, nil, 21000, nil, nil)
+	tx = ethtypes.NewTransaction(100, [20]byte{0x64}, nil, 21000, nil, nil)
 	_ = iq.Push(tx)
 
 	// Push another tx into the full queue, should be rejected


server/json_rpc.go
```diff
@@ -24,13 +23,14 @@ import (
...
 	"github.com/cosmos/cosmos-sdk/server"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
 const shutdownTimeout = 200 * time.Millisecond
 
 type AppWithPendingTxStream interface {
 	RegisterPendingTxListener(listener func(common.Hash))
-	InsertTx(req *abci.RequestInsertTx) (*abci.ResponseInsertTx, error)
+	GetContextForCheckTx(txBytes []byte) sdk.Context
```
mattac21, line 33: The rpc backend doesnt want to use the abci `InsertTx` method from the `baseapp` anymore, since that obfuscates the errors returned by `mempool.Insert` and wraps it within an error code. We want to return the raw errors to the clients. So instead of calling `abci.InsertTx`, we use the reference to the mempool that the backend already has a call `mempool.Insert` directly on that. We now need to get a context to run `mempool.Insert` on though, so we need to provide `GetContextForCheckTx` to the backend.

mempool/inserter.go
```diff
@@ -32,31 +38,53 @@ type insertQueue struct {
...
+// Push enqueues a tx to eventually be added to the pool. The sub param, if non
+// nil, must be buffered with capacity 1, and will be used to notify the caller
+// of the txs insertion status, once it is processed through the InsertQueue
+// and added to the TxPool. The InsertQueue will close the sub after this
+// happens. If no error occurs, a nil error will be pushed to the sub.
+func (iq *insertQueue) Push(tx *ethtypes.Transaction, sub chan<- error) {
 	if tx == nil {
 		return
 	}
+	if sub != nil && cap(sub) != 1 {
```
mattac21: if the caller does not supply us with a channel that is cap 1, panic, since we dont want to block the insert queue on sending these notifications.

mempool/mempool.go
```diff
@@ -276,34 +286,46 @@ func (m *ExperimentalEVMMempool) Insert(ctx context.Context, tx sdk.Tx) error {
...
-		}
-		if errs[0] != nil {
-			m.logger.Error("error inserting evm tx into pool", "tx_hash", ethTx.Hash(), "err", errs[0])
-		}
-		return errs[0]
+		// we push the tx onto the insert queue so the tx will be inserted at a
+		// later point. We pass along the `subscription` chan, so that the
+		// insert queue will notify the caller of any errors that occurred when
+		// inserting into the mempool.
+		m.iq.Push(ethTx, subscription)
```
mattac21: all evm txs now use the insert queue, no matter if using `Insert` vs `InsertAsync`, the async behavior just changes how you wait for the result of the insert (async meaning dont wait for it to reach the mempool, non async meaning it will wait for it to reach the mempool before returning).

rpc/backend/backend.go
```diff
@@ -184,7 +183,7 @@ type Opt func(*Backend)
 // Application represents ABCI application itself (us).
 // This is used for opaque ABCI calls instead of injecting BaseApp into Backend.
 type Application interface {
-	InsertTx(req *abci.RequestInsertTx) (*abci.ResponseInsertTx, error)
+	GetContextForCheckTx(txBytes []byte) sdk.Context
```
mattac21, line 186: see https://github.com/cosmos/evm/pull/972/changes#r2755983187

server/config/config.go
```diff
@@ -175,18 +175,26 @@ type MempoolConfig struct {
 	GlobalQueue uint64 `mapstructure:"global-queue"`
 	// Lifetime is the maximum amount of time non-executable transaction are queued
 	Lifetime time.Duration `mapstructure:"lifetime"`
+	// PendingTxProposalTimeout is the amount of time to spend waiting for
+	// rechecking of the mempool to complete when creating a proposal
+	PendingTxProposalTimeout time.Duration `mapstructure:"pending-tx-proposal-timeout"`
```
mattac21, line 180: realized we didnt have this config setup for `PendingTxProposalTimeout`, adding it now

x/vm/types/config_testing.go
```diff
@@ -19,6 +20,9 @@ import (
 // opcodes are active based on Ethereum upgrades.
 var testChainConfig *ChainConfig
 
+// testChainConfigMu protects concurrent access to testChainConfig
+var testChainConfigMu sync.RWMutex
```
mattac21, line 24: Not 100% sure why we were not seeing data races on this before, but resetting this at the beginning of tests while having a legacypool running reading the evm coin info and evm chain config was causing data races in the tests, adding these locks fixes that.

mempool/txpool/legacypool/tx_collector.go
```diff
@@ -112,14 +112,15 @@ func (c *txCollector) Collect(ctx context.Context, height *big.Int, filter txpoo
 
 	for {
 		c.mu.RLock()
+		currentHeight := new(big.Int).Set(c.currentHeight)
```
mattac21, line 115: fixing another race I noticed in the tests

server/server_app_options.go
```diff
@@ -147,15 +147,22 @@ func GetLegacyPoolConfig(appOpts servertypes.AppOptions, logger log.Logger) *leg
...
-		// we don't want to return 0 here, as then appOpts.Get() will return nil and that will be
-		// "accidentally" cast to the correct evm max tx default of 0, thereby hiding the error
 		logger.Error("app options is nil, using pending tx proposal timeout of 0 (unlimited)")
 		return 0
 	}
 
 	return cast.ToDuration(appOpts.Get(srvflags.EVMMempoolPendingTxProposalTimeout))
 }
 
+func GetInsertQueueSize(appOpts servertypes.AppOptions, logger log.Logger) int {
```
swift1337:
```suggestion
func GetMempoolInsertQueueSize(appOpts servertypes.AppOptions, logger log.Logger) int {
```

server/json_rpc.go
swift1337: Should we keep the API for non-Krakatoa setups? Or do you want to tackle this closer to the release in the following PRs?

rpc/backend/sign_tx.go
```diff
@@ -101,6 +101,26 @@ func (b *Backend) SendTransaction(args evmtypes.TransactionArgs) (common.Hash, e
...
+		// returned via codes, and we would like to have the full error to
+		// return to clients.
+		err := b.Mempool.Insert(b.Application.GetContextForCheckTx(txBytes), tx)
+		if err != nil {
+			// no need for special error handling like in the broadcast tx case
+			// since this is coming directly from the evm mempool insert.
+			return common.Hash{}, err
+		}
+
+		if err := b.Mempool.TrackTx(txHash); err != nil {
```
swift1337, line 118: Why do we need two separate calls of `Insert` and `TrackTx`? What would happen if we only call `Insert`?

mempool/inserter.go
```diff
@@ -17,11 +18,16 @@ type TxPool interface {
 	Add(txs []*ethtypes.Transaction, sync bool) []error
 }
 
+type item struct {
```
swift1337:
```suggestion
type insertItem struct {
```

mempool/mempool.go
```diff
@@ -276,34 +286,46 @@ func (m *ExperimentalEVMMempool) Insert(ctx context.Context, tx sdk.Tx) error {
...
+	default:
+		// result was not ready immediately, return nil while async things happen
+		return nil
+	}
 }
 
-func (m *ExperimentalEVMMempool) insert(ctx context.Context, tx sdk.Tx, sync bool) error {
-	ethMsg, err := evmTxFromCosmosTx(tx)
+func (m *ExperimentalEVMMempool) insert(ctx context.Context, tx sdk.Tx) chan error {
+	subscription := make(chan error, 1)
```
swift1337: Can we remodel this a bit so that only `insertQueue` controls the channel's lifecycle? It's better to isolate this responsibility in a single entity. eg
```go
sub := m.iq.Push(ethTx, async bool)
```

rpc/backend/sign_tx.go
mattac21: If we didnt call track tx here, the metrics for tx inclusion would not work. We need to start 'tracking' a tx right when it is ingested via rpc. We can't put the call inside of insertTx since that is also called by comet and we dont want to track those txs (only 'local' txs should be tracked).

mattac21: which actually makes me realize that now that insert tx waits for txs to make it through the insert queue fully, we should start tracking txs before the call to insert tx happens, and remove on error. since if we track later, then a potentially long delay inside the insert queue would be hidden from the metrics.

mempool/mempool.go
mattac21: I agree with you and I actually tried to implement it like this originally, but this impl is also quite weird, I'll push it up and lmk what you think. It's weird since we have a synchronous error bring returned by a possible parsing error or when inserting into the cosmos mempool, and then an async error from the evm insert queue. so the return value is (chan error, error) which I dont love, and it makes handling it on the callers end a bit more confusing imo. but lmk what your thoughts are, I could go either way on this.

rpc/backend/sign_tx.go
mattac21: fixed here https://github.com/cosmos/evm/pull/972/commits/280765761e313ae4c22ebe4a720c88fba663bde6

mempool/mempool.go
mattac21: changed here https://github.com/cosmos/evm/pull/972/commits/52ae8743d792d2433e66d4f08983e6e32f35c0b0

mattac21:
> Should we keep the API for non-Krakatoa setups? Or do you want to tackle this closer to the release in the following PRs?

The InsertTx method that im removing from this interface was only added for Krakato, so this isn't a break (but I guess adding anything to this interface technically is, but I doubt anyone else is implementing `AppWithPendingTxStream`). Although I do agree we should make it something different when we do the pr to make krakatoa fully feature flag-able on and off, but that's for later.

rpc/backend/call_tx.go
```diff
@@ -151,10 +148,24 @@ func (b *Backend) SendRawTransaction(data hexutil.Bytes) (common.Hash, error) {
...
 	txHash := tx.Hash()
 
-	// publish tx directly to app-side mempool, avoiding broadcasting to consensus layer.
+	// publish tx directly to app-side mempool, avoiding broadcasting to
+	// consensus layer.
 	if b.UseAppMempool {
-		if err := b.mempoolInsertTx(txHash, txBytes); err != nil {
-			return b.handleSendTxError(tx, ethSigner, err)
+		// track the tx for tx inclusion timing metrics of local txs
+		if err := b.Mempool.TrackTx(txHash); err != nil {
```
swift1337, line 155: imho, the name is a bit confusing because it sounds like something whose absence would break the mempool, but it's actually more of an "analytical" or "observability" tool.

Consider: `StartTrackingTx(hash)`


also, I think this method should not return anything and log an error within itself

mattac21: > imho, the name is a bit confusing because it sounds like something whose absence would break the mempool, but it's actually more of an "analytical" or "observability" tool.
> 
> Consider: StartTrackingTx(hash)

down to consider renaming, but that should be in another pr. ill create a productionization ticket to do this.

mattac21: > also, I think this method should not return anything and log an error within itself

I dont agree with this point, imo it's almost always better to return the error to the caller. What if the caller cares so much about tracking txs that they would want to reject txs outright from the rpc if they failed to be tracked? if we remove the error from the return value, then it would be impossible for a caller to implement that behavior.

PR #984, opened 2026-02-06
chore(heightsync): Refactor `TxCollector` to generic `HeightSync` struct for cosmos & evm tx collection

mattac21:
# Description

Refactors the existing `TxCollector` used by the `legacypool` into its own package (`HeightSync`) and makes it generic of the things that it is storing, so that it can be used by the `legacypool` and `cosmospool` (and potentially other pools later too).

Closes: STACK-2288

---

## Author Checklist

I have...

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,6 +48,8 @@ require (
 	github.com/tidwall/sjson v1.2.5
 	github.com/tyler-smith/go-bip39 v1.1.0
 	github.com/zondax/hid v0.9.2
+	go.opentelemetry.io/otel v1.39.0
+	go.opentelemetry.io/otel/metric v1.39.0
 	go.uber.org/mock v0.6.0
 	golang.org/x/crypto v0.47.0
 	golang.org/x/net v0.49.0
@@ -317,7 +319,6 @@ require (
 	go.opentelemetry.io/contrib/otelconf v0.19.0 // indirect
 	go.opentelemetry.io/contrib/propagators/b3 v1.39.0 // indirect
 	go.opentelemetry.io/contrib/propagators/jaeger v1.39.0 // indirect
-	go.opentelemetry.io/otel v1.39.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploggrpc v0.15.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlplog/otlploghttp v0.15.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc v1.39.0 // indirect
@@ -330,7 +331,6 @@ require (
 	go.opentelemetry.io/otel/exporters/stdout/stdoutmetric v1.39.0 // indirect
 	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.39.0 // indirect
 	go.opentelemetry.io/otel/log v0.15.0 // indirect
-	go.opentelemetry.io/otel/metric v1.39.0 // indirect
 	go.opentelemetry.io/otel/sdk v1.39.0 // indirect
 	go.opentelemetry.io/otel/sdk/log v0.15.0 // indirect
 	go.opentelemetry.io/otel/sdk/metric v1.39.0 // indirect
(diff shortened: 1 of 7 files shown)


internal/heightsync/heightsync.go
```diff
@@ -0,0 +1,171 @@
...
+	// that the HeightSync had not yet reached.
+	hsHeightBehind = metrics.NewRegisteredMeter("height_sync/heightbehind", nil)
+
+	// hsDuration is the total time callers of Get spend from invocation to
+	// return.
+	hsDuration = metrics.NewRegisteredTimer("height_sync/duration", nil)
+)
+
+// HeightSync manages per height access to an instance of a value V.
+type HeightSync[V any] struct {
```
mattac21: This is all of the non 'tx collection' related logic moved out of the tx collector and into its own struct that is decoupled with 'tx collecting'. It is now just a thing you can use to synchronize access to an instance of a generic value at any a height. Currently, this is the `TxStore` for evm txs in the legacypool (i.e. a new `TxStore` is created at every height, and putting the `TxStore` in a `HeightSync` allows for the pool to easily manage a per height instance of the `TxStore` and for callers to be able to 'wait' for the correct height to be reached if the pool is running behind the chains current height, etc (all the same tx collector functionality).

mempool/txpool/legacypool/tx_store.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+var (
+	// txsCollected is the total amount of txs returned by Collect.
+	txsCollected = metrics.NewRegisteredMeter("legacypool/txstore/txscollected", nil)
+)
+
+// TxStore is a set of transactions at a height that can be added to or
+// removed from.
+type TxStore struct {
```
mattac21, line 22: this is the tx collection specific logic from the tx collector refactored out into its own structure. this is specific to evm txs, and we will implement another for cosmos txs now.

mempool/tx_store.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"sync"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
+)
+
+// CosmosTxStore is a set of cosmos transactions that can be added to or
+// removed from.
+type CosmosTxStore struct {
```
mattac21, line 12: the new cosmos tx specific store that we will manage via `HeightSync`

mempool/tx_store.go
```diff
@@ -0,0 +1,105 @@
...
+
+// Txs returns a copy of the current set of txs in the store.
+func (s *CosmosTxStore) Txs() []sdk.Tx {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return append([]sdk.Tx(nil), s.txs...)
+}
+
+// Iterator returns an sdkmempool.Iterator over the txs in the store.
+func (s *CosmosTxStore) Iterator() sdkmempool.Iterator {
```
mattac21, line 69: This iterator is 'dumb' and iterates over the txs in the order that they were put into the store. that means that when we use this for rechecking, we must insert (and therefore recheck) cosmos txs in their 'execution' order, or the order that would have been returned by the actual mempool implementations iterator. So we will recheck by first getting the mempool iterator via `mempool.Select`, then iterate over that, recheck each tx, and if it passes, put it in this `TxStore`.

mempool/internal/heightsync/heightsync.go
technicallyty, line 1: maybe worth considering that this could just go mempool/internal. admittedly was a little confused at first what this was for with the `value` and `vFactory` stuff, so i my first reaction was that this might be over-generic-ized without much reason. if anything, maybe `value` can be `state` and `vFactory` could be `resetState` 🤷🏻 

ofc, this is just a small organizational nit. overall the impl is great ^^

mattac21: agree with this 100%, as I was writing I also felt like it was over generic-izing, love the the idea of mempool/internal and then making it a bit more mempool specific since its within the mempool/, instead of a root internal/. will change

mattac21: have moved things around and made the height sync a bit more specific to what it is doing for the mempool https://github.com/cosmos/evm/pull/984/commits/d1f789c1460ed9b3dfedb38c92e38ff2c83f5802, lmk what you think.

PR #991, opened 2026-02-09
feat(mempool): cosmos rechecker

technicallyty:
# Description

- Adds a rechecking background loop to the cosmos pool. Block-drive like the legacy pool
- Wraps the ExtMempool to better encapsulate cosmos pool management

Closes: STACK-2287

---

## Author Checklist

I have...

### DIFF
--- a/evmd/go.mod
+++ b/evmd/go.mod
@@ -385,6 +385,6 @@ replace (
 	github.com/tidwall/btree => github.com/cosmos/btree v0.0.0-20250924232609-2c6195d95951
 )
 
-replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260210154435-f833026c47ba
+replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260212220946-a6184c7d834d
 
 replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.54.0-rc.1.0.20260206171802-11b322099d1a
--- a/evmd/go.sum
+++ b/evmd/go.sum
@@ -258,8 +258,8 @@ github.com/cockroachdb/redact v1.1.6/go.mod h1:BVNblN9mBWFyMyqK1k3AAiSxhvhfK2oOZ
 github.com/cockroachdb/tokenbucket v0.0.0-20250429170803-42689b6311bb h1:3bCgBvB8PbJVMX1ouCcSIxvsqKPYM7gs72o0zC76n9g=
 github.com/cockroachdb/tokenbucket v0.0.0-20250429170803-42689b6311bb/go.mod h1:7nc4anLGjupUW/PeY5qiNYsdNXj7zopG+eqsS7To5IQ=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd/go.mod h1:sE/e/2PUdi/liOCUjSTXgM1o87ZssimdTWN964YiIeI=
-github.com/cometbft/cometbft v0.39.0-beta.2.0.20260210154435-f833026c47ba h1:mqUaNeYem221Qwa90MCN8flCAmP4tvl/xe7//5B4x+Y=
-github.com/cometbft/cometbft v0.39.0-beta.2.0.20260210154435-f833026c47ba/go.mod h1:Q6T7e2zLsng7uxG+d5b4+YnmuxyUeztEPcrab6gcV5E=
+github.com/cometbft/cometbft v0.39.0-beta.2.0.20260212220946-a6184c7d834d h1:HfteAUFCSdu1n8Db5udqOj+/isThkh/eTVvy8k0mSDs=
+github.com/cometbft/cometbft v0.39.0-beta.2.0.20260212220946-a6184c7d834d/go.mod h1:Q6T7e2zLsng7uxG+d5b4+YnmuxyUeztEPcrab6gcV5E=
 github.com/cometbft/cometbft-db v0.14.1 h1:SxoamPghqICBAIcGpleHbmoPqy+crij/++eZz3DlerQ=
 github.com/cometbft/cometbft-db v0.14.1/go.mod h1:KHP1YghilyGV/xjD5DP3+2hyigWx0WTp9X+0Gnx0RxQ=
 github.com/consensys/gnark-crypto v0.18.1 h1:RyLV6UhPRoYYzaFnPQA4qK3DyuDgkTgskDdoGqFt3fI=
--- a/go.mod
+++ b/go.mod
@@ -378,7 +378,7 @@ replace (
 	github.com/tidwall/btree => github.com/cosmos/btree v0.0.0-20250924232609-2c6195d95951
 )
 
-replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260210154435-f833026c47ba
+replace github.com/cometbft/cometbft => github.com/cometbft/cometbft v0.39.0-beta.2.0.20260212220946-a6184c7d834d
 
 replace github.com/cosmos/cosmos-sdk => github.com/cosmos/cosmos-sdk v0.54.0-rc.1.0.20260206171802-11b322099d1a
 
--- a/go.sum
+++ b/go.sum
@@ -253,8 +253,8 @@ github.com/cockroachdb/redact v1.1.6/go.mod h1:BVNblN9mBWFyMyqK1k3AAiSxhvhfK2oOZ
 github.com/cockroachdb/tokenbucket v0.0.0-20250429170803-42689b6311bb h1:3bCgBvB8PbJVMX1ouCcSIxvsqKPYM7gs72o0zC76n9g=
 github.com/cockroachdb/tokenbucket v0.0.0-20250429170803-42689b6311bb/go.mod h1:7nc4anLGjupUW/PeY5qiNYsdNXj7zopG+eqsS7To5IQ=
 github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd/go.mod h1:sE/e/2PUdi/liOCUjSTXgM1o87ZssimdTWN964YiIeI=
-github.com/cometbft/cometbft v0.39.0-beta.2.0.20260210154435-f833026c47ba h1:mqUaNeYem221Qwa90MCN8flCAmP4tvl/xe7//5B4x+Y=
-github.com/cometbft/cometbft v0.39.0-beta.2.0.20260210154435-f833026c47ba/go.mod h1:Q6T7e2zLsng7uxG+d5b4+YnmuxyUeztEPcrab6gcV5E=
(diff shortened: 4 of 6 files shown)


mempool/mempool.go
```diff
@@ -686,3 +629,24 @@ func (m *ExperimentalEVMMempool) getIterators(ctx context.Context, txs [][]byte)
...
 	return m.txTracker.Track(hash)
 }
+
+// RecheckCosmosTxs triggers a synchronous recheck of cosmos transactions.
+// This is primarily used for testing.
+func (m *ExperimentalEVMMempool) RecheckCosmosTxs() {
+	m.cosmosPool.TriggerRecheckSync()
+}
+
+type signerSequence struct {
```
mattac21: this should move into `recheck_pool.go` since its only used there

mempool/mempool.go
```diff
@@ -686,3 +629,24 @@ func (m *ExperimentalEVMMempool) getIterators(ctx context.Context, txs [][]byte)
...
+	account string
+	seq     uint64
+}
+
+func Equal(s1, s2 signerSequence) bool {
```
mattac21: I may be missing this in some `==` case but I dont think this is used

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+
+	"cosmossdk.io/log"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
+	authsigning "github.com/cosmos/cosmos-sdk/x/auth/signing"
+)
+
+var (
+	meter           = otel.Meter("github.com/cosmos/evm/mempool")
```
mattac21: nit but should we define the meter for the mempool in `mempool.go` instead of the `recheck_pool.go`? im assuming all other metrics we want to define in this package via otel would share this `meter` instance, right?

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+	anteHandler sdk.AnteHandler
+	getCtx      func() (sdk.Context, error)
+	logger      log.Logger
+
+	// event channels
+	reqRecheckCh    chan struct{}
+	recheckDoneCh   chan chan struct{}
+	shutdownCh      chan struct{}
+	shutdownOnce    sync.Once
+	recheckShutdown chan struct{} // closed when scheduleRecheckLoop exits
```
mattac21: I think we should have comments outlining how each of these works, I know when first learning how the legacypool works this channel sending a channel that would be used to wait on the completion of the recheck is pretty confusing at first glance.

mempool/mempool.go
```diff
@@ -197,12 +191,20 @@ func NewExperimentalEVMMempool(
 	cosmosPoolConfig.MaxTx = cosmosPoolMaxTx
 	cosmosPool = sdkmempool.NewPriorityMempool(*cosmosPoolConfig)
 
+	// Wrap cosmos pool with recheck functionality
+	recheckPool := NewRecheckMempool(
+		logger,
+		cosmosPool,
+		tracker.NewHandle(-1),
+		config.AnteHandler,
+		blockchain.GetLatestContext,
```
mattac21, line 209: instead of directly using the `anteHandler` and a `getCtx` callback, could we use the same `rechecker` (but another instance of it) that the legacypool uses? just so those recheck in essentially the same way? it would have to move to its own package to avoid the import cycle I think. I'm not super tied to this tho if this would make it much more complicated.

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+
+// Start begins the background recheck loop.
+func (m *RecheckMempool) Start() {
+	m.wg.Add(1)
+	go m.scheduleRecheckLoop()
+}
+
+// Close gracefully shuts down the recheck loop.
+func (m *RecheckMempool) Close() error {
+	m.shutdownOnce.Do(func() {
```
mattac21, line 115: do we expect people to call `Close` more than once? or is this just an added protection?

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+	})
+	m.wg.Wait()
+	return nil
+}
+
+// Insert adds a transaction to the pool after running the ante handler.
+// This is the main entry point for new cosmos transactions.
+func (m *RecheckMempool) Insert(goCtx context.Context, tx sdk.Tx) error {
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
mattac21: as the worlds smallest optimization, both `signerAddressesFromTx` and the `reserver.Hold` are thread safe, so you could lock the `RecheckMempool` down on line 121 instead of holding the lock for those operations as well.

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+	// Reserve addresses to prevent conflicts with EVM pool
+	addrs, err := signerAddressesFromTx(tx)
+	if err != nil {
+		return err
+	}
+	if err := m.reserver.Hold(addrs...); err != nil {
+		return err
+	}
+
+	if _, err := m.anteHandler(sdk.UnwrapSDKContext(goCtx), tx, false); err != nil {
```
mattac21, line 137: do we need this still? I think once we add in the tx collector piece from my pr then we will guarantee (even without this check) that all txs being proposed have passed recheck, just based on the recheck being done in the recheck loop.

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+	defer m.mu.Unlock()
+
+	if err := m.ExtMempool.Remove(tx); err != nil {
+		return err
+	}
+
+	addrs, err := signerAddressesFromTx(tx)
+	if err != nil {
+		m.logger.Error("failed to extract signer addresses for release", "err", err)
+		return nil // tx was removed, just log the error
```
mattac21: should we panic here? if we had the tx in the mempool, we would have already had to have successfully called `signerAddressFromTx(tx)` on it, so I don't think it should be possible to fail here unless something super weird happens. also if this does error and we return nil without releasing their reservation, the user will never be able to add any evm txs since the cosmos pool will hold their reservation forever, right?

mattac21: or maybe if they send another cosmos tx, then remove that one, it would release their reservation

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+// runRecheck performs the actual recheck work. It holds the write lock for the
+// entire duration, iterates through all txs, runs them through the ante handler,
+// and removes any that fail (plus dependent txs with higher sequences).
+func (m *RecheckMempool) runRecheck(done chan struct{}, cancelled <-chan struct{}) {
+	defer close(done)
+	start := time.Now()
+	txsRemoved := 0
+	defer func() {
+		recheckDuration.Record(context.Background(), float64(time.Since(start).Milliseconds()),
+			metric.WithAttributes(attribute.Int("txs_removed", txsRemoved)))
```
mattac21: may be misunderstanding how otel works, but I think `txs_removed` would be nice to have as a standalone counter metric we can view instead of as an attribute attached to a histogram metric.

mempool/recheck_pool.go
```diff
@@ -0,0 +1,388 @@
...
+		m.logger.Debug("recheck cancelled before removal - new block arrived")
+		return
+	}
+
+	for _, txn := range removeTxs {
+		if err := m.ExtMempool.Remove(txn); err != nil {
+			m.logger.Error("failed to remove tx during recheck", "err", err)
+			continue
+		}
+		addrs, err := signerAddressesFromTx(txn)
```
mattac21, line 341: idk how expensive doing this signature lookup is so maybe a premature optimization suggestion here, but since we are doing this calculation twice, maybe inside of the loop when we lookup the addresses and sequence nums with `extractSignerSequences` we could maintain another map of tx hash -> addresses for txs we are going to remove, since we already have all that info there, and then can remove this call to `signerAddressesFromTx` int eh second loop

mempool/mempool.go
technicallyty: the rechecker seems quite specific to evm. by another instance, did you mean another implementation?

mempool/recheck_pool.go
technicallyty: just makes it idempotent

mempool/recheck_pool.go
technicallyty: Seems like we'd be subjecting ourselves to unnecessary work by putting in a tx that we could check is bad upfront

mempool/recheck_pool.go
technicallyty: yeah i think error cases here would be super duper rare, ill change to panic

mempool/recheck_pool.go
technicallyty: getting the hash from a tx might be even more expensive tbh. the sdk.Tx interface is not really that good. to get the hash we'd need:
- tx encoder
- encode tx to bytes
- sha256 hash them

which i bet is a lot more expensive to do here. i could do this and write a bench to check though, but my gut feeling is it will be worse

mempool/recheck_pool.go
technicallyty: updated

mempool/recheck_pool.go
technicallyty: updated

mempool/recheck_pool.go
technicallyty: added comments

mempool/recheck_pool.go
technicallyty: moved

mempool/mempool.go
technicallyty: wasn't used, deleted it

mempool/recheck_pool.go
mattac21: makes sense, down to keep it as is.

mempool/recheck_pool.go
mattac21: hmm yeah its a bit of a tradeoff. I think that we can keep it actually but probably need to do something similar to how we have evm txs setup (can be done later, we already have a ticket for this). 

Having a heavy-ish check like this in the comet gossip hot path will hurt performance of cosmos txs quite a lot. but also we want to give feedback to rpc users if their tx made it into the mempool or not and was valid. so we will need a way for txs received over p2p to not wait for this ante handler to run before returning back to comet, and for rpcs to wait for this ante handler check. this can come later tho and we can just copy how this is done for evm txs.

mempool/mempool.go
mattac21: i meant the same rechecker struct, just a different instance so they dont concurrently use the same context. I dont think it is that specific to evm, it would just need a new recheck function to recheck cosmos txs rather than evm txs, right?

technicallyty: i looked at rearranging this to work with cosmos txs, and after messing around with it for about 30mins, i feel like this could be its own issue. making it work with cosmos could be done in a quick n dirty way, but to do it more cleanly would require a bit of refactoring: cosmos doesn't need a tx converter, i dont need to pass in a header or legacypool.Blockchain during `Update`. of course i could just make an `UpdateCosmosCtx` style method, but that just doesn't feel great.

the current state of the PR should be safe anywho, the base context is branched and the `write` function is discarded, so it shouldn't ever clobber the ctx that the evm side might have.

mattac21: ok makes sense, appreciate you looking into it!

PR #1002, opened 2026-02-17
chore(queue): Update insert queue to work with cosmos txs

mattac21:
# Description

Updates the insert queue to be generic over the type of tx that we are inserting, create two separate insert queues, one for evm txs and one for cosmos txs. Apply the same insert queue mechanics for cosmos txs that we use for evm txs.

Closes: STACK-2289

---

## Author Checklist

I have...


mempool/mempool.go
```diff
@@ -224,9 +226,32 @@ func NewExperimentalEVMMempool(
...
+			}
+			return errs
+		}
+
+		for i, tx := range txs {
+			errs[i] = evmMempool.insertCosmosTx(ctx, *tx)
+		}
+		return errs
+	}, config.InsertQueueSize, logger)
+	evmMempool.cosmosQueue = cosmosQueue
```
technicallyty: small nit, but could we put each arg on its own line?

PR #1012, opened 2026-02-23
feat(mempool): krakatoa cosmos txs height sync tx store

mattac21:
# Description

This PR integrates the 'height sync tx store' behavior for cosmos txs that already exists for evm txs. As cosmos txs are rechecked in the `RecheckPool`, they will be pushed to a `CosmosTxStore` that is used to collect rechecked txs at a particular height. Once the chain advances to the next height, that tx store is cleared and all txs must be rechecked and pushed into the store again at the new height.

Closes: STACK-2367

---

## Author Checklist

I have...

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -603,11 +603,10 @@ func (m *ExperimentalEVMMempool) SetEventBus(eventBus *cmttypes.EventBus) {
 	}
 	go func() {
 		bc := m.GetBlockchain()
-		currentBlock := bc.CurrentBlock()
 		for range sub.Out() {
 			bc.NotifyNewBlock()
 			// Trigger cosmos pool recheck on new block (non-blocking)
-			m.cosmosPool.TriggerRecheck(currentBlock.Number)
+			m.cosmosPool.TriggerRecheck(bc.CurrentBlock().Number)
 		}
 	}()
 }


mempool/recheck_pool.go
```diff
@@ -179,11 +200,15 @@ func (m *RecheckMempool) CountTx() int {
 	return m.ExtMempool.CountTx()
 }
 
+type recheckRequest struct {
```
mattac21, line 203: copying the same pattern used in the `legacypool`, rechecks must now inform what height we are advancing to, so that we know what height we have rechecked txs on top of, that way the `HeightSync[CosmosTxStore]` knows if it can serve its set of txs for a requested height or not.

mempool/recheck_pool.go
```diff
@@ -193,8 +218,22 @@ func (m *RecheckMempool) TriggerRecheck() <-chan struct{} {
...
+
+// RecheckedTxs returns the txs that have been rechecked for a height. The
+// RecheckMempool must be currently operating on this height (i.e. recheck has
+// been triggered on this height via TriggerRecheck). If height is in the past
+// (TriggerRecheck has been called on height + 1), this will panic. If height
+// is in the future, this will block until TriggerReset is called for height,
+// or the context times out.
+func (m *RecheckMempool) RecheckedTxs(ctx context.Context, height *big.Int) sdkmempool.Iterator {
+	txStore := m.recheckedTxs.GetStore(ctx, height)
+	if txStore == nil {
```
mattac21, line 233: will be nil on restart for very first proposal. when replaying blocks, comet does not publish new height headers so the pool will not be reset at the proposal height.

mempool/tx_store.go
```diff
@@ -38,26 +38,6 @@ func (s *CosmosTxStore) AddTx(tx sdk.Tx) {
 	s.txs = append(s.txs, tx)
 }
 
-// RemoveTx removes a tx from the store using swap-with-last for O(1) removal.
-func (s *CosmosTxStore) RemoveTx(tx sdk.Tx) {
```
mattac21, line 42: unused, and we dont want support this, we should never remove from this store directly

mempool/internal/heightsync/heightsync.go
```diff
@@ -68,8 +68,8 @@ type HeightSync[Store any] struct {
 	mu sync.RWMutex
 }
 
-// NewHeightSync creates a new HeightSync starting at the given height.
-func NewHeightSync[Store any](startHeight *big.Int, reset func() *Store) *HeightSync[Store] {
+// New creates a new HeightSync starting at the given height.
+func New[Store any](startHeight *big.Int, reset func() *Store) *HeightSync[Store] {
```
mattac21, line 72: rename to not stutter package name

mempool/reserver/reserver.go
```diff
@@ -85,13 +85,10 @@ func (h *ReservationHandle) Hold(addrs ...common.Address) error {
 	defer h.tracker.lock.Unlock()
 
 	for _, addr := range addrs {
-		// Double reservations are forbidden even from the same pool to
-		// avoid subtle bugs in the long term.
 		owner, exists := h.tracker.accounts[addr]
 		if exists {
 			if owner == h.id {
-				log.Error("pool attempted to reserve already-owned address", "address", addr)
-				return nil // Ignore fault to give the pool a chance to recover while the bug gets fixed
```
mattac21, line 94: the `legacypool` never calls `Hold` on addresses that it already has reserved since it is implemented as as a map of address -> txs. however for the cosmos pool we have no quick way to look up if an account has txs in the pool, thus we simply remove this error log since this is not a bug for us, we just let the `RecheckPool` call `Hold` on addresses that it already has reserver and let it be a no op.

mempool/mempool.go
```diff
@@ -637,31 +639,46 @@ func evmTxFromCosmosTx(tx sdk.Tx) (*evmtypes.MsgEthereumTx, error) {
 // It configures EVM transactions with proper base fee filtering and priority ordering,
 // while setting up the Cosmos iterator with the provided exclusion list.
 func (m *ExperimentalEVMMempool) getIterators(ctx context.Context, txs [][]byte) (*miner.TransactionsByPriceAndNonce, sdkmempool.Iterator) {
+	selectHeight := new(big.Int).SetInt64(sdk.UnwrapSDKContext(ctx).BlockHeight() - 1)
+	return m.evmIterator(ctx, selectHeight), m.cosmosIterator(ctx, selectHeight)
```
mattac21: tiny refactor to separate evm and cosmos logic

mempool/mempool.go
```diff
@@ -637,31 +640,62 @@ func evmTxFromCosmosTx(tx sdk.Tx) (*evmtypes.MsgEthereumTx, error) {
...
+	selectHeight := new(big.Int).SetInt64(sdk.UnwrapSDKContext(ctx).BlockHeight() - 1)
+	wg.Go(func() {
+		evmIterator = m.evmIterator(ctx, selectHeight)
+	})
+
+	wg.Go(func() {
+		cosmosIterator = m.cosmosIterator(ctx, selectHeight)
+	})
+
+	wg.Wait()
```
mattac21, line 697: fetching these iterators is completely independent, so moved them to goroutines

mattac21: currently they both use the same `PendingTxProposalTimeout`, can separate one for each tx type later if we want

mempool/mempool.go
```diff
@@ -602,7 +605,7 @@ func (m *ExperimentalEVMMempool) SetEventBus(eventBus *cmttypes.EventBus) {
 		for range sub.Out() {
 			m.GetBlockchain().NotifyNewBlock()
 			// Trigger cosmos pool recheck on new block (non-blocking)
-			m.cosmosPool.TriggerRecheck()
+			m.cosmosPool.TriggerRecheck(m.GetBlockchain().CurrentBlock().Number)
```
swift1337: nit: should `m.GetBlockchain()` stay outside of the loop?

mempool/mempool.go
```diff
@@ -670,32 +673,63 @@ func evmTxFromCosmosTx(tx sdk.Tx) (*evmtypes.MsgEthereumTx, error) {
 // getIterators prepares iterators over pending EVM and Cosmos transactions.
 // It configures EVM transactions with proper base fee filtering and priority ordering,
 // while setting up the Cosmos iterator with the provided exclusion list.
-func (m *ExperimentalEVMMempool) getIterators(ctx context.Context, txs [][]byte) (*miner.TransactionsByPriceAndNonce, sdkmempool.Iterator) {
+func (m *ExperimentalEVMMempool) getIterators(ctx context.Context, _ [][]byte) (*miner.TransactionsByPriceAndNonce, sdkmempool.Iterator) {
```
```suggestion
func (m *ExperimentalEVMMempool) getIterators(ctx context.Context) (evm *miner.TransactionsByPriceAndNonce, cosmos sdkmempool.Iterator) {
```

mempool/mempool.go
```diff
@@ -670,32 +673,63 @@ func evmTxFromCosmosTx(tx sdk.Tx) (*evmtypes.MsgEthereumTx, error) {
...
+	var (
+		evmIterator    *miner.TransactionsByPriceAndNonce
+		cosmosIterator sdkmempool.Iterator
+		wg             sync.WaitGroup
+	)
+
+	selectHeight := new(big.Int).SetInt64(sdk.UnwrapSDKContext(ctx).BlockHeight() - 1)
```
swift1337, line 687: Could you clarify why we need `BlockHeight()-1`? Does `BlockHeight()` return uncommitted height?

mempool/recheck_pool.go
```diff
@@ -280,6 +309,9 @@ func (m *RecheckMempool) runRecheck(done chan struct{}, cancelled <-chan struct{
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	m.recheckedTxs.StartNewHeight(height)
+	defer m.recheckedTxs.EndCurrentHeight()
```
swift1337, line 313: Q: Is it possible to have two runRecheck at the same time and `defer m.recheckedTxs.EndCurrentHeight()` for `runRecheck(block=100)` will actually mark `runRecheck(block=101)` as finished?

mempool/mempool.go
mattac21: > Does BlockHeight() return uncommitted height?

exactly yeah, ill add a comment to clarify

mempool/recheck_pool.go
mattac21: this should not be possible and if it does happen it is a bug. I have added some additional panics to protect against this happening.

PR #1015, opened 2026-02-23
feat: cross compilation it works

Eric-Warehime:
Cross compile in the makefile


Makefile
```diff
@@ -103,6 +103,8 @@ build: go.sum $(BUILDDIR)/
 
 # Cross-compile for Linux AMD64
 build-linux:
+	CC="zig cc -target x86_64-linux-gnu" \
```
aljo242, line 106: nope

PR #1019, opened 2026-02-26
chore!: Remove x/precisebank

Eric-Warehime:
# Description

Removes the precisebank module.

Does not include an upgrade handler for such--not sure if we should write one or leave that to each chain in this case.

---

### DIFF
--- a/tests/integration/precompiles/werc20/test_integration.go
+++ b/tests/integration/precompiles/werc20/test_integration.go
@@ -526,11 +526,8 @@ func TestPrecompileIntegrationTestSuite(t *testing.T, create network.CreateEvmAp
 					err = is.precompile.UnpackIntoInterface(&decimals, erc20.DecimalsMethod, ethRes.Ret)
 					Expect(err).ToNot(HaveOccurred(), "failed to unpack result")
 
-					coinInfo := testconstants.ExampleChainCoinInfo[testconstants.ChainID{
-						ChainID:    is.network.GetChainID(),
-						EVMChainID: is.network.GetEIP155ChainID().Uint64(),
-					}]
-					Expect(decimals).To(Equal(uint8(coinInfo.Decimals)), "expected different decimals") //nolint:gosec // G115
+					// Without precisebank, all chains operate natively with 18 decimals
+					Expect(decimals).To(Equal(uint8(evmtypes.EighteenDecimals)), "expected 18 decimals")
 				},
 				)
 			})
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -75,24 +75,3 @@ jobs:
           make test-scripts
         if: env.GIT_DIFF
 
-  test-fuzz:
```
aljo242, line 78: why are we removing this?

aljo242 (approved):

precompiles/werc20/tx.go
```diff
@@ -69,7 +68,7 @@ func (p Precompile) Withdraw(ctx sdk.Context, contract *vm.Contract, stateDB vm.
 	caller := contract.Caller()
 	callerAccAddress := sdk.AccAddress(caller.Bytes())
 	nativeBalance := p.BankKeeper.SpendableCoin(ctx, callerAccAddress, evmtypes.GetEVMCoinDenom())
-	if nativeBalance.Amount.Mul(types.ConversionFactor()).LT(amountInt) {
```
vladjdk, line 72: Just want to note that we still have other ConversionFactor() uses in the repo, and that we should consider removing those and the support for non-18-decimal chains.

tests/integration/precompiles/werc20/test_integration.go
```diff
@@ -552,11 +526,8 @@ func TestPrecompileIntegrationTestSuite(t *testing.T, create network.CreateEvmAp
 					err = is.precompile.UnpackIntoInterface(&decimals, erc20.DecimalsMethod, ethRes.Ret)
 					Expect(err).ToNot(HaveOccurred(), "failed to unpack result")
 
-					coinInfo := testconstants.ExampleChainCoinInfo[testconstants.ChainID{
-						ChainID:    is.network.GetChainID(),
-						EVMChainID: is.network.GetEIP155ChainID().Uint64(),
-					}]
-					Expect(decimals).To(Equal(uint8(coinInfo.Decimals)), "expected different decimals") //nolint:gosec // G115
+					// Without precisebank, all chains operate natively with 18 decimals
```
vladjdk, line 529: We still have a different ConversionFactor in `x/vm/types`, so we can't yet make this statement without fully removing support for non-18-decimal chains

aljo242: follow up issue

precompiles/werc20/tx.go
aljo242: follow up issue

tests/integration/precompiles/werc20/test_integration.go
Eric-Warehime: I shall do this in a followup PR.

PR #1030, opened 2026-03-02
feat!: Krakatoa

Eric-Warehime:
# Description

PR for merging krakatoa feature branch to main.

### DIFF
--- a/mempool/internal/heightsync/heightsync.go
+++ b/mempool/internal/heightsync/heightsync.go
@@ -64,7 +64,7 @@ type HeightSync[Store any] struct {
 	done chan struct{}
 
 	// mu protects all of the above fields; it does not protect internal
-	// fields of the value V itself
+	// fields of the Store itself
 	mu sync.RWMutex
 }
 
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -131,10 +131,6 @@ var evmTracers = []string{"json", "markdown", "struct", "access_list"}
 type Config struct {
 	config.Config `mapstructure:",squash"`
 
-	// UseIAVLX enables the IAVLX storage engine.
-	// todo: move to appropriate place?
-	UseIAVLX bool `mapstructure:"use-iavlx"`
-
 	EVM     EVMConfig     `mapstructure:"evm"`
 	JSONRPC JSONRPCConfig `mapstructure:"json-rpc"`
 	TLS     TLSConfig     `mapstructure:"tls"`
--- a/server/flags/flags.go
+++ b/server/flags/flags.go
@@ -83,9 +83,6 @@ const (
 	EVMMempoolInsertQueueSize          = "evm.mempool.insert-queue-size"
 )
 
-// IAVLXEnable (experimental) enables the IAVLX storage engine.
-const IAVLXEnable = "iavlx-enable"
-
 // TLS flags
 const (
 	TLSCertPath = "tls.certificate-path"


tests/integration/mempool/test_mempool_integration_abci.go
```diff
@@ -82,8 +84,8 @@ func (s *IntegrationTestSuite) TestTransactionOrderingWithABCIMethodCalls() {
 		{
 			name: "mixed EVM and Cosmos transactions with equal effective tips",
 			setupTxs: func() ([]sdk.Tx, []string) {
-				// Cosmos with same effective tip: 1000 * 200000 = 200000000 aatom total fee
-				cosmosTx := s.createCosmosSendTx(s.keyring.GetKey(0), big.NewInt(1000000000)) // 1 gaatom/gas effective tip
+				// Cosmos with same effective tip (use different key to avoid address reservation conflict)
+				cosmosTx := s.createCosmosSendTx(s.keyring.GetKey(9), big.NewInt(1000000000)) // 1 gaatom/gas effective tip
```
mattac21, line 88: feels like we are kind of changing what these tests are testing with these changes. I think theyre trying to test that when we have two txs of different types from the same account at the same nonce with varying gas prices, we will only include a single one of them (I think). but now that we are going from a different account, they will both just always be included.

Eric-Warehime: Yeah maybe, but the new test for that would just be ensuring the reserver denies the second txn--which we have tests for elsewhere. I think the new test ensures that we can handle ordering of cosmos and evm txns based on tip and cosmos versus evm ordering in general.

evmd/app.go
```diff
@@ -7,9 +7,13 @@ import (
 	"io"
 	"os"
 
+	goruntime "runtime"
```
swift1337, line 10:
```suggestion
"runtime"
```

ante/evm/05_signature_verification.go
```diff
@@ -60,11 +60,7 @@ func (esvd EthSigVerificationDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, s
 // that the signer address matches the one defined on the message.
 // The function set the field from of the given message equal to the sender
 // computed from the signature of the Ethereum transaction.
-func SignatureVerification(
-	msg *evmtypes.MsgEthereumTx,
-	ethTx *ethtypes.Transaction,
-	signer ethtypes.Signer,
-) error {
+func SignatureVerification(msg *evmtypes.MsgEthereumTx, _ *ethtypes.Transaction, signer ethtypes.Signer) error {
```
vladjdk, line 63: If we're making other breaking changes, we should just remove this parameter. If the goal is to just not break anyone, then we should leave it.

swift1337 (approved):

mempool/internal/heightsync/heightsync.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// heightChanged is closed when currentHeight is no longer being worked
+	heightChanged chan struct{}
+
+	// done is closed when no more operations will happen on the value at
+	// currentHeight
+	done chan struct{}
+
+	// mu protects all of the above fields; it does not protect internal
+	// fields of the value V itself
```
vladjdk: Value V as in store?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1399,18 +1569,50 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
+		queuedNofundsMeter.Mark(int64(len(costDrops)))
+
+		// Drop all transactions that now fail the pools RecheckTxFn
+		//
+		// Note this is happening after the nonce removal above since this
+		// check is slower, we would like it to happen on the fewest txs as
+		// possible.
+		recheckStart := time.Now()
+		recheckDrops, _ := list.FilterSorted(func(tx *types.Transaction) bool {
+			newCtx, err := pool.rechecker.Recheck(ctx, tx)
```
vladjdk, line 1593: In Geth, when the baseFee climbs higher than the maxFeePerGas, the transactions don't get directly dropped. Instead, they remain in the pending pool until the baseFee climbs back down.

Seems like we're removing them entirely since we check the baseFee in the rechecker now.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1576,36 +1787,78 @@ func (pool *LegacyPool) truncateQueue() {
...
+				ctx = newCtx
+			}
+			if err == nil {
+				// always write changes back to the original context
+				write()
+			}
+			return tolerateRecheckErr(err) != nil
+		})
+
+		for _, tx := range recheckDrops {
```
vladjdk, line 1843: Same comment as above - we're modifying geth mempool behaviour by dropping these

server/config/config.go
```diff
@@ -131,6 +131,10 @@ var evmTracers = []string{"json", "markdown", "struct", "access_list"}
 type Config struct {
 	config.Config `mapstructure:",squash"`
 
+	// UseIAVLX enables the IAVLX storage engine.
+	// todo: move to appropriate place?
+	UseIAVLX bool `mapstructure:"use-iavlx"`
```
vladjdk: We should remove this until IAVLX is ready.

server/flags/flags.go
```diff
@@ -72,15 +72,20 @@ const (
...
+	EVMMempoolGlobalSlots              = "evm.mempool.global-slots"
+	EVMMempoolAccountQueue             = "evm.mempool.account-queue"
+	EVMMempoolGlobalQueue              = "evm.mempool.global-queue"
+	EVMMempoolLifetime                 = "evm.mempool.lifetime"
+	EVMMempoolPendingTxProposalTimeout = "evm.mempool.pending-tx-proposal-timeout"
+	EVMMempoolInsertQueueSize          = "evm.mempool.insert-queue-size"
 )
 
+// IAVLXEnable (experimental) enables the IAVLX storage engine.
+const IAVLXEnable = "iavlx-enable"
```
vladjdk: Ditto

mempool/internal/heightsync/heightsync.go
mattac21: old comment that needs to be updated, it was the old name of the generic parameter

mempool/txpool/legacypool/legacypool.go
mattac21: good callout, we can just add in a small exception to the rechecker on when it should return an error for evm txs.

mattac21: this is the excepted behavior for both pending and queued txs? they both just sit in their respective pools? or would a pending tx be demoted back to queued if it drops below base fee?

vladjdk: Afaiu the pending txes just get filtered out in this spot but remain in pending - I don't think there's ever a way that underpriced (via basefee) transactions get put into queued

ante/evm/05_signature_verification.go
Eric-Warehime: Yeah I think this is just for linting for now. We can remove in another PR I think.

PR #1031, opened 2026-03-03
feat(mempool): shared ctx beween cosmos tx rechecks and inserts

mattac21:
# Description

This PR implements context sharing between inserting cosmos txs and rechecking cosmos txs. 

Previously, inserting cosmos txs and rechecking them happened on independent contexts. This led to errors when inserting txs that were only valid given the context of txs already in the mempool. For example, if acc1 at height h has txs with sequence 2 and 3 in the mempool, and tries to insert sequence 4 at height h+1, previously this would fail with a nonce gap error, since the context that the tx is validated against in insert tx does not know that acc1 already has sequences 2 and 3 in the mempool and its 4 should be valid.

Closes: STACK-2405

---

## Author Checklist

I have...

### DIFF
--- a/mempool/rechecker.go
+++ b/mempool/rechecker.go
@@ -18,7 +18,7 @@ type TxConverter interface {
 	EVMTxToCosmosTx(tx *ethtypes.Transaction) (sdk.Tx, error)
 }
 
-// Rechecker runs recheckFn on pending and queued txs in the pool, given an
+// TxRechecker runs recheckFn on pending and queued txs in the pool, given an
 // sdk context via UpdateCtx.
 //
 // NOTE: None of the recheckers functions are thread safe.
@@ -31,7 +31,7 @@ type TxRechecker struct {
 	txConverter TxConverter
 }
 
-// New creates a new rechecker that can recheck transactions.
+// NewTxRechecker creates a new rechecker that can recheck transactions.
 func NewTxRechecker(anteHandler sdk.AnteHandler, txConverter TxConverter) *TxRechecker {
 	return &TxRechecker{
 		anteHandler: anteHandler,


mempool/mempool.go
```diff
@@ -245,16 +242,8 @@ func NewExperimentalEVMMempool(
 	cosmosQueue := queue.New(
 		func(txs []*sdk.Tx) []error {
 			errs := make([]error, len(txs))
-			ctx, err := blockchain.GetLatestContext()
```
mattac21, line 248: we use the `cosmosRechecker`'s context now that is fetched internally in `insertCosmosTx`, so `insertCosmosTx` doesnt even need to take in a context anymore.

mempool/mempool.go
```diff
@@ -394,11 +383,11 @@ func (m *ExperimentalEVMMempool) insert(_ context.Context, tx sdk.Tx) (<-chan er
 
 // insertCosmosTx inserts a cosmos tx into the cosmos mempool.
 // The RecheckMempool handles ante handler validation, address reservation, and locking internally.
-func (m *ExperimentalEVMMempool) insertCosmosTx(goCtx context.Context, tx sdk.Tx) error {
+func (m *ExperimentalEVMMempool) insertCosmosTx(tx sdk.Tx) error {
 	m.logger.Debug("inserting Cosmos transaction")
 
 	// Insert into cosmos pool (handles locking, ante handler, and address reservation internally)
-	if err := m.cosmosPool.Insert(goCtx, tx); err != nil {
+	if err := m.cosmosPool.Insert(context.Background(), tx); err != nil {
```
mattac21, line 390: this function needs to take a context to conform to the mempool interface of the sdk, but the context passed in is not used, instead the recheckers context is used.

mempool/mempool_test.go
```diff
@@ -54,10 +54,8 @@ func TestMempool_Iterate(t *testing.T) {
 	storeKey := storetypes.NewKVStoreKey("test")
 	transientKey := storetypes.NewTransientStoreKey("transient_test")
 	ctx := testutil.DefaultContext(storeKey, transientKey) //nolint:staticcheck // false positive.
-	anteHandler := func(ctx sdk.Context, tx sdk.Tx, simulate bool) (sdk.Context, error) {
-		return ctx, nil
-	}
```
mattac21, line 59: ante handler overriding was unused, removed it

mempool/mempool_test.go
```diff
@@ -54,10 +54,8 @@ func TestMempool_Iterate(t *testing.T) {
...
-	mp, _, txConfig, _, _, accounts := setupMempoolWithAnteHandler(t, anteHandler, numAccs)
+	s := setupMempoolWithAccounts(t, numAccs)
```
mattac21, line 57: refactored return value of this to be a struct instead of multiple return values

mempool/recheck_pool.go
```diff
@@ -142,16 +175,20 @@ func (m *RecheckMempool) Insert(goCtx context.Context, tx sdk.Tx) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	if _, err := m.anteHandler(sdk.UnwrapSDKContext(goCtx), tx, false); err != nil {
-		m.reserver.Release(addrs...) //nolint:errcheck // best effort cleanup
+	// Branch from the Rechecker's internal ctx (post-recheck cache).
+	// This ctx has chain_state + all pending txs' nonce increments.
+	ctx, write := m.rechecker.GetContext()
+	if _, err := m.rechecker.RecheckCosmos(ctx, tx); err != nil {
```
mattac21, line 185: main change: use the recheckers context here instead of the passed in context. the recheckers context is updated with the rechecks of all the txs already in the mempool

mempool/rechecker.go
```diff
@@ -23,7 +22,7 @@ type TxConverter interface {
 // sdk context via UpdateCtx.
 //
 // NOTE: None of the recheckers functions are thread safe.
-type Rechecker struct {
+type TxRechecker struct {
```
mattac21, line 25: renamed so the interface in the recheck_pool.go can use the name Rechecker (same as legacypool)

evmd/mempool.go
```diff
@@ -42,7 +42,8 @@ func (app *EVMD) configureEVMMempool(appOpts servertypes.AppOptions, logger log.
 	}
 
 	txEncoder := evmmempool.NewTxEncoder(app.txConfig)
-	rechecker := evmmempool.NewRechecker(mempoolConfig.AnteHandler, txEncoder)
+	evmRechecker := evmmempool.NewTxRechecker(mempoolConfig.AnteHandler, txEncoder)
+	cosmosRechecker := evmmempool.NewTxRechecker(mempoolConfig.AnteHandler, txEncoder)
```
aljo242, line 46: feels like these can actually just be created inside the Mempool constructor?

We're providing txEncoder and mempoolConfig

mempool/rechecker.go
```diff
@@ -32,9 +31,9 @@ type Rechecker struct {
 	txConverter TxConverter
 }
 
-// NewRechecker creates a new rechecker that can recheck transactions.
-func NewRechecker(anteHandler sdk.AnteHandler, txConverter TxConverter) *Rechecker {
-	return &Rechecker{
+// New creates a new rechecker that can recheck transactions.
+func NewTxRechecker(anteHandler sdk.AnteHandler, txConverter TxConverter) *TxRechecker {
```
aljo242: mismatch

mempool/rechecker.go
technicallyty, line 25: comment should be TxRechecker now

evmd/mempool.go
technicallyty: i guess we could just have these be the defaults, always enabled in the mempool. then make an option for people to pass in their own impl's of rechecker if they wish.

the `NewExperimentEVMMempool` constructor is starting to look a little thick

mattac21: yeah we could do what tyler is suggesting I like that. That is actually already what the legacypool does with its rechecker (except it's disabled by default, only enabled if you pass in your own rechecker).

mattac21: but I want the rechecker impl to be injectable so you can override it in tests, its very useful to make the rechecker slow/blocking for example and then test that behavior. if you make it in the constructor then we can't do that for a full `ExperimentalEVMMempool` test.

mattac21: im just gonna create a productioniztion ticket to simplify this configuration. I want to move these into the evm config struct and have it rechecking default off if they are nil, but we need the feature flagging work to be done first

mattac21: https://linear.app/cosmoslabs/issue/STACK-2456/simplify-experiemntalevmmempool-constructor

aljo242 (approved):

Issue #1037, opened 2026-03-04
bug: RPCStream initSubscriptions panics on CometBFT subscribe failure, crashing node


Aboudjem:
## Bug Description

In `rpc/stream/rpc.go`, `initSubscriptions()` uses `panic(err)` when CometBFT event subscription fails (lines ~91, ~99). Any transient CometBFT connectivity issue crashes the entire RPC process instead of degrading gracefully.

## Impact

- Network blips, slow starts, or reconnection attempts crash the node permanently
- The node must be manually restarted
- Subscription streams are called lazily via `HeaderStream()`/`LogStream()`, so the panic can happen at any time

## Proposed Fix

Replace `panic(err)` with `s.logger.Error(...)` + `return`. Streams are allocated before subscribe, so callers get non-nil empty streams that degrade gracefully.

PR #1043, opened 2026-03-04
fix(rpc): replace panic with graceful degradation in RPCStream subscriptions

Aboudjem:
# Description

Closes: #1037

## Summary

`initSubscriptions()` in `rpc/stream/rpc.go` uses `panic(err)` when CometBFT event subscription fails. Any transient connectivity issue (network blip, slow CometBFT start) permanently crashes the RPC process instead of degrading gracefully.

## Fix

Replace `panic(err)` with `s.logger.Error(...)` + `return`. Streams are already allocated before subscribe calls, so callers receive non-nil (empty) streams that degrade gracefully — no new data arrives until the next initialization attempt.

This matches how `Close()` already handles the "not initialized" case.

## Author Checklist

### DIFF
--- a/rpc/stream/rpc_test.go
+++ b/rpc/stream/rpc_test.go
@@ -1,0 +1,70 @@
+package stream
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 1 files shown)


Aboudjem:
Tests added in `rpc/stream/rpc_test.go`.

PR #1059, opened 2026-03-08
feat(indexer): enable cosmos module events to be indexed by EVM indexers

### LINKED ISSUE #1000: Enable Cosmos Module Events to be Indexed by EVM Indexers
## Problem

Currently, when Cosmos SDK modules emit events, these events are not visible to EVM-side indexers and explorers because they don't emit `EventTypeEthereumTx` events. This creates a tracking gap for important state changes that involve fund movements.

### Examples

**Staking Module:**
- When unbonding occurs, funds are transferred but the transaction is not trackable in EVM explorers
- When rewards are claimed, the fund movement is invisible to EVM indexers

**Impact:** Users and applications relying on EVM explorers (Etherscan-like tools) cannot track these important transactions, leading to incomplete transaction history.

djm07073:
## Summary

Introduces a Transformer architecture that enables EVM indexers to index events emitted by Cosmos SDK modules (staking unbonding, bank transfer, etc.).

Closes #1000

## Problem

External services relying on EVM indexers often face confusion when certain on-chain activities are not visible through standard EVM RPC endpoints.

For Cosmos EVM chains, events emitted by native Cosmos SDK modules (staking, bank, etc.) do not generate `EventTypeEthereumTx`. This creates a gap where important state changes are invisible to EVM explorers and indexers.

**The core limitation:** The current indexer architecture has no mechanism to add events for non-Ethereum transactions. Only transactions that go through the EVM execution path can be indexed.

**Specific issues:**
- Unbonding completion transactions cannot be tracked in EVM explorers like Etherscan
- Bank transfer events are not visible to EVM indexers  
- Events occurring in BeginBlock/EndBlock are completely missing

This prevents users and external services relying on EVM-based tools from tracking important state changes, causing operational confusion.

## Solution

This PR solves the problem **at the eth indexer layer** by introducing a Transformer architecture. It transforms Cosmos events into EVM-compatible receipts/logs while minimizing core logic modifications.

### Transformer Interface

```go
type CosmosEventTransformer interface {
    CanHandle(eventType string) bool
    Transform(event abci.Event, height int64, ethTxHash common.Hash) (*TransformedTxData, error)
}
```

- Extensible plugin architecture
- Adding support for new Cosmos event types only requires implementing a new transformer

### Example Transformers

**Bank Transfer Transformer**
- `coin_spent` / `coin_received` → ERC20 `Transfer(address,address,uint256)` log

**Staking Unbonding Transformer**
- BeginBlock `complete_unbonding` → `CompleteUnbonding(address,address,uint256)` log

### Backend Integration

The RPC backend (`GetTransactionByHash`, `GetTransactionReceipt`) now checks the KV indexer for pre-built transformed tx/receipt data before falling back to the original `MsgEthereumTx` path:

```go
// Check for pre-built eth receipt (from cosmos event transformers)
if b.Indexer != nil {
    if receiptJSON, err := b.Indexer.GetEthReceipt(hash); err == nil {
        var receipt map[string]interface{}
        if err := json.Unmarshal(receiptJSON, &receipt); err == nil {
            return receipt, nil
        }
    }
}
```

### App Integration

Register transformers in your app's root command:

```go
startOpts := cosmosevmserver.NewDefaultStartOptions(newApp, defaultNodeHome).
    WithIndexerSetup(func(idx *indexer.KVIndexer) {
        // Register bank transfer transformer
        tokenAddress := common.HexToAddress("0x0000000000000000000000000000000000000001")
        idx.RegisterTransformer(evmdindexer.NewBankTransferTransformer(tokenAddress))

        // Register staking unbonding transformer
        stakingPrecompileAddress := common.HexToAddress("0x0000000000000000000000000000000000000800")
        idx.RegisterTransformer(evmdindexer.NewStakingUnbondingTransformer(stakingPrecompileAddress))
    })
```

## Trade-offs

| Aspect | Impact |
|--------|--------|
| **RPC Latency** | Reduced - Pre-built receipts are served directly from kv db without reconstructing from `MsgEthereumTx` of comet rpc |
| **DB Size** | Increased - Additional storage for transformed tx/receipt data |
| **Indexing Time** | Increased - Additional processing for cosmos event transformation |

- [x] Impl: Add Tranformer architecture in eth indexer
- [x] Unit tests: Transformer logic verification
- [x] Integration tests: Indexer operation verification
- [x] System tests: E2E unbonding scenario verification
- [ ]  Benchmark & analyze trade-offs of Transformer-based indexer vs baseline
- [ ]  Add ethereum compatliblity test (eth_getlogs, eth_subscribe)

### DIFF
--- a/server/indexer_cmd.go
+++ b/server/indexer_cmd.go
@@ -16,7 +16,9 @@ import (
 )
 
 // NewIndexTxCmd creates a new Cobra command to index historical Ethereum transactions.
-func NewIndexTxCmd() *cobra.Command {
+// The optional indexerSetup function is called after creating the indexer to register
+// cosmos event transformers, ensuring historical re-indexing captures all transformed events.
+func NewIndexTxCmd(indexerSetup func(*indexer.KVIndexer)) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "index-eth-tx [backward|forward]",
 		Short: "Index historical eth txs",
@@ -49,6 +51,11 @@ func NewIndexTxCmd() *cobra.Command {
 				return err
 			}
 			idxer := indexer.NewKVIndexer(idxDB, logger.With("module", "evmindex"), clientCtx)
+
+			// Register cosmos event transformers if setup function is provided
+			if indexerSetup != nil {
+				indexerSetup(idxer)
+			}
 
 			// open local CometBFT db, because the local rpc won't be available.
 			tmdb, err := cmtconfig.DefaultDBProvider(&cmtconfig.DBContext{ID: "blockstore", Config: cfg})
--- a/server/util.go
+++ b/server/util.go
@@ -54,7 +54,7 @@ func AddCommands(
 		sdkserver.NewRollbackCmd(opts.AppCreator, opts.DefaultNodeHome),
 
 		// custom tx indexer command
-		NewIndexTxCmd(),
+		NewIndexTxCmd(opts.IndexerSetup),
 	)
 }
 


indexer/transformer.go
```diff
@@ -0,0 +1,37 @@
...
+	CanHandle(eventType string) bool
+
+	// Transform converts a cosmos event into EVM receipt/tx data.
+	// ethTxHash is provided by the caller to ensure consistency.
+	Transform(event abci.Event, height int64, ethTxHash common.Hash) (*TransformedTxData, error)
+}
+
+// GenerateTransformedEthTxHash generates a deterministic eth tx hash by concatenating and hashing inputs.
+// Used for cosmos txs (cosmosTxHash) and block phases (phase + blockHash).
+func GenerateTransformedEthTxHash(data ...[]byte) common.Hash {
```
qj0r9j0vc2, line 35: `GenerateTransformedEthTxHash` concatenates all byte slices before hashing, which means different input groupings can produce the same hash. 
```go
GenerateTransformedEthTxHash([]byte("AB"), []byte("C"))  // Keccak256("ABC")
GenerateTransformedEthTxHash([]byte("A"), []byte("BC"))  // Keccak256("ABC") - same!
```

Currently the callers use fixed-length phase strings so this may not be an issue in practice, but it’s fragile for future use.
so, I'd like to suggest you to handle it with length-prefixing each input.

```go 

for _, d := range data {
binary.Write(&buf, binary.BigEndian, uint32(len(d)))
buf.Write(d)
}


```

evmd/indexer/bank_transformer.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+// CanHandle returns true for coin_spent and coin_received events.
+func (t *BankTransferTransformer) CanHandle(eventType string) bool {
+	return eventType == banktypes.EventTypeCoinSpent || eventType == banktypes.EventTypeCoinReceived
+}
+
+// Transform converts a bank event to TransformedTxData with ERC20 Transfer log.
+func (t *BankTransferTransformer) Transform(
```
qj0r9j0vc2, line 44: `coin_spent` and `coin_received` are handled independently, so a single bank Send from Alice to Bob produces two logs.

- Transfer(Alice, 0x0, amount) - looks like a burn
- Transfer(0x0, Bob, amount) - looks like a mint

EVM block explorers would display this as a burn + mint rather than a transfer. Was this intentional, or is there a plan to pair `coin_spent` + `coin_received` events to produce a single `Transfer(Alice, Bob, amount)` log?

indexer/receipt_data.go
```diff
@@ -0,0 +1,162 @@
...
+		logs = []*ethtypes.Log{}
+	}
+
+	// Calculate bloom filter from logs
+	bloom := ethtypes.CreateBloom(&ethtypes.Receipt{Logs: logs})
+
+	return &ethtypes.Receipt{
+		Type:              0,
+		Status:            d.Status,
+		CumulativeGasUsed: d.GasUsed,
```
qj0r9j0vc2: Per the Ethereum spec, CumulativeGasUsed should be the running total of gas used by all transactions up to and including this one in the block - not the individual tx’s gas.
Ref - https://github.com/cosmos/evm/blob/5eca4967d474ea23fe26f71b3f92fb0c2cc2aaad/indexer/kv_indexer_process.go#L240

server/indexer_cmd.go
```diff
@@ -74,7 +74,7 @@ func NewIndexTxCmd() *cobra.Command {
 				if err != nil {
 					return err
 				}
-				if err := idxer.IndexBlock(blk, resBlk.TxResults); err != nil {
+				if err := idxer.IndexBlock(blk, resBlk.TxResults, resBlk.Events); err != nil {
```
qj0r9j0vc2, line 84: The `index-eth-tx` re-indexing command creates a bare KVIndexer without registering any transformers. This means historical re-indexing would miss all cosmos event transformations. 

Should the command also accept the `IndexerSetup` function from `StartOptions`?

qj0r9j0vc2 (changes requested):

evmd/indexer/bank_transformer.go
djm07073: Yes, this was intentional.

I've considered that 1:1 pairing between cosmos events and EVM events could provide more flexibility and enable richer representations. However, implementing proper pairing would require receiving events as a batch rather than processing them individually, which introduces additional complexity.

Whether it's better to receive events as a batch for pairing or to process them individually (as currently designed) is still an open question. For now, we chose the simpler individual event approach.

Note that this `BankTransferTransformer` is just an example implementation. Developers can customize it to fit their needs; for instance, instead of mapping to `Transfer(from, to, amount)`, one could create separate Receive and Spent event representations, or implement a batch-aware transformer that pairs events within a transaction context.

indexer/transformer.go
djm07073: In real cases, the situation you mentioned does not actually occur.
Anyway, it has been applied in [9aeda78](https://github.com/cosmos/evm/pull/1059/commits/9aeda788ac7994b368f4ad599c96285f57860cb7).

indexer/receipt_data.go
djm07073: Applied in [1919766](https://github.com/cosmos/evm/pull/1059/commits/191976626d244d3c019e0013243b104c1ef9bcd5)

server/indexer_cmd.go
djm07073: nice catch, Applied in 
9989a14

indexer/kv_indexer.go
```diff
@@ -34,90 +34,103 @@ var _ servertypes.EVMTxIndexer = &KVIndexer{}
...
-				return errorsmod.Wrapf(err, "IndexBlock %d", height)
-			}
-		}
+	// Process BeginBlock events (single synthetic tx per phase)
+	if err := kv.processBlockPhaseEvents(block, BlockPhaseBeginBlock, beginBlockEvents); err != nil {
+		return err
+	}
+
+	// Process DeliverTx events
+	if err := kv.processDeliverTxEvents(block, txResults); err != nil {
```
mmsqe, line 124: seems need move processDeliverTxEvents before PreBlock to keep transactionIndex match tx log decoding used by eth_getLogs

indexer/kv_indexer_process.go
```diff
@@ -0,0 +1,372 @@
...
+		return nil
+	}
+
+	// Build and save receipt
+	var status uint64 = 1
+	if ethTx.Failed() {
+		status = 0
+	}
+
+	receipt := &ethtypes.Receipt{
```
mmsqe, line 237: Receipt still miss some fields like ContractAddress

PR #1060, opened 2026-03-09
feat(krakatoa): `operate_exclusively` toggles krakatoa features

mattac21:
# Description

This PR adds a flag `evm.mempool.operate-exclusively` that will enable the `KrakatoaMempool` features. These features have been pulled out of the `ExperimentalEVMMempool` and into their own struct in `krakatoa_mempool.go`.

Mempool related system and integration tests now run two copies, they run once on the `ExperimentalEVMMempool` + CometBFT `flood` mempool, and once on the `KrakatoaMempool`+ CometBFT `app` mempool.

One slight modification was made to the `ExperimentalEVMMempool ` though, when interacting with the `legacypool`, both mempools use the lifecycle hooks instead of directly modifying the `legacypool`. The `ExperimentalEVMMempool` calls the `BroadcastTxFn` when a tx is promoted via the `onTxPromoted` hook, instead of passing the `BroadcastTxFn` into the `legacypool` itself.

Closes: STACK-2091

---

## Author Checklist

I have...


mempool/krakatoa_mempool.go
```diff
@@ -0,0 +1,630 @@
...
+		// if we have a result immediately, ready on the channel returned from
+		// insert, return that (cosmos tx or unable to try and insert the tx
+		// due to parsing error).
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+		// result was not ready immediately, return nil while async things happen
+		return nil
+	}
```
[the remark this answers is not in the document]
mattac21: good callout greptile, but this is a known limitation that we are addressing in a different pr!

mempool/internal/queue/queue.go
```diff
@@ -40,18 +38,16 @@ type Queue[Tx any] struct {
...
 
-	logger log.Logger
-	done   chan struct{}
+	done chan struct{}
 }
 
 var ErrQueueFull = errors.New("queue full")
 
 // New creates a new queue.
-func New[Tx any](insert func(txs []*Tx) []error, maxSize int, logger log.Logger) *Queue[Tx] {
```
mattac21, line 50: logger was unused

mempool/mocks/VMKeeperI.go
mattac21, line 1: the name of this mock was wrong (missing the I on the interface)

mempool/txpool/legacypool/legacypool.go
```diff
@@ -642,19 +642,84 @@ func (pool *LegacyPool) ContentFrom(addr common.Address) ([]*types.Transaction,
...
 }
 
-// Pending retrieves all currently processable transactions, grouped by origin
+// Rechecked retrieves all currently rechecked transactions, grouped by origin
 // account and sorted by nonce.
 //
 // The transactions can also be pre-filtered by the dynamic fee components to
 // reduce allocations and load on downstream subsystems.
-func (pool *LegacyPool) Pending(ctx context.Context, height *big.Int, filter txpool.PendingFilter) map[common.Address][]*txpool.LazyTransaction {
+func (pool *LegacyPool) Rechecked(ctx context.Context, height *big.Int, filter txpool.PendingFilter) map[common.Address][]*txpool.LazyTransaction {
```
mattac21, line 651: the `ExperimentalEVMMempool` calls `Pending` and the `KrakataoMempool` calls `Rechecked`.

mempool/txpool/subpool.go
```diff
@@ -34,7 +34,6 @@ import (
 // enough for the miner and other APIs to handle large batches of transactions;
 // and supports pulling up the entire transaction when really needed.
 type LazyTransaction struct {
-	Pool LazyResolver       // Transaction resolver to pull the real transaction up
```
mattac21, line 37: unused and made the `filterAndWrapTxs` helper have an extra param, so I removed it

mempool/krakatoa_mempool.go
mattac21, line 1: moved all krakatoa specific features into here. there is a bit of code duplication between this and the experimental evm mempool still. if we think it is worth breaking the common pieces out into more helper functions or a shared embedded struct, I can address in another pr

mempool/krakatoa_mempool_test.go
mattac21, line 1: all of the mempool tests were specific to krakatoa, so moved them to `krakatoa_mempool_test.go`

mempool/tx_tracker.go
```diff
@@ -148,12 +148,6 @@ func (txt *txTracker) RemoveTxFromPool(hash common.Hash, pool legacypool.PoolTyp
 	return nil
 }
 
-// RemoveTx removes a tx from the tx tracker and does not record any metrics as
-// it exits the tracker.
-func (txt *txTracker) RemoveTx(hash common.Hash) {
```
mattac21, line 153: unused

rpc/backend/call_tx.go
```diff
@@ -162,23 +162,19 @@ func (b *Backend) SendRawTransaction(ctx context.Context, data hexutil.Bytes) (r
...
-		// track the tx for tx inclusion timing metrics of local txs
-		if err := b.Mempool.TrackTx(txHash); err != nil {
-			b.Logger.Error("error tracking inserted inserted into mempool", "hash", txHash, "err", err)
-		}
 
 		// we are directly calling into the mempool rather than the ABCI
 		// handler for InsertTx, since the ABCI handler obfuscates the error's
 		// returned via codes, and we would like to have the full error to
 		// return to clients.
-		err := b.Mempool.Insert(b.Application.GetContextForCheckTx(txBytes), cosmosTx)
```
mattac21, line 174: internally the mempool will use the context on the rechecker to run any validation that is happening during insert tx, so there is no need to specifically get the check context

evmd/tests/integration/create_app.go
```diff
@@ -23,9 +22,8 @@ import (
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 )
 
-// CreateEvmd creates an evm app for regular integration tests (non-mempool)
-// This version uses a noop mempool to avoid state issues during transaction processing
-func CreateEvmd(chainID string, evmChainID uint64, customBaseAppOptions ...func(*baseapp.BaseApp)) evm.EvmApp {
+// CreateEvmd creates an evm app for integration tests
+func CreateEvmd(chainID string, evmChainID uint64, exclusiveMempool bool, customBaseAppOptions ...func(*baseapp.BaseApp)) evm.EvmApp {
```
aljo242, line 26: feels like this should be refactored into a struct  arg

```go

type CreateEVMDArgs struct {
 ... 
... 
fields
}

func CreateEvmd(args CreateEVMDArgs, ...func(*baseapp.BaseApp)) evm.EvmApp {

```

aljo242: probably just for a follow up ticket

evmd/app.go
```diff
@@ -185,7 +185,7 @@ type EVMD struct {
 	FeeMarketKeeper feemarketkeeper.Keeper
 	EVMKeeper       *evmkeeper.Keeper
 	Erc20Keeper     erc20keeper.Keeper
-	EVMMempool      *evmmempool.ExperimentalEVMMempool
+	EVMMempool      sdkmempool.ExtMempool
```
aljo242, line 188: is this just the interface instead of concrete type?

mempool/interface.go
```diff
@@ -13,6 +13,17 @@ import (
...
 
+// NotifiedMempool is the set of methods that a mempool must implement in order
+// to be notified of new blocks by this Keeper via the EndBlocker.
+type NotifiedMempool interface {
+	// HasEventBus returns true if the mempool has an event bus configured to
+	// get cometbft events
+	HasEventBus() bool
+
+	// GetBlockchain returns the mempools blockchain representation.
+	GetBlockchain() *Blockchain
```
aljo242, line 24: What is `BlockChain`?

Should this be an interface too or is that too much indirection?

evmd/app.go
mattac21: yeah its what the sdk wants from an external mempool, both the krakatoa mempool and experimental evm mempool implement it

evmd/tests/integration/create_app.go
mattac21: agree

mattac21: https://linear.app/cosmoslabs/issue/STACK-2481/extract-create-evmd-app-args-in-integration-test-to-struct

mempool/interface.go
mattac21: this blockchain manages the mempools view of the chain, it handles getting the latest context, latest header, etc.

mattac21: all mempools use the same impl, so imo its fine to keep it as the struct for now until we actually need to extract it out as an interface if we have multiple ways (structs) to represent the chain for different mempools

evmd/mempool.go
```diff
@@ -34,65 +30,97 @@ func (app *EVMD) configureEVMMempool(appOpts servertypes.AppOptions, logger log.
 		return nil
 	}
 
-	mempoolConfig, err := app.createMempoolConfig(appOpts, logger)
-	if err != nil {
-		return fmt.Errorf("failed to get mempool config: %w", err)
+	if server.GetShouldOperateExclusively(appOpts, logger) {
```
vladjdk, line 33: Wondering if we could completely move these out now that we have configs for this and a flag to turn this on and off. Is there any real reason for keeping this setup logic directly in people's applications? This was already here, so maybe better to track in a separate issue if you agree with this.

evmd/mempool.go
```diff
@@ -34,65 +30,97 @@ func (app *EVMD) configureEVMMempool(appOpts servertypes.AppOptions, logger log.
...
+		logger.Info("app-side mempool is operating exclusively, setting up Krakatoa mempool")
```
technicallyty, line 34: so operate exclusively just sets up krakatoa? seems like we should just have it be like`--krakatoa=enabled` or something. that way its more like `--<feature>=<on/off>`

evmd/mempool.go
```diff
@@ -34,65 +30,97 @@ func (app *EVMD) configureEVMMempool(appOpts servertypes.AppOptions, logger log.
...
+			),
+		)
+		app.SetPrepareProposal(abciProposalHandler.PrepareProposalHandler())
+
+		app.EVMMempool = krakatoaMempool
+		app.SetMempool(krakatoaMempool)
+	} else {
+		logger.Info("app-side mempool is not operating exclusively, setting up default EVM mempool")
+
+		evmMempool := evmmempool.NewExperimentalEVMMempool(
```
technicallyty, line 70: side note: should we drop the `Experimental` now? i mean, krakatoa is also technically experimental so seems a tad confusing

mattac21: I was wondering this as well, maybe @vladjdk has thoughts

vladjdk: I feel like we should work towards something that denotes keeps them both as experimental. Maybe having godocs paired with comments in the app.toml could be a good start

evmd/mempool.go
mattac21: I'm not 100% sure what you mean by this, where would it move to? fetching the flags and populating the config feels like a responsibility of evmd not the evm library

evmd/mempool.go
mattac21: yeah, I think we need to work on the naming. we had just already called this operate exclusively before so I stuck with it (i guess describing the specific behavior you're enabling rather than specifically the feature name). but I would be down to change it

evmd/mempool.go
mattac21: https://linear.app/cosmoslabs/issue/STACK-2489/consider-dropping-experimental-from-experimentalevmmempool

evmd/mempool.go
mattac21: https://linear.app/cosmoslabs/issue/STACK-2487/rename-operate-exclusively-to-be-krakatoa

evmd/mempool.go
mattac21: created a ticket for this

PR #1062, opened 2026-03-10
feat: async recheck support 

technicallyty:
# Description

we recently [updated comet](https://github.com/cometbft/cometbft/pull/5666) to no longer lock on recheck, pushing concurrency responsibility to the application.

changes:

- route CheckTx through the app-side mempool insert worker instead of BaseApp.runTx
- fix Cosmos rechecked tx tracking to replace entries by signer/nonce identity rather than pointer identity. the previous code was not allowing cosmos fee replacement txs

Closes: STACK-2455

---

## Author Checklist

I have...

### DIFF
--- a/mempool/check_tx.go
+++ b/mempool/check_tx.go
@@ -1,49 +1,42 @@
 package mempool
 
 import (
+	"context"
 	"errors"
+	"time"
 
 	abci "github.com/cometbft/cometbft/abci/types"
-
-	"github.com/cosmos/evm/mempool/txpool"
 
 	"github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
 // NewCheckTxHandler creates a CheckTx handler that integrates with the EVM mempool for transaction validation.
-// It wraps the standard transaction execution flow to handle EVM-specific nonce gap errors by routing
-// transactions with higher tx sequence numbers to the mempool for potential future execution.
-// Returns a handler function that processes ABCI CheckTx requests and manages EVM transaction sequencing.
-func NewCheckTxHandler(mempool *ExperimentalEVMMempool) types.CheckTxHandler {
-	return func(runTx types.RunTx, request *abci.RequestCheckTx) (*abci.ResponseCheckTx, error) {
-		gInfo, result, anteEvents, err := runTx(request.Tx, nil)
-		if err != nil {
-			// detect if there is a nonce gap error (only returned for EVM transactions)
-			if errors.Is(err, ErrNonceGap) || errors.Is(err, ErrNonceLow) {
-				// send it to the mempool for further triage
-				err := mempool.InsertInvalidNonce(request.Tx)
-				if err != nil {
-					return sdkerrors.ResponseCheckTxWithEvents(err, gInfo.GasWanted, gInfo.GasUsed, anteEvents, false), nil
-				}
-			}
-			// If its already known, this can mean the the tx was promoted from nonce gap to valid
-			// and by allowing ErrAlreadyKnown to be silent, we allow re-gossiping of such txs
-			// this also covers the case of re-submission of the same tx enforcing overpricing for replacement
-			if errors.Is(err, txpool.ErrAlreadyKnown) {
-				return sdkerrors.ResponseCheckTxWithEvents(nil, gInfo.GasWanted, gInfo.GasUsed, anteEvents, false), nil
-			}
-
-			// anything else, return regular error
-			return sdkerrors.ResponseCheckTxWithEvents(err, gInfo.GasWanted, gInfo.GasUsed, anteEvents, false), nil
+// It routes new CheckTx requests through the same async insert worker path used by
+// the app-side mempool and waits for the insert result.
+func NewCheckTxHandler(mempool *ExperimentalEVMMempool, debug bool, timeout time.Duration) types.CheckTxHandler {
+	if timeout <= 0 {
+		panic("invalid timeout CheckTxHandler timeout value")
+	}
+	return func(_ types.RunTx, request *abci.RequestCheckTx) (*abci.ResponseCheckTx, error) {
+		// TODO: do we even do recheck anymore?
+		if request.Type == abci.CheckTxType_Recheck {
+			return &abci.ResponseCheckTx{Code: abci.CodeTypeOK}, nil
 		}
 
-		return &abci.ResponseCheckTx{
-			GasWanted: int64(gInfo.GasWanted), // #nosec G115 -- this is copied from the Cosmos SDK
-			GasUsed:   int64(gInfo.GasUsed),   // #nosec G115 -- this is copied from the Cosmos SDK
-			Log:       result.Log,
-			Data:      result.Data,
-			Events:    types.MarkEventsToIndex(result.Events, nil),
-		}, nil
+		tx, err := mempool.txConfig.TxDecoder()(request.Tx)
+		if err != nil {
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, debug), nil
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		if err := mempool.Insert(ctx, tx); err != nil {
+			if errors.Is(err, context.DeadlineExceeded) {
+				err = ctx.Err()
+			}
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, debug), nil
+		}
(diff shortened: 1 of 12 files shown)


mempool/check_tx.go
```diff
@@ -10,31 +10,31 @@ import (
...
-	return func(runTx types.RunTx, request *abci.RequestCheckTx) (*abci.ResponseCheckTx, error) {
-		gInfo, result, anteEvents, err := runTx(request.Tx, nil)
+// It routes new CheckTx requests through the same async insert worker path used by
+// the app-side mempool and waits for the insert result.
+func NewCheckTxHandler(mempool *ExperimentalEVMMempool, debug bool) types.CheckTxHandler {
+	return func(_ types.RunTx, request *abci.RequestCheckTx) (*abci.ResponseCheckTx, error) {
+		// TODO: do we even do recheck anymore?
+		if request.Type == abci.CheckTxType_Recheck {
+			return &abci.ResponseCheckTx{Code: abci.CodeTypeOK}, nil
+		}
```
technicallyty: we can expect this to never be called with this mempool, yes? cc @mattac21

mempool/check_tx.go
```diff
@@ -10,31 +10,31 @@ import (
...
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, debug), nil
+		}
+
+		if errC != nil {
+			if err := <-errC; err != nil {
+				return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, debug), nil
 			}
-			// anything else, return regular error
-			return sdkerrors.ResponseCheckTxWithEvents(err, gInfo.GasWanted, gInfo.GasUsed, anteEvents, false), nil
 		}
```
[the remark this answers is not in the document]
technicallyty: this is a bad suggestion because you're operating on a new background context in every call

mempool/tx_store.go
```diff
@@ -22,22 +25,55 @@ type CosmosTxStore struct {
...
 
+	if key, ok := cosmosTxKey(tx); ok {
+		if idx, exists := s.keys[key]; exists {
+			delete(s.index, s.txs[idx])
+			s.txs[idx] = tx
+			s.index[tx] = idx
+			return
+		}
+		s.keys[key] = len(s.txs)
+	}
```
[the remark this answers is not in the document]
technicallyty: I don't think you understand this code. The store is replaced wholesale at every new block. So the growth only happens within a block, and a new store is created, old one dropped from memory.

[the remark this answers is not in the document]
technicallyty: @greptile I don't think you understand this code. The store is replaced wholesale at every new block. So the growth only happens within a block, and a new store is created, old one dropped from memory.

mempool/check_tx.go
```diff
@@ -10,31 +10,31 @@ import (
...
 
-		return &abci.ResponseCheckTx{
-			GasWanted: int64(gInfo.GasWanted), // #nosec G115 -- this is copied from the Cosmos SDK
-			GasUsed:   int64(gInfo.GasUsed),   // #nosec G115 -- this is copied from the Cosmos SDK
-			Log:       result.Log,
-			Data:      result.Data,
-			Events:    types.MarkEventsToIndex(result.Events, nil),
-		}, nil
+		return &abci.ResponseCheckTx{Code: abci.CodeTypeOK}, nil
```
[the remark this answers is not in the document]
technicallyty: @greptile gas values aren't populated in regular checktx anyways. this is not a regression

mempool/check_tx.go
mattac21: yes

mempool/check_tx.go
```diff
@@ -10,31 +11,40 @@ import (
...
-			// detect if there is a nonce gap error (only returned for EVM transactions)
-			if errors.Is(err, ErrNonceGap) || errors.Is(err, ErrNonceLow) {
-				// send it to the mempool for further triage
-				err := mempool.InsertInvalidNonce(request.Tx)
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, debug), nil
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		errC, err := mempool.insert(ctx, tx)
```
mattac21: the `mempool.Insert` function already waits on the `errC` or `ctx` to be done like you are doing here. could we use that instead of using the private `mempool.insert`?

mempool/check_tx.go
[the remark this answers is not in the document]
mattac21: is this true that the values were always zero? I thought you do get gas wanted and used values back from checktx?

mempool/tx_store.go
```diff
@@ -14,6 +16,7 @@ type CosmosTxStore struct {
 
 	// index maps a tx to its position in the txs slice for fast removal
 	index map[sdk.Tx]int
+	keys  map[string]int
```
mattac21, line 19: keys is now serving the same purpose as index right? we can remove index now I think

mempool/tx_store.go
```diff
@@ -22,22 +25,55 @@ type CosmosTxStore struct {
...
 
 // AddTx adds a single tx to the store. Duplicate txs (by pointer identity)
-// are ignored.
+// are ignored. Transactions with the same signer/nonce tuple overwrite the
+// existing entry to mirror the SDK PriorityNonceMempool replacement model.
 func (s *CosmosTxStore) AddTx(tx sdk.Tx) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	if key, ok := cosmosTxKey(tx); ok {
```
mattac21, line 39: hm thinking through this, im not actually sure this replacement is safe to do without rechecking all txs for this account, when we replace a tx we may want to just remove all txs > replaced nonce from the tx store for this account and then let them just be included in the next block.

I think this isn't safe because we would need to recheck all txs after the replaced one onto of the state of this new tx, which we are not doing. for example if we recheck txs 4 5 and 6 of an an account and include them in the tx store, then someone can replace tx 4 with a completely different tx that may have invalidated 5 and 6, but we are not rechecking those against the new tx 4's context, which may then cause the proposal to be invalid.

I think this is an issue for evm txs in the legacypool too that we need to address.

mempool/check_tx.go
```diff
@@ -10,31 +11,40 @@ import (
...
+		if err != nil {
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, debug), nil
+		}
+
+		if errC != nil {
+			select {
+			case err := <-errC:
 				if err != nil {
-					return sdkerrors.ResponseCheckTxWithEvents(err, gInfo.GasWanted, gInfo.GasUsed, anteEvents, false), nil
```
mattac21, line 27: were the `anteEvents` also always nil here? we are missing out on events now as well with this? should we modify the response of `mempool.Insert` to return some of this info?

mattac21 (commented):
could you also rebase this off of main? the tests are pretty broken on feat/krakatoa but if you go to main you should be able to get all of the system unit and integration tests passing

mempool/check_tx.go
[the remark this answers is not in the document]
technicallyty: @greptile why would those values be populated if checkTx is just a mempool admission api and not a tx execution check?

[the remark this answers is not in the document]
technicallyty: @greptile cometbft reduces the fields in the response anyway: https://github.com/cometbft/cometbft/blob/1bb8b386fc366bc4655dede0535e16d1ad669c7d/rpc/core/mempool.go#L58-L64. gas has never been returned here

mempool/check_tx.go
technicallyty: i checked on v0.53.x and main, calling broadcast tx sync never returned anything other than code, tx hash, and a log if it failed.

you can see comet stripping down the response here: https://github.com/cometbft/cometbft/blob/1bb8b386fc366bc4655dede0535e16d1ad669c7d/rpc/core/mempool.go#L58-L64

mempool/check_tx.go
technicallyty: updated [change to public method](https://github.com/cosmos/evm/pull/1062/commits/f7e5c4710faa0aeeb0569d7f844f5d759060b13e)

mempool/check_tx.go
mattac21: got it, tested on gaia as well earlier and this was the case too

mempool/tx_store.go
technicallyty: working on a separate PR to address this issue in both evm and cosmos tx stores

PR #1067, opened 2026-03-11
fix(rpc): guard nil height deref in CometBlockResultByNumber

### LINKED ISSUE #1066: nil pointer deref in CometBlockResultByNumber error path
Was tracing an intermittent RPC crash on a pruned archive node. The function sets height = nil when the incoming pointer holds 0, to let CometBFT resolve the latest block. If the subsequent BlockResults call then fails (timeout, pruned state, network error), the error format string at line 71 tries to dereference height, which is now nil. The panic only surfaces under error conditions, so it never shows up in happy-path tests.

Fix is one line: capture the int64 value before nullifying the pointer, then use that value in the error message instead of dereferencing height.

Aboudjem:
Fixes #1066.

Was tracing an intermittent crash on a pruned node that only showed up under load. The stack trace pointed at `comet.go:71` in the error path of `CometBlockResultByNumber`. When the caller passes height=0, the pointer gets set to nil before the `BlockResults` RPC call. If that call fails (timeout, pruned state, etc), the error formatting dereferences the nil pointer and panics.

The fix captures the height value before the nil reassignment and uses it in the error message. The variable already existed for the tracing span attribute so this just reuses it in the format string.

Added a table-driven test covering the zero-height error path to confirm it no longer panics.

### DIFF
--- a/rpc/backend/comet.go
+++ b/rpc/backend/comet.go
@@ -68,7 +68,7 @@ func (b *Backend) CometBlockResultByNumber(ctx context.Context, height *int64) (
 	}
 	res, err := b.RPCClient.BlockResults(ctx, height)
 	if err != nil {
-		return nil, fmt.Errorf("failed to fetch block result from CometBFT %d: %w", *height, err)
+		return nil, fmt.Errorf("failed to fetch block result from CometBFT %d: %w", heightAttr, err)
 	}
 
 	return res, nil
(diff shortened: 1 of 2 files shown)


swift1337 (approved):

vladjdk (approved):

vladjdk:
@Aboudjem make sure you have signed commits only

Aboudjem:
@vladjdk rebased + signed: bff0971 and 447d761 both have valid SSH signatures now (force-pushed onto latest main). 

@swift1337 sorry for the dismiss-by-force-push — content is identical to what you approved (`git diff` shows just the original 2 files: comet.go nil-deref guard + comet_test.go test). Could you re-review when you have a moment? Thanks!

vladjdk (approved):

vladjdk:
Thank you @Aboudjem!

PR #1074, opened 2026-03-11
fix: handle replacement txs in TxStore

technicallyty:
# Description

updates the TxStore to handle tx replacements. txs that get replaced as a result of an insert should invalidate all same-sender txs with a higher nonce value, as they were dependent on the original tx's execution.

Closes: STACK-2479

---

## Author Checklist

I have...

### DIFF
--- a/mempool/krakatoa_mempool_test.go
+++ b/mempool/krakatoa_mempool_test.go
@@ -353,7 +353,11 @@ func TestKrakatoaMempool_ReapNewBlock(t *testing.T) {
 	// wait for another reset to make sure the pool processes the above
 	// txns into pending
 	require.NoError(t, mp.GetTxPool().Sync())
-	require.Equal(t, 3, mp.CountTx())
+	legacyPool := mp.GetTxPool().Subpools[0].(*legacypool.LegacyPool)
+	require.Eventually(t, func() bool {
+		pending, queued := legacyPool.ContentFrom(accounts[0].address)
+		return len(pending) == 3 && len(queued) == 0 && mp.CountTx() == 3
+	}, time.Second, 10*time.Millisecond)
 
 	// simulate comet calling removeTx, a new height being published, and
 	// our accounts nonce increments to 1, so tx 0 will be invalidated
@@ -375,8 +379,10 @@ func TestKrakatoaMempool_ReapNewBlock(t *testing.T) {
 	// sync the pool to make sure the above happens, tx0 should be dropped
 	// from the pool and the reap list
 	require.NoError(t, mp.GetTxPool().Sync())
-
-	legacyPool := mp.GetTxPool().Subpools[0].(*legacypool.LegacyPool)
+	require.Eventually(t, func() bool {
+		pending, queued := legacyPool.ContentFrom(accounts[0].address)
+		return len(pending) == 2 && len(queued) == 0
+	}, time.Second, 10*time.Millisecond)
 	pending, queued := legacyPool.ContentFrom(accounts[0].address)
 	require.Len(t, pending, 2)
 	require.Len(t, queued, 0)


mempool/txpool/legacypool/tx_store.go
```diff
@@ -122,24 +123,31 @@ func (t *TxStore) AddTx(addr common.Address, tx *types.Transaction) {
 
 // RemoveTx removes a tx for an address from the current set.
 func (t *TxStore) RemoveTx(addr common.Address, tx *types.Transaction) {
+	t.RemoveTxsFromNonce(addr, tx.Nonce())
+}
```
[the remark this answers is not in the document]
technicallyty: @greptile This is intended behavior because if we remove tx with nonce 6, then any tx with a higher nonce can no longer be considered valid for block building. those txs cannot be executed

mempool/txpool/legacypool/tx_store_test.go
```diff
@@ -101,34 +100,7 @@ func TestTxStoreRemoveTx(t *testing.T) {
 	store.RemoveTx(addr1, tx1)
 
 	result := store.Txs(txpool.PendingFilter{})
-	require.Len(t, result[addr1], 1)
-	require.Equal(t, uint64(1), result[addr1][0].Tx.Nonce())
-}
-
-func TestTxStoreConcurrentRemove(t *testing.T) {
```
technicallyty, line 108: not sure if this test makes sense to have anymore

mempool/txpool/legacypool/tx_store_test.go
```diff
@@ -101,34 +100,7 @@ func TestTxStoreRemoveTx(t *testing.T) {
...
-		go func(nonce uint64) {
-			defer wg.Done()
-			store.RemoveTx(addr1, createTestTx(nonce, big.NewInt(1e9), big.NewInt(2e9)))
-		}(nonce)
-	}
-	wg.Wait()
-
-	result := store.Txs(txpool.PendingFilter{})
-	require.Len(t, result[addr1], 500)
+	require.Len(t, result[addr1], 0)
```
technicallyty, line 128: updating this bc i think the behavior we want now is if a tx is removed, all txs with a greater nonce should now be considered invalid and removed as well. they cannot be included in a block proposal.

mempool/txpool/legacypool/tx_store.go
```diff
@@ -136,25 +137,31 @@ func (t *TxStore) AddTx(addr common.Address, tx *types.Transaction) {
...
 		}
+		next = append(next, existing)
+	}
+
+	if len(next) == 0 {
+		delete(t.txs, addr)
+		return
 	}
+	t.txs[addr] = next
 }
```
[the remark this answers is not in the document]
technicallyty: fixed in [update total in removeTxsFromNonce](https://github.com/cosmos/evm/pull/1074/commits/64c1ccf56b995ca9f44e4179fd1b6e4786364268)

mempool/txpool/legacypool/tx_store.go
```diff
@@ -136,25 +137,34 @@ func (t *TxStore) AddTx(addr common.Address, tx *types.Transaction) {
...
+		next = append(next, existing)
+	}
+	t.total -= uint64(numRemoved)
+
+	if len(next) == 0 {
+		delete(t.txs, addr)
+		return
 	}
+	t.txs[addr] = next
 }
```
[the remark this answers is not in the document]
technicallyty: fixed it, thanks clanker. [memory retain](https://github.com/cosmos/evm/pull/1074/commits/facec3248ab95b8312442bf5d0887a6d2ce91d9b)

mempool/tx_store.go
```diff
@@ -14,32 +17,152 @@ import (
...
+		fmt.Fprintf(&b, "%s/%d", sig.account, sig.seq)
+	}
+
+	return b.String(), true
+}
+
+// cosmosTxNonceMap extracts the signers from the transaction
+// and returns a signer -> nonce map.
+func cosmosTxNonceMap(tx sdk.Tx) (map[string]uint64, bool) {
+	signerSeqs, err := extractSignerSequences(tx)
```
mattac21, line 179: is there a limit on the amount of signers a tx can have that is validated somewhere before this? if not then this is a user supplied value we are allocating memory on which isn't good (i.e. someone sends us a tx with 5 mil tiny invalid signers) (I can't comment on the line since it isn't changed, but `extractSignerSequences` line 487)

mattac21: mm i guess we are running ante handlers just before this, so that would probably be validated away in there

mempool/tx_store.go
```diff
@@ -14,32 +17,152 @@ import (
...
+	if !ok {
+		return "", false
+	}
+
+	var b strings.Builder
+	for i, sig := range sortedSignerNonces(nonceMap) {
+		if i > 0 {
+			b.WriteByte('|')
+		}
+		fmt.Fprintf(&b, "%s/%d", sig.account, sig.seq)
```
mattac21: i guess same question on the number of max signers allowed here. could someone force us to allocate a massive string here? if so maybe we have to hash a key made of account + seq together with other signers keys in order to have a fixed sized key? but that seems way slower

mempool/tx_store.go
```diff
@@ -14,32 +17,152 @@ import (
...
+	thresholds, ok := cosmosTxNonceMap(tx)
+	if !ok {
+		return 0
+	}
+
+	// rebuild the txs list, skipping txs that are invalidated.
+	removed := 0
+	nextTxs := make([]sdk.Tx, 0, len(s.txs))
+	for _, existing := range s.txs {
+		if invalidatesCosmosTx(existing, thresholds) {
```
mattac21: we compute the nonce map in `invalidatedCosmosTx` but we have already computed it in `InvalidateFrom`, could we pass it in?

mempool/tx_store.go
```diff
@@ -14,32 +17,152 @@ import (
...
+
+	// nonce thresholds for each signer.
+	thresholds, ok := cosmosTxNonceMap(tx)
+	if !ok {
+		return 0
+	}
+
+	// rebuild the txs list, skipping txs that are invalidated.
+	removed := 0
+	nextTxs := make([]sdk.Tx, 0, len(s.txs))
```
mattac21: damn this feels super rough that we have to rebuild the entire list again. do you think it would be better if we changed the data structure of the tx store here to be more like the evm one which is a map? so we only have to rebuild per account

mempool/tx_store.go
```diff
@@ -14,32 +17,152 @@ import (
...
+		}
+	}
+	return false
+}
+
+func (s *CosmosTxStore) reindex(txs []sdk.Tx) {
+	s.txs = txs
+	s.keys = make(map[string]int, len(txs))
+	for i, tx := range txs {
+		if key, ok := cosmosTxKey(tx); ok {
```
mattac21: computing this key again seems really rough, would love if we could maybe cache it on the tx itself, kind of like how signatures are cached on eth transaction structs, but I agree probably fine to just leave this for now

mempool/tx_store.go
technicallyty: are you referring to the nonceMap on line 68? that one is the nonceMap for the tx we are checking, the one here is for each tx we check _against_

mattac21: Ahh I see

PR #1078, opened 2026-03-15
fix: add nil check for OnTxStart before calling tracer hook

### LINKED ISSUE #209: Tracer.OnTxStart may cause panic if not implemented by custom tracers
Currently, during state transition in ApplyMessageWithConfig, if a tracer is provided, the following line is invoked: https://github.com/cosmos/evm/blob/73bc2435c3886d7ae181b992a7e6dd2e1c04093b/x/vm/keeper/state_transition.go#L344

This assumes that the `OnTxStart` method is implemented on all tracers. However, some custom or default tracers may not implement this method, which can result in a panic.

In a previous test using the default tracer, I recall encountering a panic due to the missing implementation of `OnTxStart`.

We would need to reproduce the exact condition to share a concrete panic trace, but I'm filing this issue for now to keep track and allow follow-up.

**Suggested action:**
- Consider adding a type assertion or interface check before invoking OnTxStart
- Alternatively, provide a no-op default implementation to avoid unexpected crashes

Bue-von-hon:
# Description

Calling vmCfg.Tracer.OnTxStart without a nil check causes a panic when a custom tracer does not implement the OnTxStart field. Added a nil guard consistent with the existing OnTxEnd check.

Closes #209

---

## Author Checklist


vladjdk (approved):

Bue-von-hon:
Please review.
I'm new to this and don't know who to mention, so I'm just leaving a comment here.

PR #1087, opened 2026-03-25
fix: logindex json rpc

### LINKED ISSUE #1086: bug: evm log index doesn't match go-ethereum JSON-RPC response
## Description

After the Block STM change (#589), log.Index returned by JSON-RPC endpoints is scoped per-transaction (starting at 0 for each tx) instead of being a block-global index as required by the https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getfilterchanges.

Per the spec, logIndex is defined as:
> integer of the log index position in the block

For example, if tx0 emits 3 logs and tx1 emits 2 logs, tx1's logs should have indices 3 and 4. Currently they are returned as 0 and 1.

## Root Cause

The Block STM commit removed the LogIndex field from TxConfig in x/vm/statedb/config.go to eliminate cross-transaction state dependencies that block parallel execution. This was the correct change for the execution layer.

Before:
```go
type TxConfig struct {
    TxHash   common.Hash
    TxIndex  uint
    LogIndex uint        // removed by Block STM
}
```

In StateDB.AddLog (x/vm/statedb/statedb.go), the index assignment changed from:
```go
 log.Index = s.txConfig.LogIndex + uint(len(s.logs))  // block-global
```
to:
```go
 log.Index = uint(len(s.logs))  // per-transaction, starting at 0
```

However, no compensating logic was added at the RPC layer to reassign block-global indices when serving responses to client.

## Affected RPC endpoints

 - eth_getTransactionReceipt - via ReceiptsFromCometBlock
 - eth_getBlockReceipts / eth_getBlockByNumber / eth_getBlockByHash - via ReceiptsFromCometBlock
 - eth_getLogs / eth_getFilterLogs - via GetLogsFromBlockResults
 - eth_getTransactionLogs - via direct DecodeMsgLogs call in GetTransactionLogs
 - eth_subscribe (logs) / eth_getFilterChanges - via log stream in rpc/stream/rpc.go

## Impact

Any tooling that relies on logIndex being unique within a block will break. This includes event indexers, block explorers, and any client code that uses log index as a unique identifier within a block.

## Suggested Fix

Reassign block-global log indices at the RPC layer when building responses. This keeps the execution layer parallel-friendly (no cross-tx dependency in StateDB) while maintaining spec compliance in API responses.

canu0205:
# Description

Closes: #1086

---

## Author Checklist

I have...

### DIFF
--- a/rpc/backend/tx_info.go
+++ b/rpc/backend/tx_info.go
@@ -252,6 +252,16 @@ func (b *Backend) GetTransactionLogs(ctx context.Context, hash common.Hash) (res
 	if err != nil {
 		return nil, err
 	}
+	// compute cumulative log index offset from prior txs in the block
+	cumulatedLogIndex := uint(0)
+	for ti := int64(0); ti < int64(res.TxIndex); ti++ {
+		priorLogs, err := evmtypes.DecodeTxLogs(resBlockResult.TxsResults[ti].Data, height)
+		if err != nil {
+			continue
+		}
+		cumulatedLogIndex += uint(len(priorLogs))
+	}
+
 	// parse tx logs from events
 	index := int(res.MsgIndex) // #nosec G701
 	logs, err := evmtypes.DecodeMsgLogs(
@@ -261,6 +271,12 @@ func (b *Backend) GetTransactionLogs(ctx context.Context, hash common.Hash) (res
 	)
 	if err != nil {
 		b.Logger.Debug("failed to parse tx logs", "error", err.Error())
+	}
+
+	// reassign log indices to be block-global per Ethereum spec
+	for _, log := range logs {
+		log.Index = cumulatedLogIndex
+		cumulatedLogIndex++
 	}
 
 	return logs, nil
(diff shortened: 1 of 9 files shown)


rpc/backend/tx_info.go
```diff
@@ -210,6 +209,10 @@ func (b *Backend) GetTransactionReceipt(ctx context.Context, hash common.Hash) (
 		return nil, fmt.Errorf("failed to get receipts from comet block")
 	}
 
+	// fix log indices to be block-global per Ethereum spec
+	offset := countPriorBlockLogs(blockRes.TxsResults, res.TxIndex, blockRes.Height)
```
cloudgray: [trivial]
I think, we better name it "startIdx" instead of "offset".
It is because, `receipts` has logs whose idx is 0,1,2... -> "offset"
And countPriorBlockLogs counts accumulated log index by `txIdex - 1`. -> startIndex

cloudgray (commented):
Looks good to me.
I just left trivial comment.

rpc/backend/tx_info.go
canu0205: yeah applied [b57e9d3](https://github.com/cosmos/evm/pull/1087/commits/b57e9d333fbe8878e6a22f0045fcf975f08ce5d3)

canu0205:
#812 already handles this issue

PR #1113, opened 2026-04-17
feat(mempool): Remove EVM txs without state lookup for nonce

mattac21:
# Description

Modifies evm tx removals to not require a state lookup. Previously, the behavior had evm txs that were successfully included in a block not be removed during `FinalizeBlock` and the removal was deferred until the legacypool was reset. The legacypool would then check the latest nonce on chain for the sender (+1 of the tx that was sent) and then remove the tx if the nonce was too low. This was severely hurting performance when the mempool is large, taking >1.5s simply to remove txs that we already know must be removed since they were included in the previous block.

The PR now modifies the behavior so that we create LRU cache of sender -> nonce when we see txs included on chain. Now during the reorg loop in order to remove txs included in the previous block, we look up the latest nonce via this cache, instead of via state which is significantly faster.

Closes: STACK-2492

---

## Author Checklist

I have...


mempool/txpool/legacypool/legacypool.go
```diff
@@ -417,6 +423,38 @@ func New(config Config, logger cosmoslog.Logger, chain BlockChain, opts ...Optio
...
+	sender, err := types.Sender(pool.signer, tx)
+	if err != nil {
+		return fmt.Errorf("getting tx signer: %w", err)
+	}
+
+	existing, ok := pool.latestIncludedNonce.Get(sender)
+	if !ok || tx.Nonce() > existing {
+		pool.latestIncludedNonce.Add(sender, tx.Nonce())
+	}
+	return nil
```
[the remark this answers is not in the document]
mattac21: this seems fine. the tx with nonce 5 is invalid anyways since nonce 5 was just included on chain so we actually want to silently forward past it and drop it. we also dont want to remove nonces within the lru inside of remove olds since a sender may have txs in both queued and pending. if we process removals in queued and then remove the entry for this sender, we will not drop any old txs that they have in pending.

mattac21:
@greptile re review this pls

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1639,23 +1676,15 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address, cancelled
...
-		// Drop all transactions that are deemed too old (low nonce)
-		forwards := list.Forward(pool.currentState.GetNonce(addr))
-		for _, tx := range forwards {
-			pool.all.Remove(tx.Hash())
-			pool.markTxRemoved(addr, tx, Queue)
-			queueRemovalMetric(RemovalReasonOld).Mark(1)
-		}
-		log.Trace("Removed old queued transactions", "count", len(forwards))
-		// Drop all transactions that are too costly (low balance or out of gas)
-		costDrops, _ := list.CostFilter(pool.currentState.GetBalance(addr), gasLimit)
```
mattac21, line 1651: we also dont have to do the cost filter here either. this will be caught by the ante handler called during recheck.

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1639,23 +1676,15 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address, cancelled
 		if list == nil {
 			continue // Just in case someone calls with a non existing account
 		}
-		// Drop all transactions that are deemed too old (low nonce)
-		forwards := list.Forward(pool.currentState.GetNonce(addr))
```
mattac21, line 1643: now replaced by `pool.removeOlds`

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1639,23 +1676,15 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address, cancelled
...
-		for _, tx := range costDrops {
-			pool.all.Remove(tx.Hash())
-			pool.markTxRemoved(addr, tx, Queue)
-			queueRemovalMetric(RemovalReasonCostly).Mark(1)
+
+		// Drop all transactions that are below the latest included nonce for
+		// this account based on what we have seen during the latest block
+		// execution. Only do this if we are resetting.
+		var olds types.Transactions
+		if reset != nil {
```
mattac21, line 1684: we already do a nonce check on txs when we are inserting, so we know at insert time a tx's nonce is valid. so they only time a tx's nonce can become invalid is after a block is committed and we are resetting the mempool, thus we only call `pool.removeOlds` during a reset

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1924,29 +1935,16 @@ func (pool *LegacyPool) demoteUnexecutables(cancelled chan struct{}, reset *txpo
...
 			log.Trace("Demoting pending transaction", "hash", hash)
 
 			// Internal shuffle shouldn't touch the lookup set.
 			pool.enqueueTx(hash, tx, false)
 		}
-		pendingGauge.Dec(int64(len(olds) + len(drops) + len(recheckDrops) + len(invalids)))
+		pendingGauge.Dec(int64(len(recheckDrops) + len(invalids) + len(olds)))
 
-		// If there's a gap in front, alert (should never happen) and postpone all transactions
-		if list.Len() > 0 && list.txs.Get(nonce) == nil {
```
mattac21, line 1938: removed this, I dont think this is possible to have these gaps (the comment also suggests this) and we need the nonce lookup for this to be possible, which is what we are trying to avoid

vladjdk (approved):

aljo242 (approved):

swift1337 (approved):

Issue #962, opened 2026-01-29
krakatoa: no error message on cosmos txs that fail ante


technicallyty:
1. send tx with too low gas-prices
2. attempt to query the returned hash
3. tx not found

basically have no way of knowing now why my tx has failed to send.

PR #1114, opened 2026-04-20
feat(mempool): add thread-safe checkTx handler

swift1337:
**Changes** 
- [x] Add non-blocking `CheckTx` handler for `BroadcastTxSync()`
- [x] Wire configs
- [x] Move ABCI handlers to `mempool/` package
- [x] Streamline `mempoo.Insert()` and `mempoo.InsertAsync()`
- [x] Test coverage

Closes STACK-2455
Closes STACK-2612


mempool/mempool_handlers.go
```diff
@@ -0,0 +1,88 @@
...
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+
+		tx, err := txDecoder(req.GetTx())
+		if err != nil {
+			return nil, fmt.Errorf("decoding tx: %w", err)
+		}
+
+		if err := m.Insert(ctx, tx); err != nil {
+			return sdkerrors.ResponseCheckTxWithEvents(err, 0, 0, nil, false), nil
```
mattac21: I played around with `local_node.sh` a bit with this and how this function parses out the error code and codespace from the `err` argument, and I think we need to remove the error wrapping from `recheck_pool.go` when it calls `RecheckCosmos` on `Insert` (it wraps it with `ante handler failed: %w`). I haven't dug too too much into why, but this seems to break the parsing this function does and the codespace and code that get returned are incorrect. For example when I tried this with the error wrapping I got this for a tx that failed recheck:
```
code: 1
codespace: undefined
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: 'ante handler failed: gas prices too low, got: 10000000.000000000000000000atest
  required: 103308701.143592696070072636atest. Please retry using a higher gas price
  or a higher fee: insufficient fee [cosmos/evm@v0.2.0/ante/evm/fee_checker.go:99]'
timestamp: ""
tx: null
txhash: F890CD9B55155B2FDDA6F797C95D737E61495FBEC188D5C0344FD931F9B1B2D3
```
and when i removed the wrapping I got this (notice code + codespace):
```
code: 13
codespace: sdk
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: 'gas prices too low, got: 10000000.000000000000000000atest required: 176240083.250377210788428785atest.
  Please retry using a higher gas price or a higher fee: insufficient fee'
timestamp: ""
tx: null
txhash: 02E235E0F5850F245C0267275F557C7A39E8C69AC5589322AB2DFAAEA3FB0EFC
```

mempool/mempool_handlers.go
```diff
@@ -0,0 +1,88 @@
...
+		return &abci.ResponseInsertTx{Code: code}, nil
+	}
+}
+
+// NewCheckTxHandler is the handler for ABCI.CheckTx. Note: it's async and doesn't expect the caller
+// to acquire ABCI lock. Used ONLY to support BroadcastTxSync (cosmos rpc). All EVM txs
+// should be inserted via InsertTx handler or EVM RPC.
+func (m *Mempool) NewCheckTxHandler(txDecoder sdk.TxDecoder, timeout time.Duration) sdk.CheckTxHandler {
+	return func(_ sdk.RunTx, req *abci.RequestCheckTx) (*abci.ResponseCheckTx, error) {
+		if req.Type != abci.CheckTxType_New {
```
mattac21, line 71: the only other possible type here is recheck right? so here im assuming comet will panic if we return this error, making it so that if someone has recheck enabled + evm, they will panic.

swift1337: comet will indeed panic on this. recheck + evm is breaking the invariant of using krakatoa, right ?

mattac21: Yeah I think this is fine, I was just confirming my understanding

mempool/mempool_handlers.go
swift1337: Fixed!

```bash
evmd tx bank send dev0 $(evmd keys show dev1 -a --keyring-backend test --home ~/.evmd) \
    1001atest \
    --chain-id 9001 --keyring-backend test --home ~/.evmd \
    --gas 200000 --fees 1atest -y;

code: 13
codespace: sdk
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: 'ante handler failed: gas prices too low, got: 0.000005000000000000atest
  required: 378856.836311450120878629atest. Please retry using a higher gas price
  or a higher fee: insufficient fee [cosmos/evm@v0.2.0/ante/evm/fee_checker.go:99]'
timestamp: ""
tx: null
txhash: 8E03A239AF46D625112FCF590AB9C6B3CBA76E6913A8928DDDD3961B3CE8DA6A
```

mempool/mempool_handlers.go
```diff
@@ -0,0 +1,111 @@
...
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+
+		tx, err := txDecoder(req.GetTx())
+		if err != nil {
+			return nil, fmt.Errorf("decoding tx: %w", err)
+		}
+
+		if err := m.Insert(ctx, tx); err != nil {
+			return ErrAsCheckTxResponse(err), nil
```
mattac21: imo we should still remove the `ante handler failed: ` from this, it feels unnecessary to have that prefix all of the check tx errors and it differs from the typical behavior here (only having the ante handler error string in the log, plus when it undoes the wrapping it looks like its appending that tiny stack trace as well in the log which is not typically there). I wouldn't be surprised at all if some people depend on the string structure of the log value and this would possibly break them.

mempool/mempool_handlers.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		return &abci.ResponseCheckTx{Code: abci.CodeTypeOK}, nil
+	}
+}
+
+// ErrAsCheckTxResponse converts an error to a ResponseCheckTx object, respecting error wrapping.
+func ErrAsCheckTxResponse(err error) *abci.ResponseCheckTx {
```
mattac21, line 90: nit: should probably be private

mempool/mempool_handlers.go
```diff
@@ -0,0 +1,111 @@
...
+	// keep the original error message
+	log := err.Error()
+
+	space, code, _ := errorsmod.ABCIInfo(err, false)
```
mattac21: nit: the third return value `log` from `ABCIInfo` is the same as the `err.Error()` and it also handles if `err` is nil, which doing the `err.Error()` will panic on above. not a real issue since we only use this if `err` is specifically non nil, but since its public it would be good to use the version that doesnt panic on a nil `err`

mattac21 (approved):

mempool/mempool_handlers.go
swift1337: Updated

```bash
 ▸ ~/Code/evm ▸ git:feat/async-check-tx ✗ ▸  evmd tx bank send dev0 $(evmd keys show dev1 -a --keyring-backend test --home ~/.evmd) \
    1001atest \
    --chain-id 9001 --keyring-backend test --home ~/.evmd \
    --gas 200000 --fees 1atest -y
code: 13
codespace: sdk
data: ""
events: []
gas_used: "0"
gas_wanted: "0"
height: "0"
info: ""
logs: []
raw_log: 'gas prices too low, got: 0.000005000000000000atest required: 586181640.625000000000000000atest.
  Please retry using a higher gas price or a higher fee: insufficient fee'
timestamp: ""
tx: null
txhash: 8E03A239AF46D625112FCF590AB9C6B3CBA76E6913A8928DDDD3961B3CE8DA6A
```

Eric-Warehime (approved):

swift1337:
re: https://github.com/cosmos/evm/issues/962

PR #1119, opened 2026-04-21
refactor(mempool): Use direct `reaplist` + `txtracker` calls in `legacypool` instead of lifecycle callbacks

mattac21:
# Description

Refactors the legacypool & removes its use of lifecycle external hooks (`OnTxPromoted`, `OnTxRemoved`, etc) that callers could supply in order to change the behavior of promotions, removals, etc. Previously the top level mempool would supply these callbacks to add txs to the `reaplist` and `txtracker` when events happened. This however was confusing to have critical code running in the legacypool that was not clearly marked as a dependency inside of the legacypool.

Now, the `reaplist` and `txtracker` are directly passed into the legacypool and called directly within private `onTxPromted`, `onTxRemoved` functions. In order to import these in the legacypool, they had to move to their own package to avoid the import cycle (alternatively we could move the legacypool + all its file into the mempool package (no sub package), but this is a much larger refactor).

This refactor sets up being able to directly call the `reaplist` from within the recheck pool in the same way as the legacypool. We did not want to have two ways of interacting with the `reaplist` within the pools, since that would be very confusing.

Related to: STACK-1939, STACK-2450

---

## Author Checklist

I have...


mattac21:
@greptile try to review again

mempool/internal/reaplist/reaplist_test.go
```diff
@@ -1,4 +1,4 @@
-package mempool_test
+package reaplist_test
```
swift1337: Let's use the same package name (`reaplist`) as these tests have no cyclic deps

mempool/internal/txtracker/txtracker.go
```diff
@@ -36,36 +34,36 @@ var (
...
 		txCheckpoints: make(map[common.Hash]*checkpoints),
 	}
 }
 
 // Track initializes tracking for a tx. This should only be called from
 // SendRawTransaction when a tx enters this node via a RPC.
-func (txt *txTracker) Track(hash common.Hash) error {
-	txt.lock.Lock()
-	defer txt.lock.Unlock()
+func (t *Tracker) Track(hash common.Hash) error {
```
swift1337: Nit: how about renaming `Track`? It feels a bit generic right now, but the comment suggests it's actually the entry point of the tracking lifecycle.

swift1337 (approved):
LGTM, left minor comments

mattac21:
@greptile review pls

mempool/internal/txtracker/txtracker.go
mattac21: I feel like track is also just as generic, I think I'll go back to tx tracker I think that made the most sense. the clanker made this change and I just went with it but I dont love it to be honest.

mempool/internal/reaplist/reaplist_test.go
mattac21: oh yeah i missed updating this good catch, im surprised this is even allowed

PR #1120, opened 2026-04-22
fix(rpc): use latest block in eth_fillTransaction gas estimation

luchenqun:
## Summary

- use `rpctypes.EthLatestBlockNumber` when `eth_fillTransaction` auto-estimates gas
- align `SetTxDefaults` with geth behavior for transaction defaulting

## Why

geth uses `rpc.LatestBlockNumber` when it auto-estimates gas while filling transaction defaults:

Reference:
https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/transaction_args.go#L160-L161

```go
latestBlockNr := rpc.BlockNumberOrHashWithNumber(rpc.LatestBlockNumber)
estimated, err := DoEstimateGas(ctx, b, callArgs, latestBlockNr, nil, nil, b.RPCGasCap())
```

In contrast, the current implementation passes a numeric block `0`:

```go
blockNr := rpctypes.NewBlockNumber(big.NewInt(0))
```

On Cosmos EVM, numeric block `0` is not treated as `latest`, so `eth_fillTransaction` may estimate gas against the wrong historical state. This PR switches the default block selection to `latest` so the behavior matches geth.

## Testing

- `go test ./rpc/backend`


swift1337:
https://github.com/cosmos/evm/blob/0719a9f1058c80fb478f4d937cfb40ef19187d73/rpc/backend/comet.go#L111

PR #1121, opened 2026-04-22
fix: align local node mempool and browser access settings

luchenqun:
## Summary
- enable the app-side mempool flags required when `local_node.sh` switches CometBFT to app mempool mode
- update the generated CometBFT RPC config to bind on `0.0.0.0` and allow browser-origin requests during local development
- keep the local node startup flow consistent for browser-based debugging against local endpoints

## Problem
Running `local_node.sh` was repeatedly logging this startup error:

```text
10:42AM ERR AppMempool.reapTxs: error reaping txs error="ReapTxs handler not set" module=mempool server=node
```

The script switched CometBFT from flood mempool mode to app mode, but it did not also enable the app-side mempool flags required for the exclusive Krakatoa path.

## Testing
- `bash -n local_node.sh`

### DIFF
--- a/local_node.sh
+++ b/local_node.sh
@@ -267,7 +267,8 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
   sed -i.bak 's/timeout_commit = "5s"/timeout_commit = "1s"/g' "$CONFIG_TOML"
   sed -i.bak 's/timeout_broadcast_tx_commit = "10s"/timeout_broadcast_tx_commit = "5s"/g' "$CONFIG_TOML"
   sed -i.bak 's/type = "flood"/type = "app"/g' "$CONFIG_TOML"
-
+  sed -i.bak 's/laddr = "tcp:\/\/127.0.0.1:26657"/laddr = "tcp:\/\/0.0.0.0:26657"/g' "$CONFIG_TOML"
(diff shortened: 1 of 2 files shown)


local_node.sh
```diff
@@ -343,8 +344,11 @@ fi
 evmd start "$TRACE" \
 	--pruning nothing \
 	--log_level $LOGLEVEL \
+	--mempool.max-txs=0 \
 	--minimum-gas-prices=0atest \
 	--evm.min-tip=0 \
+	--evm.mempool.operate-exclusively=true \
```
swift1337, line 350: this flag was removed.

local_node.sh
```diff
@@ -343,8 +344,11 @@ fi
 evmd start "$TRACE" \
 	--pruning nothing \
 	--log_level $LOGLEVEL \
+	--mempool.max-txs=0 \
```
swift1337, line 347: this line is not needed as well

swift1337 (changes requested):

luchenqun:
Checked against current main: local_node.sh starts fine without these extra mempool flags. --evm.mempool.operate-exclusively was removed in #1112, and --mempool.max-txs already defaults to 0 in server/start.go, so this PR is no longer needed. Closing.

PR #1123, opened 2026-04-22
fix(mempool): Proper dropping from `reaplist` for cosmos txs during replacement and rechecking

mattac21:
# Description

Addresses two places where cosmos txs would become permanently stuck in the `reaplist`.
* When they were dropped due to failing recheck during the `runRecheck` loop.
* When they were replaced by another tx with the same sender + nonce that arrived later. 
  * To do this we use the `TxReplacement` config function for the `PriorityNonceMempool` that is internal to the `RecheckPool`. Previously, the `RecheckPool` saw its internal mempool as an `ExtMempool` interface that was passed in as a dependency. However this meant that the `RecheckPool` had to rely on the caller to create it with a correct `TxReplacement` function in order for it to function properly. This felt off, so I moved the creation of the `PriorityNonceMempool` and its config into the constructor of the `RecheckPool`, only relying on the caller to pass in a default config, if any.

Closes: STACK-2450, STACK-1939

---

## Author Checklist

I have...

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -304,20 +304,6 @@ func (m *Mempool) insert(tx sdk.Tx) (<-chan error, error) {
 	}
 }
 
-// insertAndReapCosmosTx inserts a cosmos tx into the cosmos mempool and sets
-// it to be reaped.
-func (m *Mempool) insertAndReapCosmosTx(tx sdk.Tx) error {
-	m.logger.Debug("inserting Cosmos transaction")
-
-	// Insert into cosmos pool (handles locking, ante handler, and address reservation internally)
-
-	m.logger.Debug("Cosmos transaction inserted successfully")
-	if err := m.reapList.PushCosmosTx(tx); err != nil {
-		panic(fmt.Errorf("successfully inserted cosmos tx, but failed to insert into reap list: %w", err))
-	}
-	return nil
-}
-
 // ReapNewValidTxs removes and returns the oldest transactions from the reap
 // list until maxBytes or maxGas limits are reached.
 func (m *Mempool) ReapNewValidTxs(maxBytes uint64, maxGas uint64) ([][]byte, error) {
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -160,19 +160,14 @@ func NewMempool(
 		panic("tx pool should contain only legacypool")
 	}
 
-	cosmosPoolConfig := config.CosmosPoolConfig
-	if cosmosPoolConfig == nil {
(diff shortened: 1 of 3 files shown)


mempool/mempool.go
```diff
@@ -311,24 +304,6 @@ func (m *Mempool) insert(tx sdk.Tx) (<-chan error, error) {
 	}
 }
 
-// insertAndReapCosmosTx inserts a cosmos tx into the cosmos mempool and sets
-// it to be reaped.
-func (m *Mempool) insertAndReapCosmosTx(tx sdk.Tx) error {
```
mattac21, line 316: inserting to the reaplist for cosmos txs is now fully done via the recheck pool

mempool/mempool.go
```diff
@@ -431,14 +406,13 @@ func (m *Mempool) RemoveWithReason(ctx context.Context, tx sdk.Tx, reason sdkmem
...
-	// Remove from cosmos pool (handles address reservation release internally)
+	// Remove from cosmos pool (handles address reservation release  and reap
+	// list drop internally)
 	err := sdkmempool.RemoveWithReason(ctx, m.recheckCosmosPool, tx, reason)
 	if err != nil {
 		m.logger.Error("Failed to remove Cosmos transaction", "error", err)
 		return err
 	}
-
-	m.reapList.DropCosmosTx(tx)
```
mattac21, line 441: removing cosmos txs from the reap list is now fully done by the recheck pool

mempool/mempool.go
```diff
@@ -603,31 +577,6 @@ func (m *Mempool) RecheckCosmosTxs(newHead *ethtypes.Header) {
 	m.recheckCosmosPool.TriggerRecheckSync(newHead)
 }
 
-func defaultCosmosPoolConfig(getDenom func() string) *sdkmempool.PriorityNonceMempoolConfig[math.Int] {
```
mattac21, line 606: moved to recheck pool, since we also need to define a `TxReplacement` function that requires the reaplist, which for adding and removing txs, is owned by the recheck pool

mempool/recheck_pool.go
```diff
@@ -73,13 +75,6 @@ type Rechecker interface {
 	Update(ctx sdk.Context, header *ethtypes.Header)
 }
 
-// LatestContextProvider provides the minimal methods needed by RecheckMempool
-// for context management during rechecks.
-type LatestContextProvider interface {
```
mattac21, line 78: previously this was minimal methods from the blockchain, but now we also need the GetDemon function which doesnt really fit this interface, so we are now just passing in the actual blochchain instead

mattac21:
@greptile re review this please

swift1337 (approved):

PR #1124, opened 2026-04-22
chore(tests): fix ibc tests failing with different balances

vladjdk:
closes: STACK-2601

Fixes IBC tests by copying the [upstream](https://github.com/cosmos/ibc-go/blob/v11.0.0/testing/chain.go#L219-L245) context getters to avoid wiping context between calls.

Also fixes ctx reuse by refreshing context every time a value is changed by keepers to reflect the use of cached contexts rather than the deprecated `NewUncachedContext()`.

### DIFF
--- a/testutil/ibc/chain.go
+++ b/testutil/ibc/chain.go
@@ -95,6 +95,13 @@ type TestChain struct {
 	// Short-term solution to override the logic of the standard SendMsgs function.
 	// See issue https://github.com/cosmos/ibc-go/issues/3123 for more information.
 	SendMsgsOverride func(msgs ...sdk.Msg) (*abci.ExecTxResult, error)
+
+	// nextBlockContextInitialized tracks whether NewNextBlockContext has been
+	// called for the current block. Calling it again would reset the finalize
+	// state and discard any writes made since the last call. GetContext uses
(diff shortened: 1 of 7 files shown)


testutil/ibc/chain.go
```diff
@@ -235,8 +242,32 @@ func NewTestChain(t *testing.T, isEVM bool, coord *Coordinator, chainID string)
...
+		chain.nextBlockContextInitialized = true
+	}
+
+	return chain.App.GetBaseApp().NewContextLegacy(false, chain.ProposedHeader)
+}
+
+// tryGetContextLegacy attempts to get a context from the finalize state.
+// Returns false if the finalize state is not initialized (e.g. between blocks).
+func (chain *TestChain) tryGetContextLegacy() (sdk.Context, bool) {
+	defer func() { recover() }() //nolint:errcheck // catch nil pointer panic
```
swift1337, line 268: should we return `ctx, false` here?

vladjdk: That's the default return value in this case

vladjdk: It defaults to zero-values (Context{}, false)

vladjdk:
@greptile re-review

swift1337 (approved):

PR #1126, opened 2026-04-23
fix(vm): hydrate evm globals

swift1337:
This PR streamlines the way we provision the global `evmCoinInfo`. It ensures that this variable is always present as soon as the node starts - either during genesis, immediately after startup, or in each `PreBlock`. 

Previously, after a node restart and before `PreBlock()`, all submitted RPC EVM transactions caused a panic because, during this brief window, `evmCoinInfo` was nil.

see `AppModule.HydrateGlobals()`

Example:
```
8:10PM INF Restored global evm variables from store module=vm origin=HydrateGlobals
```

Closes STACK-2380


evmd/app.go
```diff
@@ -775,6 +778,14 @@ func NewExampleApp(
...
 			os.Exit(1)
 		}
+
+		ctx := app.NewContextLegacy(true, cmtproto.Header{
+			Height:  app.LastBlockHeight(),
+			ChainID: app.ChainID(),
+		})
+
+		// set global evm variables from KV store
+		vmModule.HydrateGlobals(ctx)
```
vladjdk, line 788: Is there any way to not do this within the app.go? Ideally integrators don't even need to think about this. Seems like it would be easy to forget to include this in your binary.

aljo242:
@greptile re-review

evmd/app.go
swift1337: I agree. But atm I'm not aware of any way to "subscribe" to something like "app provisioned, here's the latest sdkCtx". 

FYI, omitting this call won't break anything, because the call is only necessary if you want to allow inbound RPC traffic right after the node starts, without waiting for the next block.

vladjdk (approved):

PR #1127, opened 2026-04-23
fix(mempool): fix race between queued tx promotion reorg and reset reorg causing stuck queued txs

mattac21:
# Description

Fixes a race condition that was causing flakey tests (`TestMempool_ReapNewBlock` & `TestMempool_ReapPromoteDemotePromote`) that would manifest as txs being not promoted from queued -> pending. 

This race happens when an account has txs in pending, then a new tx is inserted into queued, and a promotion is scheduled. At the same time, a new block arrives, and some delay (likely an existing reorg loop running, or just scheduler delay) causes the promotion reorg loop run + the new block reset reorg loop run to be combined (they both see `currDone != nil`, so they wait and run together). Since `reset != nil`, we will run `pool.resetInternalState`, which clears `pool.pendingNonces` to be the on chain nonce (every `.Get(addr)` will fall back to `statedb.GetNonce(addr)`) at the height we are resetting on top of. Now, we will then run `promoteExecutables`, which attempt to promote our queued tx to pending. This tx is ready based on the on chain state and the pending state (lets say the tx in queued has nonce 3, and in pending currently there is nonce 1 and 2). However, `promoteExecutables` will check `pool.pendingNonces` to get the latest nonce for this account based on the pending state, but we have just reset it, and it will show the on chain nonce (`0`), not the pending nonce (`2`). Thus, it will consider this queued tx still a nonce gapped tx, and incorrectly keep it in pending.

After `demoteUnexecutables` runs, we repopulate the `pendingNonces` map with the state of the pending pool based on the state of the pending pool post demotion, so the next time a promotion for this account happens (a new tx must arrive after rebuilding the `pendingNonces` map), this tx will be properly promoted.

To fix this, we rebuild the `pendingNonces` map right after we `resetInternalState` (which clears it), based on the current state of the pending pool **AND** the current state of the `latestIncludedNonce` cache. The `latestIncludedNonce` is a LRU cache of accounts -> nonce that we have seen included on chain via `FinalizeBlock` calling `mempool.Remove`. We defer the removal of these txs by included their nonces in the cache, and removing them during `demoteUnexecutables` via `pool.removeOlds`. However this means that some pending txs in the pending pool when we are rebuilding `pendingNonces` just after `resetInternalState` may actually be stale, and we are about to drop them because they were just included in a block. Thus, we also must repopulate `pendingNonces` with the max of the pending pool greatest nonce, and the `latestIncludedNonce` nonce (if available).

Additionally, there is another subtle issue with using `latestIncludedNonce` here, since this is a LRU cache. What if we evict an account -> nonce mapping since the cache is full? This would cause us to prefer the pending pool greatest nonce, which is actually stale compared to the nonce on chain, causing the promotion to again see a nonce gap and not promote our queued tx. However, it is documented that the user should configure the size of this cache to be > the maximum number of accounts they expect to see in a single block, so if configured correctly, this should never happen, plus, you would also need this race condition to happen at the same time, which is extremely rare. Thus, I am opting to leave this as a documented limitation for now.

Closes: STACK-2546

---

## Author Checklist

I have...

### DIFF
--- a/mempool/txpool/legacypool/legacypool.go
+++ b/mempool/txpool/legacypool/legacypool.go
@@ -1510,6 +1510,49 @@ type reorgInput struct {
 	events        map[common.Address]*SortedMap
 }
 
+// computePendingTipNonces returns a mapping from account address to the next
+// expected nonce, taking the pending pool entries into account, but not the
+// queued pool entries.
+func (pool *LegacyPool) computePendingTipNonces() map[common.Address]uint64 {
+	tips := make(map[common.Address]uint64, len(pool.pending))
+	for addr, list := range pool.pending {
+		// get the next nonce that should be included on chain after last
+		// pending pool tx for this account lands
+		nextPendingNonce := list.LastElement().Nonce() + 1
+
+		// grab the latest nonce that we have observed on chain for this
+		// account if we have it cached
+		latestIncludedNonce, ok := pool.latestIncludedNonce.Get(addr)
+		if ok {
+			// if we have a cached on chain nonce for this account, it likely
+			// had a tx land in the latest block and its last pending pool
+			// element may be stale if we have not yet run demoteUnexecutables
+			// yet. thus, if the latestIncludedNonce + 1 is > nextPendingNonce,
+			// then that is actually what we should be using as the 'tip' of
+			// the pending pool for this account
+			//
+			// NOTE: latestIncludedNonce is a lru cache (lru to manage its
+			// growth). It is strongly recommended to the user to configure
+			// this cache size > the max number of unique accounts they expect
+			// to see in a block. If they do not do so, an issue could manifest
+			// here where a block has more accounts than entries in the
+			// latestIncludedNonce cache. If an entry gets evicted from
+			// latestIncludedNonce for an account that was included in H-1,
+			// then calling this function at H will use a stale
+			// nextPendingNonce value (when compared to statedb.GetNonce(addr))
+			// if this function is called before demoteUnexecutables, causing a
+			// queued contiguous nonce tx (with respect to pending txs) to not
+			// be promoted if there is a race on resetting + promoting the
+			// queued tx after insert. This is unlikely and would require a
+			// misconfiguration by the user, but this is a documented possible
+			// scenario.
+			nextPendingNonce = max(nextPendingNonce, latestIncludedNonce+1)
+		}
+		tips[addr] = nextPendingNonce
+	}
+	return tips
+}
+
 // runReorg runs reset and promoteExecutables on behalf of scheduleReorgLoop.
 func (pool *LegacyPool) runReorg(input reorgInput) {
 	defer func(t0 time.Time) {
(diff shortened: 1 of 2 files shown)


mempool/txpool/legacypool/legacypool.go
```diff
@@ -1510,6 +1510,49 @@ type reorgInput struct {
...
+			// be promoted if there is a race on resetting + promoting the
+			// queued tx after insert. This is unlikely and would require a
+			// misconfiguration by the user, but this is a documented possible
+			// scenario.
+			nextPendingNonce = max(nextPendingNonce, latestIncludedNonce+1)
+		}
+		tips[addr] = nextPendingNonce
+	}
+	return tips
+}
```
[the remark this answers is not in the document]
mattac21: this is more of an existing issue with how `latestIncludedNonce` is populated (only via evm txs, not cosmos ones), right? as it stands, we also would not properly remove pending txs that become 'old' due to a cosmos tx bumping the accounts nonce. I will address this behavior in a new pr.

aljo242 (approved):

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1510,6 +1510,49 @@ type reorgInput struct {
...
 
+// computePendingTipNonces returns a mapping from account address to the next
+// expected nonce, taking the pending pool entries into account, but not the
+// queued pool entries.
+func (pool *LegacyPool) computePendingTipNonces() map[common.Address]uint64 {
+	tips := make(map[common.Address]uint64, len(pool.pending))
+	for addr, list := range pool.pending {
+		// get the next nonce that should be included on chain after last
+		// pending pool tx for this account lands
+		nextPendingNonce := list.LastElement().Nonce() + 1
```
swift1337, line 1521: I don't fully understand the logic 😅

Let's say we have an account `A` in the pending pool with `nonce=10` and its `latestIncludedNonce=9`. Why should we set `nextPendingNonce` to **11** (`list.LastElement().Nonce() + 1`)? Assuming this user won't submit any new txs in the next hour or so

mattac21: We use this map/`noncer`/`pendingNonces` to track what is the next valid nonce that should be used for an account, taking into account the state of the pending pool. In your example, we would set the `nextPendingNonce` to 11 for this account because on chain the accounts nonce is `10` (since `9` was just included, seen via `lastIncludedNonce`, and the tx with nonce `10` is still in the pending pool and hasn't been included yet). We then use this to determine what txs are ready to be promoted from the queued pool, based on pending pool state. 

To continue your example, let's say we also have a tx with `nonce=11` in the queued pool. To determine if tx with `nonce=11` is ready to promote, we check `queue[addr].Ready(pendingNonces.Get(addr))`. This says give me all queued txs for this account who's nonce is >= the next nonce that should be used for this account, taking into account the pending pool state. Since we have properly populated `pendingNonces` with `11` for this account (based on current pending pool state, i.e. last pending pool entry's nonce for this account, +1), instead of just the state on chain), we see that the queued tx is not a nonce gapped tx anymore and can be promoted.

swift1337 (approved):

PR #1132, opened 2026-04-24
fix: evm tx index and log indexes are not patched (alternative design)

vladjdk:
Alternative to #812.

Fixes EVM tx and log indexing under BlockSTM, and keeps RPC receipt indices consistent in mixed-tx blocks.

ctx.TxIndex() is actually correct under both sequential and STM runners. The SDK passes the original block-order index into deliverTx.

However, the SDK's STMRunner passes the original block-order index into deliverTx, which BaseApp forwards into ctx.TxIndex(). The only invariant parallel execution actually breaks is the cumulative log.Index, since it depends on the log count of every prior successful eth tx in the block.

These are patched once per block, post-execution, by evmtypes.PatchTxResponses, installed as a TxRunner wrapper via the new x/vm/runner package (runner.SetRunner is now required for sequential and STM runners alike).

Changes from #812:

- Restore ctx.TxIndex() in TxConfig, receipt.TransactionIndex, and the ante handler's EmitTxHashEvent. These were zeroed in the original patch.
- PatchTxResponses modifies existing event attributes in place instead of synthesizing new events.
- PatchTxResponses doesn't halt on non-TxMsgData payloads (similar to how Ethermint handled this case)
- Guard test for the SDK TxIndex assumptions (DefaultRunner/STMRunner pass the slice position into deliverTx).
- End-to-end RPC test (adapted from #1087) for block-global log.Index across receipts.


vladjdk:
@greptile review

x/vm/types/response.go
```diff
@@ -0,0 +1,101 @@
...
+	for eIdx := range events {
+		if events[eIdx].Type != EventTypeEthereumTx {
+			continue
+		}
+		for aIdx := range events[eIdx].Attributes {
+			if events[eIdx].Attributes[aIdx].Key == AttributeKeyTxIndex {
+				events[eIdx].Attributes[aIdx].Value = strconv.FormatUint(start+uint64(n), 10) //#nosec G115 -- int overflow is not a concern here
+			}
+		}
+		n++
```
mmsqe: n++ need move inside the check?

vladjdk: Good catch, fixed.

rpc/backend/utils.go
```diff
@@ -355,3 +355,17 @@ func unwrapBlockNOrHash(blockNOrHash types.BlockNumberOrHash) string {
 	}
 	return ""
 }
+
+// Fallback to find tx index by iterating all valid eth transactions for legacy blocks that don't have AttributeKeyTxIndex in events
+func FindEthTxIndexByHash(ctx context.Context, txHash common.Hash, block *cmtrpctypes.ResultBlock, blockRes *cmtrpctypes.ResultBlockResults, b *Backend) (int32, error) {
```
mattac21: nit: should this be a method on the backend?

x/vm/keeper/abci.go
```diff
@@ -49,7 +49,6 @@ func (k *Keeper) EndBlock(ctx sdk.Context) (err error) {
 	}
 
 	k.CollectTxBloom(ctx)
-	k.ResetTransientGasUsed(ctx)
```
mattac21, line 52: why did this get removed?

x/vm/types/response.go
```diff
@@ -0,0 +1,102 @@
...
+//
+// Must be invoked once per block on the full ExecTxResult slice produced by
+// the TxRunner.
+func PatchTxResponses(input []*abci.ExecTxResult) ([]*abci.ExecTxResult, error) {
+	var (
+		ethTxIndex uint64
+		logIndex   uint64
+	)
+	for _, res := range input {
+		rewritten := rewriteEthTxEventIndex(res.Events, ethTxIndex)
```
mmsqe, line 38: seems PatchTxResponses advances ethTxIndex for all failed eth txs now, might need align like indexer skips failures via TxSucessOrExpectedFailure

vladjdk: Fixed: https://github.com/cosmos/evm/pull/1132/commits/801eb0ad22c30e62bc2e5a2fbf23559c6cbdbc06

mattac21 (approved):

x/vm/keeper/abci.go
vladjdk: Dead code - the object store wipes automatically

rpc/backend/utils.go
vladjdk: Good call, yes

mmsqe (approved):

PR #1138, opened 2026-04-25
fix: use latest block for eth_fillTransaction gas estimation

luchenqun:
## Summary
- use `EthLatestBlockNumber` instead of numeric `0` when `eth_fillTransaction` estimates gas
- keep `eth_fillTransaction` aligned with the latest state instead of block 1

## Why
`SetTxDefaults` currently builds a `BlockNumberOrHash` with `NewBlockNumber(big.NewInt(0))` before calling `EstimateGas`.
That is not equivalent to `latest` in this codebase:

- `BlockNumberFromComet` returns the numeric block number unchanged
- `BlockNumber.Int64()` maps numeric `0` to `1`
- so `CometHeaderByNumber` resolves block 1, not the latest block

Using `EthLatestBlockNumber` makes the call match the intended latest-state behavior.

## Verification
- `go test ./rpc/backend -run TestDoesNotExist -count=0`

Related: #1120


luchenqun:
@swift1337 the crux is in `getHeightByBlockNum` itself. Here is the relevant code path:

```go
func (b *Backend) getHeightByBlockNum(ctx context.Context, blockNum rpctypes.BlockNumber) (height int64, err error) {
	ctx, span := tracer.Start(ctx, "getHeightByBlockNum", trace.WithAttributes(attribute.Int64("blockNum", blockNum.Int64())))
	defer func() { evmtrace.EndSpanErr(span, err) }()

	if blockNum == rpctypes.EthEarliestBlockNumber {
		status, err := b.ClientCtx.Client.Status(ctx)
		if err != nil {
			return 0, errors.New("failed to get earliest block height")
		}

		return status.SyncInfo.EarliestBlockHeight, nil
	}

	height = blockNum.Int64()
	if height <= 0 {
		// In cometBFT, LatestBlockNumber, FinalizedBlockNumber, SafeBlockNumber all map to the latest block height.
```

The important detail is that `eth_fillTransaction` was not passing `EthLatestBlockNumber`. It was passing plain numeric `0` via `NewBlockNumber(big.NewInt(0))`.

In this codebase, when `blockNum` is plain numeric `0`, `height = blockNum.Int64()` becomes `1`, not `0`.
So `if height <= 0` is never reached in that case. That behavior comes from `rpc/types/block.go`:

```go
// Int64 converts block number to primitive type
func (bn BlockNumber) Int64() int64 {
	if bn < 0 {
		return 0
	} else if bn == 0 {
		return 1
	}

	return int64(bn)
}
```

That means the old flow is effectively:

- `eth_fillTransaction` passes numeric `0`
- `BlockNumberFromComet` returns that numeric `0` unchanged
- `height = blockNum.Int64()` turns it into `1`
- the `height <= 0` fallback is skipped
- the RPC resolves block 1, not latest

So the comment in `getHeightByBlockNum` is correct for the special negative enum values like `EthLatestBlockNumber`, but it does not apply to plain numeric `0`.

This is also easy to verify in practice: if you deploy a contract after block 1 and then call `eth_fillTransaction` against that contract to estimate gas, the old path will still resolve block 1, so the estimation fails reproducibly.

vladjdk:
@greptile review

vladjdk (changes requested):
We should add a test for the intended behaviour here to prevent future regressions.

aljo242 (changes requested):
fix is correct, but two issues:

**ci has not run:** all workflows are `action_required` -- fork pr, needs maintainer approval to trigger.

**test doesn't pin the fix:** the "estimate gas with latest block" case mocks `EstimateGas` with `mock.AnythingOfType("*types.EthCallRequest")`, which matches any request regardless of block number. the specific regression (block 1 vs latest) isn't asserted and could be reintroduced without breaking the test. assert that the `BlockNumber` on the `EthCallRequest` is `EthLatestBlockNumber`.

PR #1140, opened 2026-04-28
fix: panic on krakatoa when max block gas is -1

vladjdk:
Changes -1 on max block gas to `MaxInt64` instead of `MaxUint64` to prevent Krakatoa from panicking during `SafeInt64(header.GasLimit)`


vladjdk:
@greptile review

mattac21 (approved):
I'm assuming the panic you're talking about is the one in the mempool rechecker if the cast to int64 fails? Should we change that to not panic? Is it valid for the eth header to have a uint64 gas limit with cosmos evm? since it is still typed to be a uint64? or do we expect that to never happen and that was only possible in this case because of the bad default

PR #1145, opened 2026-04-29
chore: Convert all metrics

Eric-Warehime:
# Description

We've missed a couple of metrics when converting to otel in the mempool.

These are __the last ones__.

Also adds an attribute to the insert queue metrics so we can distinguish cosmos versus evm insert queue.

Closes: #XXXX

---

## Author Checklist

I have...

### DIFF
--- a/mempool/internal/queue/queue.go
+++ b/mempool/internal/queue/queue.go
@@ -28,7 +28,7 @@ var (
 func init() {
 	var err error
 	queueSize, err = meter.Int64Gauge(
-		"mempool.inserter.queue_size",
+		"insert_queue.queue_size",
 		metric.WithDescription("Number of txs waiting in the inserter queue"),
 	)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


aljo242 (approved):

mempool/internal/queue/queue.go
```diff
@@ -1,16 +1,50 @@
...
-	"github.com/cosmos/cosmos-sdk/telemetry"
+	// insertDuration is the latency of a batch insert into the underlying
+	// mempool.
+	insertDuration metric.Float64Histogram
 )
 
+func init() {
+	var err error
+	queueSize, err = meter.Int64Gauge(
+		"mempool.inserter.queue_size",
```
mattac21: should we be using `mempool.` here? all of the other components in the internal package use their name + . (`reap_list.`, `height_sync.`). either this should be `insert_queue.` or we should change all the other ones to also be `mempool.reap_list.`, etc

mattac21 (approved):

PR #1146, opened 2026-04-29
fix(mempool)!: Remove VM keeper end blocker notification, use `PrepareCheckStater` instead

mattac21:
# Description

Removes the ability to notify the mempool of new blocks within the vm keepers end blocker. This is strictly incorrect now since it will trigger a reset of the mempool on stale state. This behavior is replicated by using the `PrepareCheckStater` hook that is called just after a block is committed and the check state is updated.

Closes: STACK-2688

---

## Author Checklist

I have...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### DEPENDENCIES
 
+### API-BREAKING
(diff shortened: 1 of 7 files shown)


mattac21:
@greptile re review pls

Eric-Warehime (approved):

PR #1147, opened 2026-04-29
chore(mempool): add tests and document 7702 authority behaviour

vladjdk:
Closes: STACK-2668

Context:
  - Type-04 EVM tx that delegates an EOA's code to a contract. Has a sender (pays gas) and authorizations signed by authorities (the EOAs being delegated).
  - Sender's nonce always +1 on inclusion. Each successfully applied authorization bumps its authority's nonce +1.

  Added:

  - `LegacyPool.LatestNonce` as read counterpart to SetLatestNonce.
  - Test for three scenarios:
    - `stale-sender-no-7702`: Tests regular removals of stale sender nonces in the reset loop
    - `authority-after-cross-account-7702`: Tests for stale removals of an authority signed by a different sender in the reset loop
    - `sender-+1-gap-after-self-sponsored-7702`: Tests for a self-sponsored stale removal of a += 2 nonce in the reset loop

Decided not to pursue adding authorities to the signer extractor as we would need to get nonce state to ensure the authorizations pass. There's a case where the authorizations fail, and that should not increment the nonce. There's just no good way of figuring that out within the extractor without threading in some kind of data. Keeping it light and following up with STACK-2686.

### DIFF
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
+	"fmt"
 	"math/big"
 	"strconv"
 	"sync"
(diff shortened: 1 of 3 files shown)


Eric-Warehime (approved):

vladjdk:
@greptile review

PR #1149, opened 2026-04-29
fix: reap list size and large tx bug

vladjdk:
Closes STACK-2669, STACK-2670.

This PR fixes a few things:
- The reap list such that a tx that exceeds absolute per-block max would wedge the reaping. Now, if the tx is oversized, we tombstone the slot and notify the owning subpool via a drop callback. If it's over budget, we keep it for the next block.
  - Also adds a callback that fires after the lock is released so sub-pool drops can re-enter the reap list safely.
- Adds a maxSize cap on the reaplist, defaulting to `cosmosPoolMaxTx` + `GlobalSlots` + `GlobalQueue`. Pushes past the cap return ErrReapListFull.
- New metric: `reap_list.evicted_txs` with reason label.

Realistically, there are mitigations for this upstream in our default config. LegacyPool gas check at insertion, cosmos RecheckCosmos runs ante's gas check. Real exposure is non-standard ante chains, bytes-oversized cosmos txs without MaxTxBytes, or runtime gov change to block.MaxGas. All pretty uncommon cases, but this could serve as a DOD for them.

This PR closes the bug regardless and removes the dependency on those upstream mitigations.

### DIFF
--- a/mempool/internal/reaplist/metrics.go
+++ b/mempool/internal/reaplist/metrics.go
@@ -96,7 +96,7 @@ func init() {
 
 	metrics.evictedTxs, err = meter.Int64Counter(
 		"reap_list.evicted_txs",
-		metric.WithDescription("Total number of transactions evicted from the reap list, by reason (oversized_bytes, oversized_gas, cap_full)"),
+		metric.WithDescription("Total number of transactions evicted from the reap list, by reason (oversized_bytes, oversized_gas, oversized_both, cap_full)"),
 	)
 	if err != nil {
 		panic(err)
--- a/mempool/internal/reaplist/metrics.go
+++ b/mempool/internal/reaplist/metrics.go
@@ -42,8 +42,7 @@ func (rlm *reapListMetrics) TxDropped(txType txType) {
 	rlm.droppedTxs.Add(context.Background(), 1, metric.WithAttributeSet(attributes))
 }
 
-// TxEvicted records that a tx was evicted from the reap list with the given
-// reason (e.g. "oversized_bytes", "oversized_gas", "cap_full").
+// TxEvicted records that a tx was evicted from the reap list with the given reason.
 func (rlm *reapListMetrics) TxEvicted(reason string) {
 	attributes := attribute.NewSet(attribute.String("reason", reason))
 	rlm.evictedTxs.Add(context.Background(), 1, metric.WithAttributeSet(attributes))
--- a/mempool/internal/reaplist/metrics.go
+++ b/mempool/internal/reaplist/metrics.go
@@ -16,6 +16,7 @@ type reapListMetrics struct {
 	numIndexTxs metric.Int64Gauge
 	pushedTxs   metric.Int64Counter
 	droppedTxs  metric.Int64Counter
+	evictedTxs  metric.Int64Counter
 }
 
 var metrics = reapListMetrics{}
@@ -39,6 +40,13 @@ func (rlm *reapListMetrics) TxPushed(txType txType) {
 func (rlm *reapListMetrics) TxDropped(txType txType) {
 	attributes := attribute.NewSet(attribute.String("tx_type", string(txType)))
 	rlm.droppedTxs.Add(context.Background(), 1, metric.WithAttributeSet(attributes))
+}
+
+// TxEvicted records that a tx was evicted from the reap list with the given
+// reason (e.g. "oversized_bytes", "oversized_gas", "cap_full").
+func (rlm *reapListMetrics) TxEvicted(reason string) {
+	attributes := attribute.NewSet(attribute.String("reason", reason))
+	rlm.evictedTxs.Add(context.Background(), 1, metric.WithAttributeSet(attributes))
 }
 
 // RecordNumTxs records the number of entires in txs
@@ -81,6 +89,14 @@ func init() {
 	metrics.droppedTxs, err = meter.Int64Counter(
 		"reap_list.dropped_txs",
 		metric.WithDescription("Total number of transactions that have been dropped from the reap list"),
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	metrics.evictedTxs, err = meter.Int64Counter(
+		"reap_list.evicted_txs",
+		metric.WithDescription("Total number of transactions evicted from the reap list, by reason (oversized_bytes, oversized_gas, cap_full)"),
 	)
 	if err != nil {
 		panic(err)
--- a/mempool/internal/reaplist/reaplist.go
+++ b/mempool/internal/reaplist/reaplist.go
@@ -1,6 +1,7 @@
 package reaplist
 
 import (
+	"errors"
 	"fmt"
 	"slices"
 	"sync"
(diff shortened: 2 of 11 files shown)


vladjdk:
@greptile review

mempool/internal/reaplist/metrics.go
```diff
@@ -41,6 +42,12 @@ func (rlm *reapListMetrics) TxDropped(txType txType) {
 	rlm.droppedTxs.Add(context.Background(), 1, metric.WithAttributeSet(attributes))
 }
 
+// TxEvicted records that a tx was evicted from the reap list with the given reason.
+func (rlm *reapListMetrics) TxEvicted(reason string) {
```
mattac21, line 46: could the reason become its own type
```go
type evictionReason string

const (
  reasonOversizedBytes evictionReason = "oversized_bytes"
  ...
)
```

mempool/internal/reaplist/reaplist.go
```diff
@@ -12,6 +13,33 @@ import (
...
+const (
+	reasonOversizedBytes = "oversized_bytes"
+	reasonOversizedGas   = "oversized_gas"
+	reasonOversizedBoth  = "oversized_both"
+	reasonCapFull        = "cap_full"
+)
+
+// TxKind identifies which sub-pool owns a tx so the drop callback can route
+// the eviction without re-decoding bytes.
+type TxKind int
```
mattac21, line 31: we now have a public and a private `TxKind` (other one is in metrics.go), can we consolidate?

mempool/internal/reaplist/metrics.go
mattac21: looks like we have almost this in reaplist.go, I think they should probably move into this file

mempool/internal/reaplist/reaplist.go
```diff
@@ -39,27 +69,51 @@ type ReapList struct {
...
+
 // Reap returns a list of the oldest to newest transactions bytes from the reap
 // list until either maxBytes or maxGas is reached for the list of transactions
 // being returned. If maxBytes and maxGas are both 0 all txs will be returned.
+//
+// Txs that exceed maxBytes or maxGas standalone are evicted (they can never fit
+// in any block) and the owning sub-pool is notified via dropCallback.
 func (rl *ReapList) Reap(maxBytes uint64, maxGas uint64) [][]byte {
 	rl.txsLock.Lock()
-	defer rl.txsLock.Unlock()
```
mattac21, line 56: since we are not deferring anymore if we ever hit the panic in here (which I dont think we should) and recover, this will deadlock on the next call to reap.

mempool/internal/reaplist/reaplist.go
```diff
@@ -73,7 +127,26 @@ func (rl *ReapList) Reap(maxBytes uint64, maxGas uint64) [][]byte {
 		txSize := uint64(len(tx.bytes))
 		txGas := tx.gas
 
-		// Check if adding this tx would exceed limits
+		// Standalone-oversized: evict so it can't wedge the head of the list.
+		oversizedBytes := maxBytes > 0 && txSize > maxBytes
+		oversizedGas := maxGas > 0 && txGas > maxGas
+		if oversizedBytes || oversizedGas {
```
mattac21, line 133: could we make figuring out the reason its own function

mempool/internal/reaplist/reaplist.go
```diff
@@ -236,12 +342,28 @@ func (rl *ReapList) drop(hash string) bool {
...
+// evict tombstones the slot at idx and removes the hash from the index.
+// Caller must invoke any associated dropCallback AFTER the lock is released.
+//
+// NOTE: this is not thread safe, callers should be holding the txsLock.
+func (rl *ReapList) evict(idx int, reason string) {
+	tx := rl.txs[idx]
+	if tx == nil {
+		return
+	}
+	delete(rl.txIndex, tx.hash)
```
mattac21, line 361: do we actually want to remove from the index here? wouldnt we want to keep the tx in the index so that it is never re added to the reap list, if the tx is potentially demoted + re promoted? because in that case it is still going to be oversized and we will just remove it again

mempool/internal/reaplist/reaplist.go
```diff
@@ -236,12 +342,28 @@ func (rl *ReapList) drop(hash string) bool {
...
+	rl.txs[idx] = nil
```
mattac21, line 362: im not sure we need to do this either, we are advancing the `nextStart` past this tx's idx when we determine that it should be evicted, so it will get removed from the list once we reslice the `txs` slice at the end of reap, right?

mempool/mempool.go
```diff
@@ -169,7 +170,33 @@ func NewMempool(
...
 
-	reapList := reaplist.New(NewTxEncoder(txConfig))
+	// Sized to hold every tx that could legally live in both sub-pools at once.
+	reapListCap := resolveReapListCap(cosmosPoolMaxTx, legacyConfig.GlobalSlots, legacyConfig.GlobalQueue)
+
+	// Pool refs are populated below; the callback only fires from Reap, which
+	// can't run before NewMempool returns.
+	var (
+		legacyPoolRef     *legacypool.LegacyPool
+		recheckCosmosPool *RecheckMempool
```
mattac21, line 180: does this have to be a separate variable from the actual `legacyPool` and `recheckPool`? couldn't we do
```go
var (
  legacyPool *legacypool.LegacyPool
  recheckPool *RecheckMempool
)

reapDropCallback := func(...) {
  legacyPool.removetx()
  ...
}

// note not using `:=`
legacyPool = legacypool.New()
recheckPool = NewRecheckPool()
```

mempool/mempool.go
```diff
@@ -169,7 +170,33 @@ func NewMempool(
...
+	)
+	reapDropCallback := func(hash string, kind reaplist.TxKind, cosmosTx sdk.Tx) {
```
mattac21, line 182: we could simplify this interface to only be a sdk.Tx that is passed for both cosmos & evm txs (so we would have to cache the evm tx as an sdk.Tx in the reap list, instead of only doing that for cosmos txs), then we could use the mempools function `evmTxFromCosmosTx` to get if the tx is cosmos or evm, and then decide which pool to use based on that. that would eliminate the hash param from and and allow us to remove the public `TxKind` all together.

mempool/internal/reaplist/reaplist.go
```diff
@@ -175,14 +270,25 @@ func (rl *ReapList) PushCosmosTx(tx sdk.Tx) error {
...
+	rl.liveCount++
 
 	metrics.RecordNumTxs(rl.txs)
 	metrics.RecordNumIndexTxs(rl.txIndex)
 }
 
+// atCapacityLocked reports whether the reap list is at its configured cap.
+//
+// NOTE: this is not thread safe, callers should be holding the txsLock.
+func (rl *ReapList) atCapacityLocked() bool {
```
mattac21, line 285: since we now may not be able to push onto the reap list when its at capacity, this means that a tx will no longer be gossiped when it should be. should we handler callers that are trying to push in a new way now? for example if we are trying to promote an evm tx from queued to pending and fail to push onto the reap list, should we delay the promotion and wait until its no longer full?

PR #1150, opened 2026-04-29
fix(mempool): align evm rechecking with baseapp & add integration tests

mattac21:
# Description

Aligns the EVM rechecking logic to be in line with how `baseapp` handles `anteHandler` failures in `RunTx`. The main flow for promotion and demotion is now as follows:
1. Fetch fresh `CacheContext` from `Rechecker` for every tx that is being checked.
2. Recheck Tx
3. Write the store updates from the ante handler execution back to the `Recheckers` parent context **only** if we have successfully executed the ante handler sequence. For promote executables, this differs slightly where we will not `write()` if we are resetting due to a new block (this logic has not changed).

We no longer update the `ctx` with the `ctx` returned from the ante handler. We only care bout doing this when the ante handler has succeeded, and we already get this property by doing a `write()` and refetching the context on each recheck.

This PR also update the `demoteUnexecutables` recheck `FilterSorted` to track if a previous tx has been removed from the list due to recheck. This removes the need for the confusing `tolerateErr(err) != nil` check that would always be happening, since the pending pool explicitly disallows this case. We now only tolerate nonce gap errors during pending rechecking _if_ a previous tx has been removed by recheck.

This also adds integration tests for different rechecking scenarios where txs should be dropped, invalided, or demoted/promoted.

Closes: STACK-2659, STACK-2658

---

## Author Checklist

I have...


mattac21:
@greptile review

mempool/txpool/legacypool/legacypool.go
```diff
@@ -1853,39 +1850,37 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address, cancelled
 		var olds types.Transactions
 		if reset != nil {
 			olds = pool.removeOlds(addr, list, Queue)
-			queuedRemovedOld.Add(context.Background(),int64(len(olds)))
+			queuedRemovedOld.Add(context.Background(), int64(len(olds)))
 		}
 
-		// Drop all transactions that now fail the pools RecheckTxFn
-		//
-		// Note this is happening after the nonce removal above since this
```
vladjdk, line 1861: Is this no longer the case?

mempool/txpool/legacypool/legacypool.go
```diff
@@ -2070,33 +2065,53 @@ func (pool *LegacyPool) demoteUnexecutables(cancelled chan struct{}, reset *txpo
...
 		// Drop all transactions that are below the latest included nonce for
 		// this account based on what we have seen during the latest block
 		// execution.
 		olds := pool.removeOlds(addr, list, Pending)
-		pendingRemovedOld.Add(context.Background(),int64(len(olds)))
+		pendingRemovedOld.Add(context.Background(), int64(len(olds)))
 
-		// Drop all transactions that now fail the pools RecheckTxFn
-		//
-		// Note this is happening after the nonce removal above since this
```
vladjdk, line 2085: Ditto

mattac21:
@greptile re review

vladjdk (approved):

mempool/txpool/legacypool/legacypool.go
mattac21: that is still the case it just didnt feel necessary to have a comment, but I can add it back

mattac21: added back

aljo242 (approved):

swift1337 (approved):

PR #1151, opened 2026-04-30
fix(mempool): atomic account reserver with reference counter

swift1337:
While working on the cosmos-only transaction benchmark, I discovered several errors in transaction reservation for RecheckMempool.

This PR guarantees that `reserver.Hold(addresses)/Reserve(addresses)` operates atomically (all or nothing). The reservation tracker now also supports `WithRefCount()`, which tracks the number of account holds for Cosmos. 

For example, if account A submits txs of nonce=0, then another nonce=0 (an invalid duplicate that fails recheck), followed by nonce=1, the 2nd tx incorrectly cleared the **entire** account because the reservation was per account rather than per tx.

Additionally, this PR streamlines certain configurations and constructors.


swift1337:
@greptile do your thing

mempool/reserver/reserver.go
```diff
@@ -86,62 +102,134 @@ type Reserver interface {
...
+			h.tracker.accounts[addr] = h.id
+			h.incReservations(1)
+		}
+
+		if h.enableRefCount {
+			h.refsCounter[addr]++
+		}
+	}
+
 	return nil
```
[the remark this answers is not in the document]
swift1337: Yeah, this is a known limitation. We don't have duplicates.

mempool/recheck_pool.go
```diff
@@ -166,13 +173,20 @@ func (m *RecheckMempool) Close() error {
...
-	addrs, err := m.acquireTxReservations(tx)
+	addrs, err := m.reserveTx(tx)
 	if err != nil {
 		return fmt.Errorf("acquiring reservations for tx: %w", err)
 	}
 
+	defer func() {
+		if err != nil {
+			// best effort cleanup
+			_ = m.reserver.Release(addrs...)
```
mattac21: should we log here or panic here if this happens? this should never happen right unless there is a bug or some weird scenario we are not thinking of

mattac21: same for the few other places where we ignore this error

mempool/reserver/reserver.go
```diff
@@ -86,62 +102,134 @@ type Reserver interface {
...
+
+	for _, addr := range addrs {
+		// add only if not already owned by us
+		if _, exists := h.tracker.accounts[addr]; !exists {
+			h.tracker.accounts[addr] = h.id
+			h.incReservations(1)
+		}
+
+		if h.enableRefCount {
+			h.refsCounter[addr]++
```
mattac21, line 144: should we check for overflow of refs counter for an address before we do this? like a safe add (and make sure we can do the addition in `canHold`. if someone overflows this then we would release their reservation during the next release call and get into a weird state. i can't think of how this would ever actually happen in a prod environments, so maybe this is overkill lol

mattac21 (approved):

mempool/reserver/reserver.go
swift1337: This requires submitting int64 txs for an account. I don't think it's realistic; we have other guard rails for this.

mempool/recheck_pool.go
swift1337: let's log an error over panicking 👍

PR #1152, opened 2026-04-30
chore(7702): authority-based nonce gap promotion testing

vladjdk:
Adds unit tests for nonce-gap closure promotions of queued txes for self and cross-account sponsored cases.

Adds systemtest for nonce-gap closure promotions of a self-sponsored 7702 case.


vladjdk:
@greptile review

mattac21 (approved):

PR #1155, opened 2026-04-30
chore(mempool): add reaplist wedge tests

vladjdk:
Adds a size check to the rechecker so that if governance lowers consensus MaxBytes below an tx, the next chain-head reset evicts it instead of letting it wedge the head of the reap list (Reap returns empty until the wedger is removed).

Imo this belongs in the antehandler alongside the GasWanted check, but that would require more surface area for an edge case we'll probably never encounter.

The fix is mostly for cosmos txes. EVM txes are largely handled by the legacypool's hardcoded txMaxSize=128KB, so the EVM wedge only happens if gov reduces MaxBytes below that. In that case, an EVM tx is added under the old MaxBytes and then removed on recheck under the new one. Slightly inefficient but it'll do.

Adds 3 test cases:
- Txs that exceedtxMaxSizeorhead.GasLimitdon't enter the mempool to begin with.
- If governance lowers block gas, a tx that exceeds it in the reap list gets booted by the next reset.
- If governance lowers max bytes, a tx that exceeds it in the reap list gets booted by the next reset.


vladjdk:
@greptile review

aljo242 (approved):

mempool/rechecker.go
```diff
@@ -76,9 +79,34 @@ func (r *TxRechecker) RecheckEVM(ctx sdk.Context, tx *ethtypes.Transaction) (sdk
...
+// oversized tx wedges the head of the reap list.
+//
+// TODO: belongs in the cosmos ante alongside the gas-wanted check; living
+// here for now to avoid the DeliverTx-skip gate the ante version needs.
+func (r *TxRechecker) checkMaxBytes(ctx sdk.Context, tx sdk.Tx) error {
+	cp := ctx.ConsensusParams()
+	if cp.Block == nil || cp.Block.MaxBytes <= 0 {
+		return nil
+	}
+	bz, err := r.txConverter.CosmosTx(tx)
```
mattac21, line 100: I think this may hurt performance quite a lot, I dont have an actual benchmark but before we add something like this check into every single recheck (the hottest path in the mempool) I think we need to benchmark it

mempool/rechecker.go
```diff
@@ -76,9 +79,34 @@ func (r *TxRechecker) RecheckEVM(ctx sdk.Context, tx *ethtypes.Transaction) (sdk
...
+	if err := r.checkMaxBytes(ctx, tx); err != nil {
+		return ctx, err
+	}
 	return r.anteHandler(ctx, tx, false)
 }
 
+// checkMaxBytes rejects txs whose encoded size exceeds the current
+// consensus MaxBytes. Comet enforces this at admission, but pool txs are
+// not re-evaluated when gov lowers MaxBytes — without this check, an
+// oversized tx wedges the head of the reap list.
```
mattac21, line 91: so when a gov prop happens and we lower the max tx bytes, we will no longer be able to ingest new txs over that size, but the actual MaxBytes or MaxGas that the reap list itself uses does not change, that is a config.toml variable that isn't derived from the consensus params.

mempool/rechecker.go
```diff
@@ -67,7 +68,9 @@ func (r *TxRechecker) RecheckEVM(ctx sdk.Context, tx *ethtypes.Transaction) (sdk
 	if err != nil {
 		return sdk.Context{}, fmt.Errorf("converting evm tx %s to cosmos tx: %w", tx.Hash(), err)
 	}
-
+	if err := r.checkMaxBytes(ctx, cosmosTx); err != nil {
```
mattac21, line 71: maybe I misunderstood yesterday but I thought the solution we aligned on was that we would check on startup if the reap maxgas/maxbytes is < consensus max tx bytes and error if so

PR #1156, opened 2026-04-30
fix(mempool): defer reap of evm tx replacements until validation

mattac21:
# Description

Previously, we were adding evm txs that replaced an existing tx in the pending pool directly to the reap list. An invariant of the reap list is that all txs have been validated via the recheckers. This was breaking that invariant. This could have led to honest nodes spamming invalid txs over p2p. To fix, we mark txs that are replacing txs in the pending pool in a `toReap` lookup in the legacypool, and then when we recheck txs in the pending pool. check if the validated txs need to still be reaped.

Closes: STACK-2687

---

## Author Checklist

I have...


aljo242 (approved):

mattac21:
@greptile gimme that re review

aljo242 (approved):

PR #1158, opened 2026-05-01
chore: remove max bytes check on hotpath and instead fail on misconfigured reaplist on startup

vladjdk:
Removes max bytes check from the rechecker and instead checks for the misconfigured reaplist on startup. One caveat is that governance can potentially set the max gas to be higher than a node's reap max gas, but this is an edge case that we don't expect will be a problem. Can be resolved via a node restart with higher settings in that case.


vladjdk:
@greptile review

mattac21 (approved):

PR #1101, opened 2026-04-12
refactor: use slices.Contains to simplify code

chuanshanjida:
# Description

There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

---

## Author Checklist

**All** items are required. Please add a note to the item if the item is not applicable and
please add links to any relevant follow up issues.

I have...

- [ ] tackled an existing issue or discussed with a team member
- [ ] left instructions on how to review the changes
- [x] targeted the `main` branch

PR #1159, opened 2026-05-01
refactor: use slices.Contains to simplify code

chuanshanjida:
# Description

There is a [new function](https://pkg.go.dev/slices@go1.21.0#Contains) added in the go1.21 standard library, which can make the code more concise and easy to read.

The new pr with unit test for https://github.com/cosmos/evm/pull/1101

---

## Author Checklist

I have...

### DIFF
--- a/ante/cosmos/authz.go
+++ b/ante/cosmos/authz.go
@@ -2,6 +2,7 @@ package cosmos
 
 import (
 	"fmt"
+	"slices"
 
 	errorsmod "cosmossdk.io/errors"
 
@@ -79,11 +80,5 @@ func (ald AuthzLimiterDecorator) checkDisabledMsgs(msgs []sdk.Msg, isAuthzInnerM
 // isDisabledMsg returns true if the given message is in the list of restricted
 // messages from the AnteHandler.
 func (ald AuthzLimiterDecorator) isDisabledMsg(msgTypeURL string) bool {
-	for _, disabledType := range ald.disabledMsgTypes {
-		if msgTypeURL == disabledType {
-			return true
-		}
-	}
-
-	return false
+	return slices.Contains(ald.disabledMsgTypes, msgTypeURL)
 }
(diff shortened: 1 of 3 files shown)


chuanshanjida:
The newly added test code fully covers the changes made in this update.

image

and


image

aljo242 (changes requested):
**behavioral regression in `filters/utils.go`:** the original loop sets `match := len(sub) == 0` -- an empty sub-slice is a wildcard (matches any topic at that position). `slices.Contains(sub, ...)` on an empty slice returns `false`, so a filter like `[][]common.Hash{{}}` now rejects all logs at position 0 instead of accepting all. this breaks the ethereum filter spec.

**test gap:** the test case named "empty topic sub-array (wildcard)" uses `topics: [][]common.Hash{}` (empty outer array), which skips the loop entirely and doesn't exercise the broken path. add a case with `[][]common.Hash{{}}` (one empty inner slice = position-0 wildcard) and verify it still returns all logs.

the `authz.go` change is correct and clean.

chuanshanjida:
> **behavioral regression in `filters/utils.go`:** the original loop sets `match := len(sub) == 0` -- an empty sub-slice is a wildcard (matches any topic at that position). `slices.Contains(sub, ...)` on an empty slice returns `false`, so a filter like `[][]common.Hash{{}}` now rejects all logs at position 0 instead of accepting all. this breaks the ethereum filter spec.
> 
> **test gap:** the test case named "empty topic sub-array (wildcard)" uses `topics: [][]common.Hash{}` (empty outer array), which skips the loop entirely and doesn't exercise the broken path. add a case with `[][]common.Hash{{}}` (one empty inner slice = position-0 wildcard) and verify it still returns all logs.
> 
> the `authz.go` change is correct and clean.

Modified and also rebased the latest code.   Please review it again.  @aljo242  Thanks~

chuanshanjida:
@aljo242 Thanks for your review!  If there's anything need I to do, please feel free to let me know.

chuanshanjida:
@aljo242 friendly ping~

PR #1163, opened 2026-05-04
fix(mempool): Cosmos noop removals & keep future txs after parent's wrong sequence

mattac21:
# Description

Adds a number of integration tests for the cosmos recheck mempool and fix some subtle bugs related to the reserver and rechecking of cosmos txs.

- Changes removals in the cosmos pool to now be deferred to recheck. Not having this caused a subtle bug where if we had a mutli signer tx in the cosmos pool of signers A, B, C, then calling Remove on a different tx with signer A would then remove the multi signer tx of A, B, C. This gave us no opportunity to unreserve signers B and C, leaving them stuck with permanently reserved addresses. Instead, we just defer the removal to happen in the recheck loop instead. We only care about removals happening during FinalizeBlock, and all of those removals will be processed by the recheck loop evicting txs due to invalid sequence errors, except during recheck, we know exactly what txs are being removed.
- Does not invalidate future txs for a signer if the parent tx failed with an invalid sequence error during recheck. if a parent tx fails with this error, we know it must be a nonce too low situation (since we run recheck on insert, and that will catch all cases where nonce is too high). In this case, we do not want to invalidate future txs since they may still be valid. For example if we have txs for signer A in the mempool at nonce 0 and 1, and nonce 0 gets included on chain, we should remove tx for signer A with nonce 0, but keep nonce 1 if it is still valid on its own. Previously we would evict nonce 1 without even checking.

Closes: STACK-2593

---

## Author Checklist

I have...

### DIFF
--- a/mempool/recheck_pool.go
+++ b/mempool/recheck_pool.go
@@ -236,12 +236,12 @@ func (m *RecheckMempool) Insert(_ context.Context, tx sdk.Tx) (err error) {
 func (m *RecheckMempool) Remove(tx sdk.Tx) error {
 	// NOTE: processing the removal here produces a subtle bug for cosmos txs
 	// with multiple signers. the underlying mempool here (priority nonce
-	// mempool) identifies txs by only the first signer of mutli signer txs. so
+	// mempool) identifies txs by only the first signer of multi signer txs. so
 	// when we see a tx included in a block (currently the only spot where we
 	// care about remove being called for this mempool) with for example,
 	// signer A at nonce 0, and we have a different tx in the mempool with
 	// signer A at nonce 0 and signer B at nonce 0, removing the tx in the
-	// block with only signer A will remove the tx in the mempool with bother
+	// block with only signer A will remove the tx in the mempool with both
 	// signers A and B. the priority nonce mempool gives us no hook to see what
 	// tx was actually removed. to account for this, we must not remove during
 	// FinalizeBlock, and process the removal during recheck where the multi
@@ -488,10 +488,10 @@ func (m *RecheckMempool) runRecheck(done chan struct{}, newHead *ethtypes.Header
 			// nonce of this tx became too low compared to the chain state. we
 			// rerun ante handlers on insert to this pool, so at insert time,
 			// we know the nonce of the tx was not too high. since nonces only
-			// increase, ErrWrongSequence seen here must mean that the txs
-			// nonce became too low. we do not want to remove this since this
-			// singers tx at nonce+1 may still be valid and at the correct
-			// nonce now.
+			// increase, ErrWrongSequence seen here must mean that the tx's
+			// nonce became too low. we still remove this tx, but we do not
+			// want to cascade and evict the signer's tx at nonce+1 since
+			// that may still be valid at the correct nonce.
 			if errors.Is(err, sdkerrors.ErrWrongSequence) {
 				dropFuturesOnError = false
 			}
--- a/mempool/recheck_pool.go
+++ b/mempool/recheck_pool.go
@@ -2,11 +2,13 @@ package mempool
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 	"sync"
 	"time"
 
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	"github.com/ethereum/go-ethereum/common"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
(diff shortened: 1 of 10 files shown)


mattac21:
@greptile gimme that re review

Eric-Warehime (approved):

mempool/recheck_pool.go
```diff
@@ -229,17 +231,29 @@ func (m *RecheckMempool) Insert(_ context.Context, tx sdk.Tx) (err error) {
...
 
-// Remove removes a transaction from the pool.
+// Remove is a noop for this pool. All removals are processed during the async
+// recheck loop.
 func (m *RecheckMempool) Remove(tx sdk.Tx) error {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-
-	return m.removeLocked(tx)
+	// NOTE: processing the removal here produces a subtle bug for cosmos txs
```
swift1337, line 237: Should we keep the comment? In my opinion, it seems too detailed. Consider adding `// Remove noop. For correctness, all removals are happening async in recheck loop.`

mempool/recheck_pool.go
```diff
@@ -229,17 +231,29 @@ func (m *RecheckMempool) Insert(_ context.Context, tx sdk.Tx) (err error) {
...
+	// properly unreserve the signer A's and signer B's address reservations.
+	return nil
 }
 
-// RemoveWithReason removes a transaction from the pool. This must be
-// explicitly defined to prevent Go from promoting the embedded ExtMempool's
-// RemoveWithReason, which would bypass the reserver release logic.
+// RemoveWithReason is a noop for this pool. All removals are processed during
+// the async recheck loop. This must be explicitly defined to prevent Go from
+// promoting the embedded ExtMempool's RemoveWithReason.
```
swift1337, line 256: Same here

mempool/recheck_pool.go
```diff
@@ -474,17 +472,40 @@ func (m *RecheckMempool) runRecheck(done chan struct{}, newHead *ethtypes.Header
...
+			// nonce of this tx became too low compared to the chain state. we
+			// rerun ante handlers on insert to this pool, so at insert time,
+			// we know the nonce of the tx was not too high. since nonces only
+			// increase, ErrWrongSequence seen here must mean that the tx's
+			// nonce became too low. we still remove this tx, but we do not
+			// want to cascade and evict the signer's tx at nonce+1 since
+			// that may still be valid at the correct nonce.
+			if errors.Is(err, sdkerrors.ErrWrongSequence) {
+				dropFuturesOnError = false
+			}
```
swift1337, line 497:
```go
	removeTxs = append(removeTxs, txn)
```

if dropFuturesOnError == true, how do we add other account txs to the same  `removeTxs`?

mempool/recheck_pool.go
```diff
@@ -474,17 +472,40 @@ func (m *RecheckMempool) runRecheck(done chan struct{}, newHead *ethtypes.Header
 			}
 		}
 
+		dropFuturesOnError := true
```
swift1337: wdyt? seems a bit easier to read imo
```suggestion
 // If true, preserves other txs from this signer in the mempool (e.g. different nonces)
 keepFutureTxsOnError := false
```

```go
if errors.Is(err, sdkerrors.ErrWrongSequence) {
				keepFutureTxsOnError = true
}
```

```go
if keepFutureTxsOnError {
			iter = iter.Next()
			continue
}
```

mempool/recheck_pool.go
mattac21: I'm going to keep it, I think its important to call out why removals are not happening here, I think it is very non obvious if we dont have this detailed of a comment for future maintenance.

mempool/recheck_pool.go
mattac21: I think this is important to have

mempool/recheck_pool.go
mattac21: im not sure what you mean by this, can you explain your concern more? `dropFuturesOnError` is resetting after every txs recheck, if that is what you mean

mempool/recheck_pool.go
mattac21: sure, changed

swift1337 (approved):

mempool/recheck_pool.go
swift1337: upd, from the first glance, I didn't see that we loop for tx signers twice to check whether previous txs have marked the account for a wrong sequence.

PR #1173, opened 2026-05-06
fix: Enable otel during app start

Eric-Warehime:
# Description

Because the evm uses a custom start command it is not getting the otel initialized as it is [here](https://github.com/cosmos/cosmos-sdk/blob/e77c24c3eda790faf5f4ee4551323c0328b83457/server/start.go#L230)

Copy this over so that otel metrics work. Tested on local network and it produces metrics now.

---

## Author Checklist

I have...

### DIFF
--- a/server/start.go
+++ b/server/start.go
@@ -128,6 +128,11 @@ which accepts a path for the resulting pprof file.
 				return err
 			}
 
+			otelFile := filepath.Join(clientCtx.HomeDir, "config", telemetry.OtelFileName)
+			if err := telemetry.InitializeOpenTelemetry(otelFile); err != nil {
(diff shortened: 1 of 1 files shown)


mattac21 (approved):

mattac21:
technically to get around using this, users could start with the environment variable that points to their local otel file, right?

mattac21 (approved):

Eric-Warehime:
> technically to get around using this, users could start with the environment variable that points to their local otel file, right?

I think it would, but that was broken in otelconfig v0.23 release it seems. See https://github.com/cosmos/cosmos-sdk/pull/26390

Eric-Warehime:
@Mergifyio backport release/v0.7.x

PR #1181, opened 2026-05-14
feat: support state overrides in debug_traceCall

### LINKED ISSUE #1180: Support stateOverrides in debug_traceCall
## Summary

  Add `stateOverrides` support to `debug_traceCall` for geth-compatible JSON-RPC tracing.

  Cosmos EVM already supports state overrides for `eth_call` and `eth_estimateGas`, but `debug_traceCall` does not currently pass state override data into the trace execution path.

  ## Reproduction (for bugs)

  Desired JSON-RPC usage:

  ```json
  {
    "method": "debug_traceCall",
    "params": [
      {
        "from": "0x...",
        "to": "0x..."
      },
      "latest",
      {
        "tracer": "callTracer",
        "stateOverrides": {
          "0x0000000000000000000000000000000000000101": {
            "code": "0x600160005260206000f3"
          }
        }
      }
    ]
  }
```

  ## Impact

  This improves geth compatibility for debugging and tracing workflows.

  Affected users include:

  - JSON-RPC clients relying on geth-compatible debug_traceCall
  - Bundler can run a simulation for eip-7702.
  - tracing/debugging tools

  ## Related
https://github.com/cosmos/evm/pull/337

  ## Checklist

  - [x] Linked to a GitHub Issue (or this is the Issue)
  - [x] Repro steps included (for bugs)
  - [x] Impact described
  - [x] I understand minor typo/style doc fixes will not be accepted

nowooj:
# Description

This PR adds geth-compatible `stateOverrides` support to `debug_traceCall`.

The implementation keeps state overrides scoped to call tracing:
- adds `rpctypes.TraceCallConfig`, embedding the existing `TraceConfig` and adding `stateOverrides`
- updates the `debug_traceCall` RPC/backend path to accept `TraceCallConfig`
- adds `overrides` bytes to `QueryTraceCallRequest`
- passes raw state override JSON through the RPC backend into the gRPC request
- unmarshals overrides in `x/vm/keeper.TraceCall` using the existing `rpctypes.StateOverride` format
- forwards parsed overrides into `ApplyMessageWithConfig`
- keeps `debug_traceTransaction` and `debug_traceBlock*` on the existing `TraceConfig` path

## Author Checklist

I have...

Closes: #1180


aljo242 (changes requested):
**dos risk:** `req.Overrides` is unmarshaled from raw bytes with no size bound. on a public query node, a caller can send an arbitrarily large blob and force a large allocation. add a max-size check before unmarshal.

**`traceTxWithMsg` has 9 parameters** -- group `stateOverride` and the config args into a struct.

**missing behavioral test:** the tests register the override struct but don't verify that an override (e.g., balance or code) actually changes the trace output. need at least one table-driven case that sets an override and asserts the result reflects it.

nowooj:
> **dos risk:** `req.Overrides` is unmarshaled from raw bytes with no size bound. on a public query node, a caller can send an arbitrarily large blob and force a large allocation. add a max-size check before unmarshal.
> 
> **`traceTxWithMsg` has 9 parameters** -- group `stateOverride` and the config args into a struct.
> 
> **missing behavioral test:** the tests register the override struct but don't verify that an override (e.g., balance or code) actually changes the trace output. need at least one table-driven case that sets an override and asserts the result reflects it.

Thanks for the review, @aljo242.

**For the DoS concern:** I checked geth’s behavior and it does not enforce a stateOverride-specific size or depth limit. The request is bounded at the transport layer instead. We are already bounded by `json-rpc.http-body-limit`, `grpc.max-recv-msg-size` and `api.rpc-max-body-bytes` before the keeper receives the request.

Also, `eth_call` and `eth_estimateGas` currently follow the same state override unmarshalling pattern, so if we want additional defense-in-depth beyond the transport limits, I think it would be better to handle it consistently for all state override query paths in a follow-up rather than only adding a `debug_traceCall`-specific guard in this PR.

**For the `traceTxWithMsg` parameter count:** agreed. I updated the tracing helpers to group the trace execution options into a `traceTxConfig`.

**For the missing behavioral test**: also agreed. I updated the test traces the same call again without overrides to verify the underlying contract state remains unchanged.

nowooj:
This feature is required for bundlers using cosmos/evm to simulate userop.

PR #1184, opened 2026-05-14
fix: avoid missed wakeups in rpc stream subscribe

luchenqun:
## Description
Fix the rpc stream subscription race that could miss the final wakeup in blocking reads.
Initialize subscription offsets from the current tail so subscribers only receive new items.
Capture the cond channel before releasing the read lock to avoid dropped broadcasts.

## Verification
go test -count=20 -race -tags=test ./rpc/stream -run 'TestStream(ReadBlocking|Subscribe)'
go test -race -tags=test ./rpc/stream

### DIFF
--- a/rpc/stream/cond.go
+++ b/rpc/stream/cond.go
@@ -15,11 +15,19 @@ func NewCond() *Cond {
 	return &Cond{ch: make(chan struct{})}
 }
 
+func (c *Cond) Channel() <-chan struct{} {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	return c.ch
+}
+
 // Wait returns true if the condition is signaled, false if the context is canceled
 func (c *Cond) Wait(ctx context.Context) bool {
-	c.mu.Lock()
-	ch := c.ch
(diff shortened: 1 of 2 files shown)


aljo242:
two concerns before merge:

1. `stream.go:74`: changing `offset = -1` to `offset = s.lastID()` is a silent behavior change. any caller relying on replay-from-beginning semantics will silently stop receiving historical items. please document this in the changelog and confirm no existing `Subscribe` call sites depend on the old behavior.

2. no regression tests. the missed-wakeup race fix and the new offset behavior both need test coverage. the pr description describes manual verification steps but those do not prevent future regressions. please add a test that exercises concurrent `Add` + `ReadBlocking` under `-race` to pin the race fix, and a test for the new subscribe-from-tail behavior.

minor: `Channel()` in `cond.go` is exported but only used internally by `ReadBlocking`. unexport it.

aljo242 (changes requested):
silent behavior change in Subscribe offset, missing regression tests, unnecessary export of Channel().

PR #1214, opened 2026-07-02
fix(rpc): emit canonical block hash in newHeads subscription

puneet2019:
## Description
`eth_subscribe("newHeads")` publishes the derived Ethereum-header hash (`keccak(RLP(eth header))`), but `eth_getBlockByNumber` / `eth_getBlockByHash` report and resolve by the **CometBFT block hash** (`RPCMarshalHeader` sets `"hash"` to it; `GetBlockByHash` looks it up via CometBFT). So a `newHeads` notification's `hash` doesn't match the same block from `eth_getBlockByNumber`, and passing it to `eth_getBlockByHash` / `eth_getLogs({blockHash})` returns "block not found".

This **completes #725** — that PR aligned the JSON-RPC responses and added the canonical hash to the stream header (`stream.RPCHeader.Hash = data.BlockID.Hash`), but the WebSocket `subscribeNewHeads` handler was left emitting `header.EthHeader`.

**Change** (`rpc/websockets.go`):
```go
- Result: header.EthHeader,
+ Result: rpctypes.RPCMarshalHeader(header.EthHeader, header.Hash.Bytes()),
```
`RPCMarshalHeader` returns the full header map with the canonical `"hash"`, so `newHeads` now matches `eth_getBlockByNumber`.

### How to review / test
`newHeads.hash` now equals `eth_getBlockByNumber(<same height>).hash`, and that hash resolves via `eth_getBlockByHash` / `eth_getLogs({blockHash})` (previously "block not found"). Verified end-to-end in a downstream chain adopting `cosmos/evm/rpc` (WS `eth_subscribe` e2e green). Happy to add an integration assertion if preferred.

## Author Checklist
- [x] follow-up completing #725 (no separate issue)

Generated with Claude Code

### DIFF
--- a/rpc/websockets.go
+++ b/rpc/websockets.go
@@ -25,6 +25,7 @@ import (
 
 	rpcfilters "github.com/cosmos/evm/rpc/namespaces/ethereum/eth/filters"
 	"github.com/cosmos/evm/rpc/stream"
+	rpctypes "github.com/cosmos/evm/rpc/types"
 	"github.com/cosmos/evm/server/config"
 
 	"cosmossdk.io/log/v2"
@@ -463,7 +464,7 @@ func (api *pubSubAPI) subscribeNewHeads(wsConn *wsConn, subID rpc.ID) (context.C
 				Method:  "eth_subscription",
 				Params: &SubscriptionResult{
 					Subscription: subID,
-					Result:       header.EthHeader,
+					Result:       rpctypes.RPCMarshalHeader(header.EthHeader, header.Hash.Bytes()),
 				},
 			}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@
 ### BUG FIXES
 
 - [\#1223](https://github.com/cosmos/evm/pull/1223) Reject EVM txs below the base fee at mempool insert instead of silently queuing them.
+- [\#1214](https://github.com/cosmos/evm/pull/1214) Emit the canonical CometBFT block hash in the `newHeads` subscription so it matches `eth_getBlockByNumber` (completes [\#725](https://github.com/cosmos/evm/pull/725)).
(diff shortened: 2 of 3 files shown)


mmsqe (approved):

rpc/websockets.go
```diff
@@ -463,7 +464,7 @@ func (api *pubSubAPI) subscribeNewHeads(wsConn *wsConn, subID rpc.ID) (context.C
 				Method:  "eth_subscription",
 				Params: &SubscriptionResult{
 					Subscription: subID,
-					Result:       header.EthHeader,
+					Result:       rpctypes.RPCMarshalHeader(header.EthHeader, header.Hash.Bytes()),
```
mattac21, line 467: could you add a test to ensure we are returning the correct hash and dont regress?

mattac21 (changes requested):

rpc/websockets.go
puneet2019: hey, found out there were no tests for these, so was unsure where to add..

mattac21: test should go in the existing `rpc/websockets_test.go`. there isn't a test for the subscription functionality but I think this would be a great first test to add! I dont think this would require any changes to the actual code, for the test I think you should be able to pass the `*stream.RPCStream` with a mocked `rpcclient.EventsClient` to the existing `newTestWebsocketServer` function, then just add a new header event to the subscription.

mattac21 (approved):

mattac21:
Thanks @puneet2019! lgtm

mattac21:
ah @puneet2019 just realized you dont have verified signatures turned on which we require. I think you should be able to rebase the commits a new commit that is properly signed.

puneet2019:
> ah @puneet2019 just realized you dont have verified signatures turned on which we require. I think you should be able to rebase the commits a new commit that is properly signed.

done!

mattac21 (approved):

PR #1220, opened 2026-07-05
fix: Update txn estimates to target latest block

Eric-Warehime:
# Description

Use the latest block for estimating txn gas defaults. 

The existing default of block `0` will always run against the genesis state because the `Int64` conversion in [EstimateGas](https://github.com/cosmos/evm/blob/7c492afda512f3720b2d387ac581a5658a8ae1fd/rpc/backend/call_tx.go#L349) converts `0` to `1` [see here](https://github.com/cosmos/evm/blob/7c492afda512f3720b2d387ac581a5658a8ae1fd/rpc/types/block.go#L110), moving the gas estimation to use the genesis block. Passing `nil` here fixes the behaviour.


Eric-Warehime:
@greptile review

swift1337 (approved):

PR #1229, opened 2026-07-13
fix: prevent cosmos mempool proposal starvation under backlog

### LINKED ISSUE #1227: App-side mempool: cosmos-pool proposal starvation under large Cosmos-tx backlogs
## Summary

With `mempool.type = "app"` and a large backlog of **Cosmos** (non-EVM) transactions in the experimental app-side mempool, block proposals starve: blocks come out empty (or block production stalls for minutes) while the cosmos pool holds tens of thousands of valid transactions.

Two distinct failure modes, both reproduced on a single-validator devnet (i7-1185G7, Docker; cosmos/evm v0.7.0, cosmos-sdk v0.54.3, cometbft v0.39.3).

## Mode 1 — recheck cancellation starvation (short block times)

Config: `timeout_commit = 800ms`, ~16k pending cosmos txs (signed bank sends from 200 distinct senders, ~100 txs each, zero fee), sustained sync-broadcast flood at ~800–1,100 tx/s.

Observed: after a few blocks (~4.2k txs committed), **all subsequent blocks are empty** while the pool still holds ~16k valid txs; the backlog never drains (observed 60s+).

Mechanism as far as we could trace it in v0.7.0 source:

- Proposals only include cosmos txs already rechecked at the latest committed height: `getIterators → cosmosIterator → RecheckMempool.OrderedRecheckedTxs → heightsync.GetStore(height)`, with `pending-tx-proposal-timeout` (default 250ms) as the wait bound.
- `scheduleRecheckLoop`: a new block **cancels** the in-flight recheck pass and schedules a fresh one.
- `runRecheck` holds `RecheckMempool.mu` for the whole pass and starts a **fresh** heightsync store each height — a cancelled pass loses its validation progress.
- With a large pool, a full pass takes longer than the block interval → every pass is cancelled → the validated set never approaches the pool size.

**Unconfirmed link (help wanted):** a cancelled `runRecheck` still publishes its partial store via `defer EndCurrentHeight()`, and the proposal path waits up to 250ms before taking partial results — so we expected *small* blocks rather than *empty* blocks. In our reproduction blocks were fully empty, which suggests an additional interaction (possibly the store height requested by `getIterators` — `ctx.BlockHeight()-1` — racing with the height the recheck loop is currently rebuilding, hitting the height-behind path in `GetStore` and returning `nil`). Happy to provide full logs, goroutine dumps, and our load generator.

## Mode 2 — one giant block blows past timeout_propose (long block times)

Config: `timeout_commit = 15s`, same 20k-tx backlog, default `consensus.block.max_gas = -1`.

Observed: the entire backlog is reaped into **one 20,000-tx block**; building/verifying that proposal exceeds `timeout_propose` repeatedly, consensus goes through several timeout-escalation rounds, and the block lands ~4 minutes after its predecessor (verified via goroutine dump + consensus logs: `num_txs:20000`, executed in ~10s once agreed; no tx loss).

Workaround: set a finite `consensus.block.max_gas`. Suggestion: apply byte/gas reap bounds to the cosmos pool analogous to the EVM pool's `ValidateReapBounds`.

## Why this matters

The app-side mempool's lock-free `InsertTx`/`CheckTx` path gives large ingestion gains (we measured ~600 → ~2,200 tx/s for Cosmos txs with an app-level verify-once + parallel pre-verification on top of it). But chains with Cosmos-heavy traffic hit these two modes as soon as a backlog forms — and faster ingestion makes backlogs *more* likely, so ingestion improvements amplify the issue.

## Suggested directions (for discussion)

1. **Carry-forward:** seed the new height's store from the previous store, minus txs invalidated by the just-committed block (sender/nonce consumed) — keeps the validated set stable across cancellations.
2. **Resume cursor:** on cancellation remember the iteration position; the next pass starts there (earlier txs validated against a state one block older; per-sender sequence chains are still enforced at propose time).
3. **Chunked recheck outside `mu`:** validate in small batches, releasing the mutex between batches, so cancellation loses at most one batch.
4. Reap bounds for the cosmos pool (mode 2), independent of the above.

## Reproduction sketch

1. Single validator, `mempool.type = "app"`, EVM mempool enabled, `timeout_commit = 800ms` (mode 1) or `15s` + `max_gas = -1` (mode 2).
2. Fund 200 accounts; pre-sign 100 bank MsgSend per account (SIGN_MODE_DIRECT, zero fee).
3. Broadcast via `broadcast_tx_sync`, one goroutine per sender (sequence order preserved), sustained until 20k accepted.
4. Watch `num_txs` per block vs pool contents.

We can share our load generator and scripts if useful.

mmsqe:
# Description

* carry rechecked cosmos snapshot across heights instead of resetting each block
* purge stale txs from carried snapshot: failed rechecks, replaced txs, unkeyed txs
* serve current snapshot when recheck falls behind, never one beyond target height
* stamp entries with the height their validation ran against; evmd re-runs ante for any proposal tx not validated at the proposal base, so committed or invalidated carried txs are dropped at PrepareProposal and a lagging notify path fails closed
* run recheck passes in recheck mode, skipping sigverify crypto already done at insert
* speed up hot paths: signer index for store scans, tx identity index for stamp lookups and re-adds, pinned-generation header cache, SDK config-registry scope pin in evmd

200-tx bench:

| | before | after |
| --- | --- | --- |
| recheck pass | 15.92 ms (79.6 µs/tx) | 8.34 ms (41.7 µs/tx) |
| PrepareProposal, steady state | 1.02 ms (5.1 µs/tx) | 0.91 ms (4.5 µs/tx) |

Closes: https://github.com/cosmos/evm/issues/1227

---

## Author Checklist

I have...


dhfang:
@greptile review

swift1337:
cc @mattac21

mattac21:
@greptile review again

mempool/internal/heightsync/heightsync.go
```diff
@@ -250,10 +279,27 @@ func (hs *HeightSync[Store]) GetStore(ctx context.Context, height *big.Int) *Sto
...
+			if !hs.staleFallback {
+				// caller opted out of stale results, return nil
+				return nil
+			}
+			// Rather than starve the caller (e.g. an empty block proposal),
+			// fall back to the current carried-forward Store: it is at a
+			// height <= target and, even mid-recheck, holds a valid subset of
+			// validated txs. Safe to serve as long as producers keep it free
+			// of state a since-committed block invalidated (the cosmos pool
+			// does this via its committed-nonce watermark).
```
mattac21: >//  Safe to serve as long as producers keep it free
			// of state a since-committed block invalidated (the cosmos pool
			// does this via its committed-nonce watermark).

If this is safe to serve would really depend on the chains `PrepareProposal` and `ProcessProposal` implementations, correct? I'm assuming you mean this is safe because `PrepareProposal`/`ProcessProposal` would catch a block where the proposer serves stale txs that are actually invalid on the latest state, but they were proposed anyway via this?

mempool/internal/heightsync/heightsync.go
```diff
@@ -250,10 +279,27 @@ func (hs *HeightSync[Store]) GetStore(ctx context.Context, height *big.Int) *Sto
...
+			hs.mu.RLock()
+			value := hs.store
```
mattac21, line 300: why do this recapture? I dont think the `value` saved above within the lock would change value if the height advances why we waited on the timeout.

mattac21 (commented):
I like the approach, I've done a quick pass over the pr but a few questions before I do a deeper review. This would be changing some of the guarantees of the evm (cosmos + evm) memepool, right? This would now allow the mempool to return a set of txs that are actually invalid on the latest state according to ante handlers? However it would not ever return a tx that was previously committed.

mattac21:
Also  have you tested this under some load and has it properly solved https://github.com/cosmos/evm/issues/1227? I'm assuming you all have experienced the same issue?

mmsqe:
> I like the approach, I've done a quick pass over the pr but a few questions before I do a deeper review. This would be changing some of the guarantees of the evm (cosmos + evm) memepool, right? This would now allow the mempool to return a set of txs that are actually invalid on the latest state according to ante handlers? However it would not ever return a tx that was previously committed.

Yes, snapshot can now serve txs that would fail ante on latest state (`PrepareProposalVerifyTx` filters those at proposal time), but it never serves committed ones since pruning runs against current store during `FinalizeBlock`. There's e2e [test](https://github.com/MANTRA-Chain/mantrachain-e2e/pull/172/changes/da14894a0bc0a8b07eb0be5a2796e6e9bd1cfa3c) to reproduce.

mempool/internal/heightsync/heightsync.go
mmsqe: On height change `hs.store` is swapped, not mutated, `StartNewHeightFrom` does `hs.store = carry(hs.store)`, so old pointer become abandoned snapshot that stops receiving prunes (every write, including `PruneCommitted`, goes through `Do` on current store). It does change while waiting: when the timeout and `heightChanged` fire together, `select` picks a ready case at random. So serving pre-wait pointer would serve already-committed txs, the recapture under a fresh `RLock` returns the store that absorbed every prune, and `Cmp(height) > 0` guard rejects it if height skiped past the target.

mempool/internal/heightsync/heightsync.go
mmsqe: Yes, per-tx re-verification is the backstop, `PrepareProposalVerifyTx` drops failing candidates, `ProcessProposal` reruns ante, and committed txs can't reexecute past sequence/nonce checks at `FinalizeBlock`. `safe` here is narrower, not left to handlers: producers keep carried store free of just-committed txs, or proposer burns its window rejecting them. Enforced at source: `PruneCommitted` runs during `FinalizeBlock` before next `PrepareProposal` reads it, and `AddTx`'s watermark blocks racing re-adds.

mattac21:
> > I like the approach, I've done a quick pass over the pr but a few questions before I do a deeper review. This would be changing some of the guarantees of the evm (cosmos + evm) memepool, right? This would now allow the mempool to return a set of txs that are actually invalid on the latest state according to ante handlers? However it would not ever return a tx that was previously committed.
> 
> Yes, snapshot can now serve txs that would fail ante on latest state (`PrepareProposalVerifyTx` filters those at proposal time), but it never serves committed ones since pruning runs against current store during `FinalizeBlock`. There's e2e [test](https://github.com/MANTRA-Chain/mantrachain-e2e/pull/172/changes/da14894a0bc0a8b07eb0be5a2796e6e9bd1cfa3c) to reproduce.

I see, we would want to change evmd to not use the no verify implementation of `PrepareProposalVerifyTx`. We added this because with these mempools, there wouldn't be any point where the mempool returns a set of txs for proposal that are ante invalid. Changing this though would obviously have its own performance implications since its another set of ante handlers running on the proposal

mempool/internal/heightsync/heightsync.go
mattac21: i see yeah, my concern is similar to https://github.com/cosmos/evm/pull/1229#issuecomment-5146979276 where we dont use any ante verification during prepare proposal in evmd because of the current guarantees that this changes. that was done for perf reasons, so im curious how that change + this would modify perf.

mattac21:
> > > I like the approach, I've done a quick pass over the pr but a few questions before I do a deeper review. This would be changing some of the guarantees of the evm (cosmos + evm) memepool, right? This would now allow the mempool to return a set of txs that are actually invalid on the latest state according to ante handlers? However it would not ever return a tx that was previously committed.
> > 
> > 
> > Yes, snapshot can now serve txs that would fail ante on latest state (`PrepareProposalVerifyTx` filters those at proposal time), but it never serves committed ones since pruning runs against current store during `FinalizeBlock`. There's e2e [test](https://github.com/MANTRA-Chain/mantrachain-e2e/pull/172/changes/da14894a0bc0a8b07eb0be5a2796e6e9bd1cfa3c) to reproduce.
> 
> I see, we would want to change evmd to not use the no verify implementation of `PrepareProposalVerifyTx`. We added this because with these mempools, there wouldn't be any point where the mempool returns a set of txs for proposal that are ante invalid. Changing this though would obviously have its own performance implications since its another set of ante handlers running on the proposal

sorry for the reviewing delay @mmsqe , ill give this a more thorough review next week and hopefully get it merged

mattac21 (commented):
I think this generally looks good, another question I have is what actually was the bug that you/the issue reporter is experiencing. I agree that this seems like it would fix it, but why is every block empty past a certain threshold? The way this works pre this pr is that the pool will recheck as many txs as possible until a request for a proposal is received, and at that point it will return back whatever txs it has rechecked against the latest state. This may not be a lot of txs if we were late to finalize the prev block, but it should be _something_ sometimes, rather than always nothing which is surprising. Do you know why that is?

mempool/internal/heightsync/heightsync.go
```diff
@@ -116,6 +116,15 @@ type HeightSync[Store any] struct {
 	// fields of the Store itself
 	mu sync.RWMutex
 
+	// staleFallback makes GetStore return the current carried-forward Store
```
mattac21, line 119: we should modify evmd to not use the custom no verify process proposal handler, since this now breaks the assumption it makes that all batches reaching it are valid

mempool/recheck_pool_test.go
```diff
@@ -726,6 +728,276 @@ func TestRecheckMempool_RecheckedTxs(t *testing.T) {
...
+	close(gate)
+}
+
+// TestRecheckMempool_CarryForwardSurvivesCancellation verifies the fix for the
+// cosmos-pool proposal starvation: a recheck pass carries the previous height's
+// validated set forward, so a pass that is cancelled (or merely still running)
+// before it validates anything does not present an empty snapshot to proposals.
+// Before the fix, StartNewHeight reset the store to empty each height, so a
+// pass that had not yet re-added txs exposed a zero-length snapshot.
+func TestRecheckMempool_CarryForwardSurvivesCancellation(t *testing.T) {
```
mattac21, line 796: i would love to see this test in a more real world condition that reproduces the empty block issue, without manually blocking the ante execution which would trivially produce all empty blocks

mempool/internal/heightsync/heightsync.go
mmsqe: Just measured on a 200-tx block: steady-state PrepareProposal stays at main's cost (0.93–1.10ms vs 0.94–1.06ms) since txs the mempool proves validated at head skip ante, only stale carried txs pay full ante (bounded by block gas), and recheck passes end ~1.9x faster (~79 -> ~42us/tx)

mempool/internal/heightsync/heightsync.go
mmsqe: Yes, evmd now runs BaseApp ante over any proposal tx the mempool can't prove was validated at head (SnapshotVerifiedTxVerifier), at-head entries just encode, keep old no-verify cost in steady state.

mmsqe:
> I think this generally looks good, another question I have is what actually was the bug that you/the issue reporter is experiencing. I agree that this seems like it would fix it, but why is every block empty past a certain threshold? The way this works pre this pr is that the pool will recheck as many txs as possible until a request for a proposal is received, and at that point it will return back whatever txs it has rechecked against the latest state. This may not be a lot of txs if we were late to finalize the prev block, but it should be _something_ sometimes, rather than always nothing which is surprising. Do you know why that is?

GetStore's height-behind branch has no partial-results path, so one height of recheck lag returns nil instead of smaller set, and with the cap-1 event-bus subscription one missed beat drops subscriber permanently — past the backlog threshold the failure is a cliff instead of gradient: every later proposal is empty. This PR fixes what gets served while loop lags, but not the permanence: with a dead subscription the carried snapshot drains as its txs commit and nothing refills it. [stalepin_main](https://github.com/cosmos/evm/pull/1245) is the other half — driving `NotifyNewBlock` from PrepareCheckState every block keeps the trigger alive.

PR #1248, opened 2026-08-04
stringify json for EIP 712 encoding

codehans:
CosmWasm msg payloads are encoded as json objects in amino representation, which breaks EIP-712's strict typic requirements. This PR stringifies untyped JSON objects to make it compatible with EIP-712 signing


ethereum/eip712/message.go
```diff
@@ -48,6 +55,62 @@ func createEIP712MessagePayload(data []byte) (eip712MessagePayload, error) {
...
 }
 
+// stringifyJSONMsgFields converts object- and array-valued "msg" fields to
+// strings. Message fields conventionally contain opaque JSON (for example,
+// CosmWasm contract messages), whose runtime shape cannot be represented by a
+// stable EIP-712 type. Existing string values are left alone so ordinary
+// string fields and already-stringified JSON are not changed.
+//
+// The conversion only affects the derived EIP-712 payload. The protobuf
+// transaction still contains the original value used during execution.
```
Eric-Warehime, line 65: AI description--please make it more concise.

Eric-Warehime:
@greptile review

Eric-Warehime:
Seems good to merge when you update the comment.

ethereum/eip712/message.go
codehans: Not sure whether to be offended or not 😅. This is all hand written

Eric-Warehime: Sorry for offending, I should have explained more the particular thing I didn't like about the comment.

It's a bit verbose for an internal method comment that is doing regular json stringify. Also it specifically calls out CosmWasm contract messages as the thing it's solving for which is something that LLMs typically do when making small changes on behalf of a specific prompt.

On second read I think it's mostly fine, but I would still like to drop the CosmWasm callout.

ethereum/eip712/message.go
```diff
@@ -48,6 +55,62 @@ func createEIP712MessagePayload(data []byte) (eip712MessagePayload, error) {
 	return messagePayload, nil
 }
 
+// stringifyJSONMsgFields converts object- and array-valued "msg" fields to
+// strings. Message fields conventionally contain opaque JSON (for example,
+// CosmWasm contract messages), whose runtime shape cannot be represented by a
+// stable EIP-712 type. Existing string values are left alone so ordinary
+// string fields and already-stringified JSON are not changed.
```
Eric-Warehime, line 62:
```suggestion
// stringifyJSONMsgFields converts object- and array-valued "msg" fields to
// strings. Message fields conventionally contain opaque JSON 
// whose runtime shape cannot be represented by a stable EIP-712 type.
// Existing string values are left alone so ordinary
// string fields and already-stringified JSON are not changed.
```

PR #1275, opened 2026-09-01
fix(rpc): serve json-rpc without EVM mempool

### LINKED ISSUE #1274: json-rpc server refuses to start when the EVM mempool is disabled
`evmd start --mempool.max-txs=-1` with `json-rpc.enable = true` aborts the whole node with `json-rpc server requires backend.Mempool`. With the mempool disabled `configureEVMMempool` [return](https://github.com/cosmos/evm/blob/3e646c4356275ea78b35eadcc298f629b86b5506/evmd/mempool.go#L32-L35) before `SetMempool`, `GetMempool()` yields baseapp's [`NoOpMempool`](https://github.com/cosmos/cosmos-sdk/blob/v0.54.3/baseapp/baseapp.go#L196), which has no `GetTxPool()` and [fails](https://github.com/cosmos/evm/blob/3e646c4356275ea78b35eadcc298f629b86b5506/server/start.go#L567-L569) in `startInProcess`. Consensus is unaffected but RPC path breaks. (Must be cli flag: `start.go` force-sets `0`, shadowing `app.toml`)

[#873](https://github.com/cosmos/evm/pull/873/changes) turned a panic into error, the assertion predates [#1112](https://github.com/cosmos/evm/pull/1112/changes) which removed is the CometBFT broadcast fallback, until then the default path, leaving [SendRawTransaction](https://github.com/cosmos/evm/blob/3e646c4356275ea78b35eadcc298f629b86b5506/rpc/backend/call_tx.go#L156) to call `b.Mempool.Insert` unconditionally. `ValidateCrossConfig` still [accepts](https://github.com/cosmos/evm/blob/3e646c4356275ea78b35eadcc298f629b86b5506/server/config/config.go#L547-L548) both mempools disabled, config layer calls setup valid while server layer makes it fatal.

mmsqe:
# Description

With mempool.max-txs=-1 there is no EVM mempool, so the backend.Mempool assertion in startInProcess aborts the node when json-rpc is enabled. Drop the assertion and route submission through submitTx, which inserts when a mempool is present and otherwise broadcasts through CometBFT.

Closes: #1274

---

## Author Checklist

I have...


rpc/backend/call_tx.go
```diff
@@ -164,6 +157,30 @@ func (b *Backend) SendRawTransaction(ctx context.Context, data hexutil.Bytes) (r
...
+// broadcasting through CometBFT would only reach the app's CheckTx handler,
+// which inserts into the same pool but flattens the error to an ABCI code.
+// Without an EVM mempool (mempool.max-txs = -1) CometBFT's mempool is the
+// pool, so broadcasting is the only way in.
+func (b *Backend) submitTx(ctx context.Context, tx sdk.Tx) error {
+	if b.Mempool != nil {
+		return b.Mempool.Insert(ctx, tx)
+	}
+
+	txBytes, err := b.ClientCtx.TxConfig.TxEncoder()(tx)
```
swift1337: This breaks the invariant. The latest Cosmos EVM expects a Krakatoa mempool: https://docs.cosmos.network/evm/latest/documentation/migrations/migration-v0.6-to-v0.7#5a-krakatoa-application-layer-mempool

I propose implementing NopMempool and passing it to RPC, so it would return something like `ErrMempoolDisabled`, converted to a json_rpc error.

swift1337 (changes requested):

rpc/backend/call_tx.go
mmsqe: A bit confused by [ValidateCrossConfig](https://github.com/cosmos/evm/blob/6cf99f5964fcbbdb357052a6271c662b5c3d681b/server/config/config.go#L547-L548) and [test](https://github.com/cosmos/evm/blob/6cf99f5964fcbbdb357052a6271c662b5c3d681b/server/config/config_test.go#L142-L145). You mean doc supports this config in Step [5a](https://github.com/cosmos/evm/blob/6cf99f5964fcbbdb357052a6271c662b5c3d681b/docs/migrations/v0.6.x_to_v0.7.0.md?plain=1#L450) and in [opt out](https://github.com/cosmos/evm/blob/6cf99f5964fcbbdb357052a6271c662b5c3d681b/docs/migrations/v0.6.x_to_v0.7.0.md?plain=1#L501) is outdated, the intent is that stock CometBFT is no longer a target?

swift1337: After the Krakatoa release, the setup is as follows:

If a chain consists of CometBFT + Cosmos/EVM, they must use the Krakatoa mempool together with ABCI.InsertTx/ABCI.ReapTxs (or don't use the mempool at all). 

We eliminated the loop where the app (evm rpc in this case) received a transaction, rebroadcasted it to Comet, and then Comet pushed it back to the app. _CometBFT drives the consensus_, while the app (Cosmos/EVM) effectively builds a block during the propose step.

For your use case, if you want a read-only RPC, I agree there's room for improvement. That's why I propose implementing a NoOp-style component with empty methods returning `ErrNotImplemented` to eliminate the nil mempool in EvmRPC.

The documentation doesn't reflect this, and the test scenario you mentioned also lacks a case where Comet's mempool is flooded and max_txs = -1.

Thank you for spotting these gaps! I really appreciate it. Feel free to open a PR-happy to help.

server/start.go
```diff
@@ -564,11 +564,8 @@ func startInProcess(svrCtx *server.Context, clientCtx client.Context, opts Start
 			return fmt.Errorf("json-rpc server requires AppWithPendingTxStream")
 		}
 
-		mp, ok := evmApp.GetMempool().(backend.Mempool)
```
swift1337, line 567: We still need this check, NoOp mempool should implement `backend.Mempool`

swift1337 (approved):

PR #1277, opened 2026-09-03
fix: tx evm send should never silently fall back to zeroized to-addr

fragwuerdig:
# Description

The command `tx evm send` uses `arg[0]` as from-address and `arg[1]` as to-address. The command description says: "Both 0x or bech32". However, the to-address is parsed using unguarded  `utils.Bech32StringFromHexAddress(args[1])` - which has no error return value and silently asks the to-address to be an 0x address.

And if it is not, then `Bech32StringFromHexAddress()` will create near-zero or zero garbage-addresses. That can inflict a severe fund loss case for CLI users sending their funds accidentally to those garbage addresses while inputting genuine bech32 encoded addresses -- which the CLI help text said it would be safe.

# What is the Fix:

The fix is to basically use `accountToHex()` from the cli package - which only ever accepts 0x or bech32 formatted addresses in the first place and gives `error` in any other case. Then pull it's output through HexToAddress from go-ethereum (or gracefully propagate errors).

I also added tests to pin the behavior of `Bech32StringFromHexAddress()` and `accountToHex()` - for demonstration purposes only, happy to remove them if so requested as a matter of hygiene.

# How To Review

On main, note how the provision of a completely valid bech32 address accidentially creates a _valid_ transaction sending to the garbage address:

```
$ build/evmd tx evm send 0xD74A2f4Ef6f7F2D00EBE039F68829458C91871b6 cosmos1wv4xjvjvm8tz7pd8dy8n9khlprahan7sgft3p2 1stake --generate-only --chain-id blachain-1 | jq -r
{
  "body": {
    "messages": [
      {
        "@type": "/cosmos.bank.v1beta1.MsgSend",
        "from_address": "cosmos16a9z7nhk7ledqr47qw0k3q55try3sudke7x3pk",
        "to_address": "cosmos1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvzax4k6",
        "amount": [
          {
            "denom": "stake",
            "amount": "1"
          }
        ]
      }
    ],
    "memo": "",
    "timeout_height": "0",
    "unordered": false,
    "timeout_timestamp": null,
    "extension_options": [],
    "non_critical_extension_options": []
  },
  "auth_info": {
    "signer_infos": [],
    "fee": {
      "amount": [],
      "gas_limit": "200000",
      "payer": "",
      "granter": ""
    },
    "tip": null
  },
  "signatures": []
}
```

On my fix branch that bech32 to-address is not garbaged but taken verbatim:

```
$ build/evmd tx evm send 0xD74A2f4Ef6f7F2D00EBE039F68829458C91871b6 cosmos1wv4xjvjvm8tz7pd8dy8n9khlprahan7sgft3p2 1stake --generate-only --chain-id blachain-1 | jq -r
{
  "body": {
    "messages": [
      {
        "@type": "/cosmos.bank.v1beta1.MsgSend",
        "from_address": "cosmos16a9z7nhk7ledqr47qw0k3q55try3sudke7x3pk",
        "to_address": "cosmos1wv4xjvjvm8tz7pd8dy8n9khlprahan7sgft3p2",
        "amount": [
          {
            "denom": "stake",
            "amount": "1"
          }
        ]
      }
    ],
    "memo": "",
    "timeout_height": "0",
    "unordered": false,
    "timeout_timestamp": null,
    "extension_options": [],
    "non_critical_extension_options": []
  },
  "auth_info": {
    "signer_infos": [],
    "fee": {
      "amount": [],
      "gas_limit": "200000",
      "payer": "",
      "granter": ""
    },
    "tip": null
  },
  "signatures": []
}
```
---

## Author Checklist

I have...

- [x] left instructions on how to review the changes -

### DIFF
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -312,6 +312,70 @@ func TestAddressConversion(t *testing.T) {
 	require.Equal(t, hex, gotAddr.Hex())
 }
 
+// TestBech32StringFromHexAddressManglesNonHexInput pins the behavior
+// that non-hex addresses are parsed into garbage addresses
+func TestBech32StringFromHexAddressManglesNonHexInput(t *testing.T) {
+	config := sdk.GetConfig()
+	config.SetBech32PrefixForAccount("cosmos", "cosmospub")
+
+	const (
+		zeroAddr       = "cosmos1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqnrql8a"
+		zeroAddrPlus0C = "cosmos1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqvzax4k6"
+	)
+
+	testCases := []struct {
+		name      string
+		hexAddr   string
+		expBech32 string
+		expHex    string
+	}{
+		{
+			name:      "valid hex address is converted faithfully",
+			hexAddr:   hex,
+			expBech32: bech32,
+			expHex:    hex,
+		},
+		{
+			name:      "garbage is silently zeroed",
+			hexAddr:   "blabla",
+			expBech32: zeroAddr,
+			expHex:    "0x0000000000000000000000000000000000000000",
+		},
+		{
+			name:      "empty string is silently zeroed",
+			hexAddr:   "",
+			expBech32: zeroAddr,
+			expHex:    "0x0000000000000000000000000000000000000000",
+		},
+		{
+			// The fund-loss case: the zero address is not on the bank blocked
+			// list, so a transfer to it succeeds and the coins are gone.
+			name:      "bech32 address is silently zeroed",
+			hexAddr:   "evmos1ltzy54ms24v590zz37r2q9hrrdcc8eslndsqwv",
+			expBech32: zeroAddr,
+			expHex:    "0x0000000000000000000000000000000000000000",
+		},
+		{
+			name:      "cosmos1 bech32 address decodes to 0x..0C, not to zero",
+			hexAddr:   "cosmos18wvvwfmq77a6d8tza4h5sfuy2yj3jj88yqg82a",
+			expBech32: zeroAddrPlus0C,
+			expHex:    "0x000000000000000000000000000000000000000C",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := utils.Bech32StringFromHexAddress(tc.hexAddr)
+			require.Equal(t, tc.expBech32, got)
+
+			// No error is ever reported, which is what makes this dangerous.
+			gotAddr, err := utils.HexAddressFromBech32String(got)
+			require.NoError(t, err)
+			require.Equal(t, tc.expHex, gotAddr.Hex())
+		})
+	}
+}
+
 func TestGetIBCDenomAddress(t *testing.T) {
 	testCases := []struct {
 		name        string
--- a/x/vm/client/cli/tx.go
+++ b/x/vm/client/cli/tx.go
@@ -7,12 +7,12 @@ import (
 	"os"
 	"strings"
 
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"github.com/cosmos/evm/utils"
 	"github.com/cosmos/evm/x/vm/types"
 
 	"cosmossdk.io/core/address"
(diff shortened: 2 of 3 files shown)


x/vm/client/cli/tx.go
```diff
@@ -156,10 +165,12 @@ When using '--dry-run' a key name cannot be used, only an 0x or bech32 address.
...
 			}
 
-			toAddr, err := ac.StringToBytes(utils.Bech32StringFromHexAddress(args[1]))
+			// accountToHex accepts only 0x and bech32.
+			toHex, err := accountToHex(args[1])
 			if err != nil {
-				return err
+				return errors.Wrap(err, "invalid to-address")
 			}
+			toAddr := common.HexToAddress(toHex).Bytes()
```
gjermundgaraba, line 173: `ac.StringToBytes` already validates hex or configured-prefix bech32 and keeps the decoded length, and it is what this line called on main underneath the mangling wrapper. going through `accountToHex` instead crops a 32-byte bech32 such as an ICA host account to its last 20 bytes and builds a `MsgSend` to an unrelated recipient without error (`cosmos1qypqxpq9qcrsszg2pvxq6rs0zqg3yyc5z5tpwxqergd3c8g7rusqqlvp8l` becomes `cosmos1p58q7yq3zgf3g9gkzuvpjxsmrsw3u8eqcxlj55`; main sent it to `0x…0C`), and it rejects `0X…` hex that main accepted.

```suggestion
			toAddr, err := ac.StringToBytes(args[1])
			if err != nil {
				return errors.Wrap(err, "invalid to-address")
			}
```

x/vm/client/cli/tx.go
```diff
@@ -142,9 +142,18 @@ When using '--dry-run' a key name cannot be used, only an 0x or bech32 address.
...
 			if strings.HasPrefix(args[0], "0x") {
-				fromAddr = utils.Bech32StringFromHexAddress(args[0])
+				fromHex, err := accountToHex(args[0])
+				if err != nil {
+					return errors.Wrap(err, "invalid from-address")
+				}
+				fromAddr, err = ac.BytesToString(common.HexToAddress(fromHex).Bytes())
+				if err != nil {
+					return err
+				}
```
gjermundgaraba, line 156: under the `0x` guard `accountToHex` takes its hex branch, so this adds a hex-string round trip that `ac.StringToBytes` avoids, and once both sites use the codec the `common` import is unused.

```suggestion
				bz, err := ac.StringToBytes(args[0])
				if err != nil {
					return errors.Wrap(err, "invalid from-address")
				}
				fromAddr, err = ac.BytesToString(bz)
				if err != nil {
					return err
				}
```

x/vm/client/cli/utils_test.go
```diff
@@ -41,6 +41,9 @@ func TestAddressFormats(t *testing.T) {
 		{"invalid hex ethereum address", "0x3B98C72760F7BBA69D62ED6F48278451251948E", "", true},
 		{"invalid Cosmos address", "cosmos18wvvwfmq77a6d8tza4h5sfuy2yj3jj88", "", true},
 		{"empty string", "", "", true},
+		{"Cosmos address with bad checksum", "cosmos18wvvwfmq77a6d8tza4h5sfuy2yj3jj88yqg82b", "", true},
+		{"unknown bech32 prefix", "notaprefix18wvvwfmq77a6d8tza4h5sfuy2yj3jj88wpxu9y", "", true},
```
gjermundgaraba, line 45: `…yqg82b` fails on charset (`b` is not in the bech32 alphabet) before any checksum runs, and `notaprefix…wpxu9y` has a bad checksum for its hrp, so a gate that accepted any well-formed foreign prefix would still reject it and this row would stay green.

```suggestion
		{"Cosmos address with bad checksum", "cosmos18wvvwfmq77a6d8tza4h5sfuy2yj3jj88yqg82c", "", true},
		{"Cosmos address with char outside bech32 charset", "cosmos18wvvwfmq77a6d8tza4h5sfuy2yj3jj88yqg82b", "", true},
		{"unknown bech32 prefix", "evmos1ltzy54ms24v590zz37r2q9hrrdcc8eslndsqwv", "", true},
```

x/vm/client/cli/utils_test.go
```diff
@@ -98,3 +101,53 @@ func TestAddressToCosmosAddress(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, baseAddr, ethFormatted)
 }
+
+// TestAccountToHexNeverReturnsZeroAddress pins that accountToHex does
+// never fall back to the zero address.
+func TestAccountToHexNeverReturnsZeroAddress(t *testing.T) {
```
gjermundgaraba, line 107: three of these five inputs are already rows in `TestAddressFormats`, and `require.EqualError` freezes btcutil's checksum wording (the btcsuite fork already phrases it differently) plus the `0x`-prepended message from `accountToHex`, so a reworded dependency or a fix of that message breaks this test with no parsing change. the name also over-claims: `accountToHex` returns the zero address with a nil error for `0x0000000000000000000000000000000000000000` and `cosmos1qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqnrql8a`.

utils/utils_test.go
```diff
@@ -312,6 +312,70 @@ func TestAddressConversion(t *testing.T) {
 	require.Equal(t, hex, gotAddr.Hex())
 }
 
+// TestBech32StringFromHexAddressManglesNonHexInput pins the behavior
+// that non-hex addresses are parsed into garbage addresses
+func TestBech32StringFromHexAddressManglesNonHexInput(t *testing.T) {
```
gjermundgaraba, line 317: this pins the mangled output as the expected value, and `query evm 0x-to-bech32` and `balance-bank` still expose that behavior to user input. Since you offered to drop this demonstration test, I'd remove it.

gjermundgaraba (changes requested):

x/vm/client/cli/tx.go
fragwuerdig: applied

utils/utils_test.go
fragwuerdig: removed

gjermundgaraba (approved):

gjermundgaraba:
@greptile review

gjermundgaraba:
@greptile review

